PR #18, opened 2018-05-02
Routing and eventList

JoakimSM:


config/rewire.config.js
```diff
@@ -135,6 +136,12 @@ function rewireAliases(config) {
     return config;
 }
 
+function addCssModuleRules(config) {
```
varl: Seems like a useless function?

public/i18n/module/i18n_module_en.properties
```diff
@@ -22,3 +22,12 @@ abort=abort
```
varl: Take a look at the [d2-i18n library](https://github.com/dhis2/d2-i18n#table-of-contents) we've standardised on for DHIS2 apps.

src/components/App/App.component.js
```diff
@@ -1,9 +1,14 @@
 // @flow
 /* eslint-disable */
-import React, {Component} from 'react';
+import './app.css';
+import * as React from 'react';
```
varl: Isn't this the same as `imoprt React from 'react'`?

public/i18n/module/i18n_module_en.properties
JoakimSM: Yes, I'll look into it during the next sprint.

src/components/App/App.component.js
JoakimSM: flow types are also imported when using * as React (among other stuff I guess)

config/rewire.config.js
JoakimSM: I probably dozed off. Fixed.

JoakimSM:
For now, you need to supply the orgUnit and program in the uri. something like: http://localhost:3000/#/programId=eBAyeGv0exc&orgUnitId=Rp268JB6Ne4

Notes:
- Icons are the wrong color due to the @material-ui/icons package importing 'material-ui/SvgIcon' which is the old material-ui package. **TODO: remove old material-ui package.** (node-modules/@material-ui/icons/utils/createSvgIcon.js line 15, change it to 'material-ui-next' and you'll see the correct colors)
- The only ordering currently working correctly is eventDate ('report date') because the api isn't returning the right results (server task created for this)

Bekkalizer:
Is this pull request ready to merge?

PR #19, opened 2018-05-03
Quick selector

lukasewvd:
The first implementation of the quick selector. Currently displays the program and the orgUnit from routing.


lukasewvd:
Quick selector has an earlier version of routing merged into it (de04bdd).

PR #22, opened 2018-05-28
Ability to select program w/ catCombo in Selector

lukasewvd:
- Program can be selected.
- If program has catCombos, it can be selected.

- Still some general CSS improvments needs to be done to the react-virtualized-select drop-down.


src/core_modules/capture-core/src/components/Pages/MainPage/MainPage.component.js
```diff
@@ -37,5 +35,5 @@ class MainPage extends Component<Props> {
         );
     }
 }
-
-export default withSelectionsUpdater()(MainPage);
+// Disabled routing for testing
+export default MainPage;
```
lukasewvd: Need to be re-enabled.

PR #23, opened 2018-05-30
new event page

JoakimSM:


flow/material-ui-core.js
```diff
@@ -0,0 +1,66 @@
...
+declare module '@material-ui/core/Paper' {
+    declare module.exports: any;
+}
+
+declare module '@material-ui/core/colors/grey' {
+    declare module.exports: any;
+}
+
+declare module '@material-ui/core/colors/blue' {
+    declare module.exports: any;
```
varl: Does `any` give us any benefits when `type` checking at all?

JoakimSM: You're right, it doesn't. The reason behind the declaration is to let flow know a module exists under '@material/core/colors/blue' and get rid of the "cannot resolve module" error in the import statement.

varl: Ok, just wondering. I'm a Flow newbie so bear with me. :dancing_men:

src/core_modules/capture-core/src/components/Pages/NewEvent/DataEntry/epics/saveNewSingleEvent.epics.js
```diff
@@ -0,0 +1,86 @@
...
+                orgUnit: state.currentSelections.orgUnitId,
+                dataValues: Object
+                    .keys(formServerValues)
+                    .map(key => ({
+                        dataElement: key,
+                        value: formServerValues[key],
+                    })),
+            };
+
+            return getApi().post('events', serverData)
```
varl: Doing the requests like this won't make use of the offline capabilities of `redux-offline`. To get write resilience we need to do some additional stuff: https://github.com/redux-offline/redux-offline/blob/develop/docs/basics/write-resilience.md

When we create the `action`, we need to pass along some additional metadata in the `meta.offline` field. I see that the `actionCreator` does expose a parameter for meta information, so that's the point where we need to add it I think. The data needed is (flow types are available):

```
type OfflineAction = {
  type: string,
  payload: any,
+ meta: {
+   offline: {
+     effect: any,
+     commit: Action,
+     rollback: Action
+   }
+ }
```

To use `getApi().post('events', serverData)` as an effect we need to write our own reconciler as outlined in: https://github.com/redux-offline/redux-offline/blob/develop/docs/recipes/customize-requests.md

> Called with action.meta.offline.effect and the action itself, this method must return a Promise. Resolve the promise if the request is a success and reject otherwise. If rejected, the error will be used by config.discard to decide whether to attempt the request again.

Or, since this is just a POST request, we could forgo d2 and use a `fetch` request and just have something like: 

```
actionCreator(actionTypes.START_SAVE_RETURN_TO_MAIN_PAGE)({ eventId, dataEntryId, {
  ...formFoundation
  offline: {
    effect: { url: '/api/29/events', headers: {...}, body: JSON.stringify(serverData) }
    commit: { type: 'NEW_EVENT_SAVED_AFTER_RETURNED_TO_MAIN_PAGE', meta: {...} },
    rollback: { type: 'SAVE_FAILED_FOR_NEW_EVENT_AFTER_RETURNED_TO_MAIN_PAGE', meta: {...} }
  }
}
})
```

Same goes for Editing requests in: https://github.com/dhis2/capture-app/pull/26/files

varl: Link to #26

PR #29, opened 2018-06-11
DHIS2-3993 Add support for d2-i18n translations

varl:
This PR adds integration to `d2-i18n` and migrates all localised strings already in the application to the new format.

Some strings, like the messages from `@dhis2/d2-ui-forms` depends on that library integrating d2-i18n and emitting translated strings to display. Not exactly sure how that's going to work right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "redux-devtools-extension": "^2.13.2"
   },
   "manifest.webapp": {
-    "name": "Dashboards app",
+    "name": "Capture app",
     "icons": {
       "48": "icon.png"
     },
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
-    "build": "yarn localize && node scripts/build.js",
+    "build": "yarn localize && react-app-rewired build",
     "postbuild": "mkdirp buildzip/ && pushd build && bestzip ../buildzip/NewCaptureApp.zip * && popd",
     "test": "react-app-rewired test --env=jsdom",
     "test:debug": "react-app-rewired --inspect-brk test --runInBand --env=jsdom",
@@ -86,22 +86,24 @@
     "redux-devtools-extension": "^2.13.2"
   },
   "manifest.webapp": {
-    "name": "Capture app",
+    "version": "0.1",
+    "name": "Capture",
+    "description": "Capture App",
+    "launch_path": "index.html",
     "icons": {
-      "48": "icon.png"
+      "16": "/assets/icons/track16.png",
+      "48": "/assets/icons/track48.png",
+      "128": "/assets/icons/track128.png"
     },
     "developer": {
-      "url": "",
-      "name": "DHIS2"
+      "name": "HISP",
+      "url": "http://dhis2.org"
     },
-    "dhis2": {
-      "apiVersion": "30"
-    },
+    "default_locale": "en",
     "activities": {
       "dhis": {
-        "href": ".."
+          "href": "../../.."
       }
-    },
-    "description": "New Capture app"
+    }
   }
 }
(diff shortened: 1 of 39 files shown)


package.json
```diff
@@ -42,14 +48,22 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
-    "build": "react-app-rewired build",
+    "build": "yarn localize && node scripts/build.js",
```
JoakimSM: what is the difference between "node scripts/build.js" and "react-app-rewired"?

package.json
```diff
@@ -70,5 +84,24 @@
     "mkdirp": "^0.5.1",
     "react-app-rewired": "^1.5.0",
     "redux-devtools-extension": "^2.13.2"
+  },
+  "manifest.webapp": {
```
JoakimSM: is this information needed in both package.json and manifest.webapp? And d2-manifest overwrites manifest.webapp?

public/manifest.webapp
```diff
@@ -1,21 +1,23 @@
...
-        "128": "/assets/icons/track128.png"
-    },
-    "developer": {
-        "name": "HISP",
-        "url": "http://dhis2.org"
-    },
-    "default_locale": "en",
-    "activities": {
-        "dhis": {
-            "href": "../../.."
```
JoakimSM: When the app is uploaded to play the path will be something like https://play.dhis2.org/2.29/api/apps/Capture/index.html#/. So I think we need the "../../.."

varl: Right, thanks. I'll fix it.

package.json
varl: d2-manifest uses the `package.json` one to generate it at build, so we only need it in `package.json`.

package.json
varl: Hmm... I didn't mean to change to `node scripts/build.js`.

src/init/init.js
```diff
@@ -40,58 +37,36 @@ async function initializeManifest() {
...
-    config.i18n.sources.add(`i18n/module/i18n_module_${locale}.properties`);
-    return keyUiLocale;
+function isLangRTL(code) {
+    const langs = ['ar', 'fa', 'ur']
+    const prefixed = langs.map(c => `${c}-`)
+    return langs.includes(code) || prefixed.filter(c => code.startsWith(c)).length > 0
 }
 
-// TODO: Make api-requests?
-function getLocaleSpecs(locale: string) {
```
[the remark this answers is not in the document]
varl: Right now we are just including one language from `locales` (en), but I think when we add additional ones we will get e.g. norwegian. I'm going to test this theory though.

```
// in app entry point
import './locales'

// to change locales
function changeLocale(locale) {
      moment.locale(locale)
      i18n.changeLanguage(locale)
      document.body.setAttribute('dir', isLangRTL(locale) ? 'rtl' : 'ltr')
}
```

package.json
varl: Fixed.

package.json
varl: Fixed.

public/manifest.webapp
varl: Fixed.

src/init/init.js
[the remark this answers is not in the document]
varl: If it doesn't work it should be fixed in d2-i18n I think since it claims to:

>On dev/build phase src/locales directory would be created. It will contain necessary setup for internationalization, date/time, calendar etc. It is auto-generated, so please don't update it, any changes to it will be lost.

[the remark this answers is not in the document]
JoakimSM: Thanks for clarifying. I like your changes. So, as far as I understand, to make other locales work we need: - The pot file for the locale -make sure the locale file for dateFn exists

PR #32, opened 2018-06-14
Offline event list, app logic for event list

JoakimSM:
- offline event list
- app logic for event list updates
- some offline logic


JoakimSM:
Thanks Viktor. I somehow forgot. Replaced.

PR #44, opened 2018-07-06
[DHIS2-4138, DHIS2-4144] File and image data type

zepto5:
File element type
Image element type
Async update fields
Data value - subvalues
minor fixes and improvements
borderBox


src/core_modules/capture-core/src/components/BorderBox/borderBox.component.js
```diff
@@ -0,0 +1,39 @@
+// @flow
```
JoakimSM: components filename should be Pascal case (BorderBox)

PR #45, opened 2018-07-10
Org. Unit Tree component

adeelshahid:
### Features
- Multiple / Single selection
- Selectable / Disabled

### Usage
```js
import { OrgUnitTree } from '../../core_modules/capture-core/src/components/FormFields/OrgUnitTree/OrgUnitTree.component'
<OrgUnitTree />
```


Bekkalizer:
I could not get the orgunit tree to show up in the data entry form. Testing with the "inpatient morbodoty and mortality" from Sierra Leone, the organisation unit field is not included by the form builder.

src/core_modules/capture-core/src/components/D2Form/field/buildField.js
```diff
@@ -201,6 +201,7 @@ const fieldForTypes = {
     [elementTypes.TIME]: (metaData: MetaDataElement) => getBaseTextField(metaData),
     [elementTypes.PERCENTAGE]: (metaData: MetaDataElement) => getBaseTextField(metaData),
     [elementTypes.URL]: (metaData: MetaDataElement) => getBaseTextField(metaData),
+    [elementTypes.ORG_UNIT_TREE]: (metaData: MetaDataElement) => getBaseTextField(metaData),
```
Bekkalizer: Shoud this insert the org unit field component? it seems to include a basic text field?

adeelshahid: I had no idea how to include the element type except copy code and typing for _elementTypes_

src/core_modules/capture-core/src/metaData/DataElement/elementTypes.js
```diff
@@ -23,6 +23,7 @@ const elementTypeCodes = [
     'DURATION_MINUTES',
     'FILE_RESOURCE',
     'URL',
+    'ORG_UNIT_TREE',
```
JoakimSM: the value type for this dataelement is ORGANISATION_UNIT

adeelshahid: updated

src/core_modules/capture-core/src/components/D2Form/field/buildField.js
```diff
@@ -210,11 +210,12 @@ const fieldForTypes = {
             props,
         }, metaData);
     },
-    [elementTypes.TIME]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData, options),
-    [elementTypes.PERCENTAGE]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData, options),
-    [elementTypes.URL]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData, options),
+    [elementTypes.TIME]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData),
+    [elementTypes.PERCENTAGE]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData),
+    [elementTypes.URL]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData),
+    [elementTypes.ORGANISATION_UNIT]: (metaData: MetaDataElement, options: Object) => getBaseTextField(metaData),
```
Bekkalizer: You would need to replace the getBaseTextField() with a call that builds your orgunit field. You can look at the BOOL field as an example.

adeelshahid:
@JoakimSM _this.props.onBlur_ added to receive **selected** nodes.

lukasewvd:
**Testing results:**
- I get an error when selecting/deselecting an orgUnit: _Cannot read property 'includes' of undefined._
- Might not be related to this PR but when an event with orgUnit is saved it displays the path not the name in the events list.
- Leaf orgUnits have expand button visible. 
- Event can be saved without orgUnit even whe marked as compulsory.

adeelshahid:
@lukasewvd 

✅ I get an error when selecting/deselecting an orgUnit: Cannot read property 'includes' of undefined.
✅ Leaf orgUnits have expand button visible.

_- Event can be saved without orgUnit even when marked as compulsory._
Looking into this one now

_- Might not be related to this PR but when an event with orgUnit is saved it displays the path not the name in the events list._
How can we resolve this. Should I send in an object in the blur event containing the label and path?

lukasewvd:
@adeelshahid 
> Event can be saved without orgUnit even when marked as compulsory.
How can I do this, can you point me to an place where I can see an example implementation to validate OU.

_If you look at a programStage in the API (i.e: https://play.dhis2.org/dev/api/29/programs/q04UBOqq3rp/programStages/pSllsjpfLH2) under programStageDataelements each dataElement has a property called **compulsory**. I that is set to true a new event cannot be saved without that dataElement being filled out._

------

> Might not be related to this PR but when an event with orgUnit is saved it displays the path not the name in the events list.
How can we resolve this. Should I send in an object in the blur event containing the label and path?

I am not sure what the best solution for this is. We need some way of "translating the path/id" to the orgUnit name. I would consult with @JoakimSM on posible solutions.

adeelshahid:
@lukasewvd how do I set this validation on the frontend side is there a function or field i can add to my OrgUnitTree field to emit valid/invalid state check

adeelshahid:
@lukasewvd I have added a Org Unit validator to reject empty as valid value for Org. Unit Tree.

adeelshahid:
Form error validations are for some reason persisted in storage rather then computed. So there is sync issue with the UI even when the *Org. Unit* is valid. How do I re-validate the form field when I select or de-select a value.

adeelshahid:
@zepto5 if everything works good with Org. Unit can you approve the merge.

adeelshahid:
@zepto5 does it look good for a merge now

PR #53, opened 2018-07-25
Save and add another when creating new events

lukasewvd:
Made by: @zepto5 
Ability to add multiple events after each other without leaving data entry.


lukasewvd:
**Testing results:**
- Needs a rebase.
- When saving and adding another a warning appears: _Warning: Each child in an array or iterator should have a unique "key" prop._  
- Should the events added list persist when a program or orgunit is deselected? 
- It is currently possible to add multiple events from different programs. They all get added to the same list, but only the header text from the first created event is showing. Should it be possible to add events from different program to the list?
- After adding multiple events and going back to the events list the newly added events are not displayed.

Bekkalizer:
This branch is outdated - will make a PR from the feedback branch instead

PR #55, opened 2018-08-06
Phone number field

adeelshahid:


zepto5:
We should leave this field as a vanilla text field for now, as a phone number can contain leading zeros, paranteses, +,#, even some text, etc. There are also different rules in different countries.  Phone number validation can be handled with program rules.

adeelshahid:
@zepto5 updated

PR #56, opened 2018-08-14
Age Field

adeelshahid:
Implemented


adeelshahid:
Similar to Joe's design and consistent with current **material-ui** implementation.

_Example Screenshot_

screenshot from 2018-08-14 14-59-01

lukasewvd:
After some testing this looks good and everything seems to work as intended.

PR #60, opened 2018-08-29
Field/username autocomplete

adeelshahid:


adeelshahid:
> -The username type fields needs to display the data elements displayFormName - this is the context under which the user selects a user name.

I wasn't able to find the above prop **displayFormName** so I have used **label* prop same as in **OrgUnitTree** to show the form label.

adeelshahid:
* label added
* onBlur is now trigged on keyboard / mouse selection of user

PR #61, opened 2018-09-03
Field/coordinate location

adeelshahid:
- Pick lat/lng from map by clicking on location marker & map or enter them manually.
- Not sure what to use for default values for lat/lng so kept it empty.


zepto5:
* Map does not open when clicking the location marker. 

* Empty lat and lng results in "Please provide valid coordinates", even when coordinate field is not required.

adeelshahid:
@zepto5 

`const isValidCoordinate = (value: Location, isEmptyValid: boolean = false) => {`

I have added the 2nd param locally. It returns **isEmptyValid = false** even when **compulsory** is not set in the form.

Should these validators not be passed the correct isEmptyValid value somehow. Then I can ignore empty value as invalid.

adeelshahid:
@zepto5 

> Map does not open when clicking the location marker.
fixed, styles/scripts were missing for leaflet

> Empty lat and lng results in "Please provide valid coordinates", even when coordinate field is not required.
fixed, emitting null as suggested.

JoakimSM:
> @zepto5
> 
> `const isValidCoordinate = (value: Location, isEmptyValid: boolean = false) => {`
> 
> I have added the 2nd param locally. It returns **isEmptyValid = false** even when **compulsory** is not set in the form.
> 
> Should these validators not be passed the correct _isEmptyValid_ value somehow. Then I can ignore empty value as invalid.

Ignore isEmptyValid. I removed them from the codebase just now, it was a bad decision adding them. The validators aren't called if the value is empty.

adeelshahid:
was removed `const isValidCoordinate = (value: Location) => {` in an earlier commit after @zepto5 pointed out _isEmptyValid_ is not used.

adeelshahid:
@zepto5 let's merge it if nothing remains on this one. I am using **Point**/**Polygon** for featureType based on this.

PR #62, opened 2018-09-17
Polygon Component

adeelshahid:
Polygon component using leaflet draw package.


adeelshahid:
@zepto5 / @joakimia who so ever is free to review :slightly_smiling_face:

zepto5:
Pushed a minor fix for converting 'Point'. 

Please make the Polygon more like a field. We want the Polygon field to behave like the coordinate field. This means that the map should only be displayed when clicking the marker icon. The field should display "Polygon captured" or "No polygon captured".  Use the example I sent September 14th.

adeelshahid:
updated with _Polygon captured_ status

PR #102, opened 2019-04-03
Adding Vietnamese translations.

jason-p-pickering:


jason-p-pickering:
@zepto5 and @JoakimSM seems the build is failing as well on Travis.

PR #522, opened 2020-04-14
fix: sanitises the long and lat value to be a six decimal

paschalidi:
Hey @JoakimSM  👋 👋 

As we talked and after your suggestions which were very helpful I am opening this PR which addresses the https://jira.dhis2.org/browse/DHIS2-7300. 

We are dealing with two cases here as you already know. 
* The first case where the user opens the map and drops a pin on the map. When the user closes the map then the `lat` and `long` will be put in the input form. In this case we sanitise the values on the `onMapPositionChange `
* The second case is when the user adds the value manually in the input. Here we sanitise the value on the blur event of the input. This way the user can type as much decimals as they like and when they "leave" the input then the value gets sanitised to 6 decimals. 

Let me know your thoughts or suggestions!!

### DIFF
--- a/src/core_modules/capture-ui/CoordinateField/CoordinateField.component.js
+++ b/src/core_modules/capture-ui/CoordinateField/CoordinateField.component.js
@@ -56,7 +56,7 @@ export default class D2Coordinate extends React.Component<Props, State> {
         };
     }
 
-    toSixDecimal = value => parseFloat(value).toFixed(6)
+    toSixDecimal = value => parseFloat(value) && parseFloat(value).toFixed(6)
 
     handleBlur = (key: string, value: any) => {
         const newValue = { ...this.props.value, [key]: this.toSixDecimal(value) };
--- a/src/core_modules/capture-ui/CoordinateField/CoordinateField.component.js
+++ b/src/core_modules/capture-ui/CoordinateField/CoordinateField.component.js
@@ -56,7 +56,7 @@ export default class D2Coordinate extends React.Component<Props, State> {
         };
     }
 
-    toSixDecimal = value => parseFloat(value) && parseFloat(value).toFixed(6)
+    toSixDecimal = (value: string) => (parseFloat(value) ? parseFloat(value).toFixed(6) : null)
 
     handleBlur = (key: string, value: any) => {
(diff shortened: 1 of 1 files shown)


paschalidi:
@JoakimSM very good catch! I am looking into it! Thanks for testing it

JoakimSM:
I can see that "NaN" is gone from the UI. Good 👍 

I do have a request for an improvement though. In the redux state when one of the fields is filled in, but the other one is not the state would look something like this: {latitude: NaN, longitude: "0.453959"}. We should follow the pattern from the other field types here and go with null instead of NaN (You could argue that "" or undefined is the better option, but let's stick with null for now). So, the the redux state should look like this:

- One field filled in: {latitude: null, longitude: "0.453959"}
- No fields filled in: null
- Both fields filled in: {latitude: "0.453959", longitude: "0.453959"}

Why?
- As mentioned, following the pattern from the other fields. 
- I know null is handled in regards to validation, program rules etc. So less likely to get any surprises later.
- NaN feels a bit off because when the field has a value this is a string (I like the string though, you should keep it because the other field types is also using string)
- Also, NaN follows several tricky rules like NaN === NaN; -> false

paschalidi:
@JoakimSM cant find whist state you are referring to. Whats the object key labeled in the redux store?

JoakimSM:
> @JoakimSM cant find whist state you are referring to. Whats the object key labeled in the redux store?

It depends on wheter the field is rendered based on a programstage data element (redux prop: formsValues) or the feature type of the program stage (redux prop: dataEntriesFieldsValue). But if your're for example looking at the topmost coordinate field of the malaria case registration form (and you are creating a new event) you will find the value in State -> dataEntriesFieldsValue -> singleEvent-newEvent -> geometry. Don't use Redux DevTools to check because that value is stringified. Check to console (all redux actions are logged together with the resulting state) or debug the first React Redux connect argument. Also, in the render function of the CoordinateField component you could check the value prop to see that the value is NaN.

paschalidi:
@JoakimSM the improvement request must be on its way now! That makes sense, thanks for raising the concern with me I didnt realised that there was a `NaN` value and as you say this would have created issued probably : ). `NaN` === evil

PR #532, opened 2020-04-15
chore: removes dependencies that are not used and moved dependencies that are not used in production into dev deps

paschalidi:
Hey @JoakimSM  👋 

I went thought the project the other day and had a look at the deps that are not in use. 
I thought deleting the ones that we dont use will reduce bundle size and also help us having an even more maintainable piece of software. You can find the ticket for this [here](https://jira.dhis2.org/browse/TECH-353)

I also moved the dependencies that are not being in use for production code at the `devDependency` section of the `package.json` 

As always please let me know your thoughts on the matter! I wanted to mention that I havent really communicated with you before opening this PR. Apologies for that. If you find it useful we could have a chat first the next time. Just let me know!! 

I of course have tested that the app works fine on my machine.


paschalidi:
@JoakimSM I will merge that after I resolve the conflicts

paschalidi:
@JoakimSM since this is not a bug fix is it supposed to be backported?

PR #534, opened 2020-04-15
fix: DHIS2-8612 moment locale using non european glyphs fix

JoakimSM:
Resolved some issues caused by using a moment locale with non european glyphs.
- Saving an event when the event was marked for completion caused the post request to fail due to the complete date being posted using non european glyphs.
- Filtering a view using a date field caused the filtering to fail.
- Program rules would interpret dates wrong.

### DIFF
--- a/src/core_modules/capture-core-utils/date/date.utils.js
+++ b/src/core_modules/capture-core-utils/date/date.utils.js
@@ -1,74 +1,15 @@
 // @flow
-import moment from '../moment/momentResolver';
 
-export const displayTypes = {
-    SHORT: 'short',
-    MEDIUM: 'medium',
-    LONG: 'long',
-};
-
-export function getMonthName(monthNumber: number) {
-    const date = moment([0, monthNumber - 1]);
-    return date.format('MMMM');
+/**
+ * Some locales are using numeral glyphs other than european. This method ensures the moment format method returns a value in european glyphs
+ * @param {*} momentDate: the moment instance
+ * @param {string} format: the moment format
+ * @returns {string} A formatted string with european glyphs
+ */
+export function getFormattedStringFromMomentUsingEuropeanGlyphs(
+    momentDate: moment$Moment,
+    format: string = 'YYYY-MM-DD',
+) {
+    const europeanMoment = momentDate.clone().locale('en');
+    return europeanMoment.format(format);
 }
-
-export function adjustUtcIsoDateToLocal(utcISO8601: string) {
-    const localDate = moment(utcISO8601, 'YYYY-MM-DD');
-    const isoString = localDate.toISOString();
-    return isoString;
-}
-
-export function adjustLocalMomentDateToUtc(momentDateLocal: any) {
-    const momentDateUTC = moment(Date.UTC(momentDateLocal.year(), momentDateLocal.month(), momentDateLocal.date()));
-    return momentDateUTC;
-}
-
-export function adjustLocalIsoDateToUtc(localISO8601: string) {
-    const momentLocale = moment(localISO8601);
-    const momentUtc = adjustLocalMomentDateToUtc(momentLocale);
-    return momentUtc.toISOString();
-}
-
-export const displayDate = (() => {
-    const typeToMomentConverter = {
-        [displayTypes.SHORT]: 'l',
-        [displayTypes.MEDIUM]: 'L',
-        [displayTypes.LONG]: 'LL',
-    };
-
-    return (ISOString: string | Date, displayType?: ?$Values<typeof displayTypes>) => {
-        const date = moment(ISOString);
-        return date.format((displayType && typeToMomentConverter[displayType]) || typeToMomentConverter[displayTypes.MEDIUM]);
-    };
-})();
-
-export const displayDateTime = (() => {
-    const typeToMomentConverter = {
-        [displayTypes.SHORT]: 'LLL',
-        [displayTypes.MEDIUM]: 'llll',
-        [displayTypes.LONG]: 'LLLL',
-    };
-
-    return (ISOString: string | Date, displayType?: ?$Values<typeof displayTypes>) => {
-        const date = moment(ISOString);
-        return date.format((displayType && typeToMomentConverter[displayType]) || typeToMomentConverter[displayTypes.MEDIUM]);
-    };
-})();
-
-export const displayTime = (() => {
-    const typeToMomentConverter = {
-        [displayTypes.SHORT]: 'LT',
-        [displayTypes.MEDIUM]: 'LTS',
-        [displayTypes.LONG]: 'LTS',
-    };
-
-    return (ISOString: string | Date, displayType?: ?$Values<typeof displayTypes>) => {
-        const date = moment(ISOString);
-        return date.format((displayType && typeToMomentConverter[displayType]) || typeToMomentConverter[displayTypes.MEDIUM]);
-    };
-})();
-
-export const RulesEngineDateUtils = {
-    format: (IsoString: string) => moment(IsoString).format('L'),
-    getToday: () => moment().format('L'),
-};
--- a/src/core_modules/capture-core-utils/date/index.js
+++ b/src/core_modules/capture-core-utils/date/index.js
@@ -1,0 +1,2 @@
+// @flow
+export { getFormattedStringFromMomentUsingEuropeanGlyphs } from './date.utils';
--- a/src/core_modules/capture-core/components/Pages/EditEvent/DataEntry/epics/saveEditSingleEvent.epics.js
+++ b/src/core_modules/capture-core/components/Pages/EditEvent/DataEntry/epics/saveEditSingleEvent.epics.js
@@ -1,6 +1,7 @@
 // @flow
 import { push } from 'connected-react-router';
-import moment from 'capture-core-utils/moment/momentResolver';
+import { moment } from 'capture-core-utils/moment';
+import { getFormattedStringFromMomentUsingEuropeanGlyphs } from 'capture-core-utils/date';
 import {
     actionTypes as editEventDataEntryActionTypes,
     startSaveEditEventAfterReturnedToMainPage,
@@ -39,7 +40,7 @@ export const saveEditEventEpic = (action$: InputObservable, store: ReduxStore) =
             const mainDataServerValues: Object = convertMainEventClientToServer(mainDataClientValues);
 
             if (mainDataServerValues.status === 'COMPLETED' && !prevEventMainData.completedDate) {
-                mainDataServerValues.completedDate = moment().format('YYYY-MM-DD');
+                mainDataServerValues.completedDate = getFormattedStringFromMomentUsingEuropeanGlyphs(moment());
             }
 
             const serverData = {
--- a/src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
+++ b/src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
@@ -1,11 +1,13 @@
 // @flow
 import { createSelector } from 'reselect';
 import { moment } from 'capture-core-utils/moment';
+import { getFormattedStringFromMomentUsingEuropeanGlyphs } from 'capture-core-utils/date';
+
 import type {
     DateFilterData,
     RelativeDateFilterData,
     AbsoluteDateFilterData,
-} from '../../../eventList.types';
+} from '../../../../EventsList/eventList.types';
 
 const periods = {
     TODAY: 'TODAY',
@@ -19,15 +21,14 @@ const periods = {
 };
 
 const selectors = {};
-const formatDateForFilterRequest = (dateMoment: moment$Moment) => dateMoment.format('YYYY-MM-DD');
 const relativeConvertersForPeriods = {
     [periods.TODAY]: () => {
         const startDate = moment();
         const endDate = startDate;
 
         return [
-            `ge:${formatDateForFilterRequest(startDate)}`,
-            `le:${formatDateForFilterRequest(endDate)}`,
+            `ge:${getFormattedStringFromMomentUsingEuropeanGlyphs(startDate)}`,
+            `le:${getFormattedStringFromMomentUsingEuropeanGlyphs(endDate)}`,
         ];
     },
     [periods.THIS_WEEK]: () => {
@@ -35,8 +36,8 @@ const relativeConvertersForPeriods = {
         const endDate = moment().endOf('week');
 
         return [
-            `ge:${formatDateForFilterRequest(startDate)}`,
-            `le:${formatDateForFilterRequest(endDate)}`,
+            `ge:${getFormattedStringFromMomentUsingEuropeanGlyphs(startDate)}`,
+            `le:${getFormattedStringFromMomentUsingEuropeanGlyphs(endDate)}`,
         ];
     },
     [periods.THIS_MONTH]: () => {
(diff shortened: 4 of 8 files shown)


paschalidi:
@JoakimSM hey thanks for the PR. 

Would it be easy for you next time to add the link of the ticket to the description? 
Also could you please take a moment and explain how would you test this? I read the ticket and that makes sense. However I have the feeling that this PR resolves more issues than the ticket describes?

The steps I follow are at the [demo](https://play.dhis2.org/2.33.3/dhis-web-capture/index.html#/)
1. change the language to Arabic. 
2. I then pick a program and add an event. 
3. I add the event as completed and press save. 
4. getting an error. 

Is this though the only usecase that this PR solves?

JoakimSM:
> @JoakimSM hey thanks for the PR.
> 
> Would it be easy for you next time to add the link of the ticket to the description?
> Also could you please take a moment and explain how would you test this? I read the ticket and that makes sense. However I have the feeling that this PR resolves more issues than the ticket describes?
> 
> The steps I follow are at the [demo](https://play.dhis2.org/2.33.3/dhis-web-capture/index.html#/)
> 
> 1. change the language to Arabic.
> 2. I then pick a program and add an event.
> 3. I add the event as completed and press save.
> 4. getting an error.
> 
> Is this though the only usecase that this PR solves?

Thanks for your feedback @paschalidi 

Adding a link to the JIRA-issue makes sense, let's standardise on that. Good suggestion.

For the testing, you can find a test case I made yesterday that is linked in the JIRA-issue. [DHIS2-8612](https://jira.dhis2.org/browse/DHIS2-8612) and look in the "Issue Links" section. I also separated some of the problems described into its own issue and made a comment in the JIRA-issue regarding the separation and the problems this ticket would address. Will do a minor change to the issue summary to make it a bit more clear.

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -1,17 +1,25 @@
 // @flow
-import moment from '../moment/momentResolver';
+import { moment } from '../moment';
```
paschalidi, line 2: could we do that instead? 
```suggestion
import moment from 'moment';
```

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -1,17 +1,25 @@
...
+ * @param {string} format: the moment format
+ * @returns {string} A formatted string with european glyphs
+ */
+export function getFormattedStringFromMomentUsingEuropeanGlyphs(momentDate: moment$Moment, format: string) {
+    const europeanMoment = momentDate.clone();
+    europeanMoment.locale('en');
+    return europeanMoment.format(format);
 }
 
+
```
paschalidi, line 22:
```suggestion
```

src/core_modules/capture-core-utils/date/index.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+export { getFormattedStringFromMomentUsingEuropeanGlyphs } from './date.utils';
```
paschalidi, line 2: wanna add this empty line in the end? Eslint complains when we dont have an empty line in the end of our js files.
```suggestion
export { getFormattedStringFromMomentUsingEuropeanGlyphs } from './date.utils';

```

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -67,8 +75,3 @@ export const displayTime = (() => {
         return date.format((displayType && typeToMomentConverter[displayType]) || typeToMomentConverter[displayTypes.MEDIUM]);
     };
 })();
-
-export const RulesEngineDateUtils = {
```
paschalidi, line 71: what makes us wanna delete this bit?

src/core_modules/capture-core/components/Pages/EditEvent/DataEntry/epics/saveEditSingleEvent.epics.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import { push } from 'connected-react-router';
-import moment from 'capture-core-utils/moment/momentResolver';
+import { moment } from 'capture-core-utils/moment';
```
paschalidi, line 3: can we import it like usual here as well?
```suggestion
import moment from 'moment';
```

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -1,17 +1,25 @@
...
 export const displayTypes = {
     SHORT: 'short',
     MEDIUM: 'medium',
     LONG: 'long',
 };
 
-export function getMonthName(monthNumber: number) {
-    const date = moment([0, monthNumber - 1]);
-    return date.format('MMMM');
+/**
```
paschalidi, line 10: Thanks for writing a comment that would explain the function here!! 🏅

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
```diff
@@ -1,11 +1,13 @@
 // @flow
 import { createSelector } from 'reselect';
 import { moment } from 'capture-core-utils/moment';
```
paschalidi, line 3: I hate to say it again but I want to make sure that if we agree that we could import the moment library as we do with all the other dependencies we also do it here. :-)

src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/epics/getConvertedNewSingleEvent.js
```diff
@@ -1,9 +1,10 @@
 // @flow
-import moment from 'capture-core-utils/moment/momentResolver';
+import { moment } from 'capture-core-utils/moment';
```
paschalidi, line 2: +1

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -1,17 +1,25 @@
...
+ * Some locales are using numeral glyphs other than european. This method ensures the moment format method returns a value in european glyphs
+ * @param {*} momentDate: the moment instance
+ * @param {string} format: the moment format
+ * @returns {string} A formatted string with european glyphs
+ */
+export function getFormattedStringFromMomentUsingEuropeanGlyphs(momentDate: moment$Moment, format: string) {
```
paschalidi, line 16: It looks like to me we could use some default values here. 

I can see for example that this function is used all over around our application now with format as `YYYY-MM-DD`, I would then set this value as default so that the user dont have to specify it each time. 
Also the default value for the `momentDate` could be `moment()`

In the end you will have something like this: 
```
export function getFormattedStringFromMomentUsingEuropeanGlyphs(momentDate: moment$Moment = moment(), format: string = 'YYYY-MM-DD') {
```
This will keep the signature of the function simpler for the end user and will also give them the ability to change both of those if they need. For example there are many times that we need to add instead of `moment()` a different instance then this would be easy to do but for the simple cases the user wont have to pass anything :)

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
```diff
@@ -19,7 +21,7 @@ const periods = {
 };
 
 const selectors = {};
-const formatDateForFilterRequest = (dateMoment: moment$Moment) => dateMoment.format('YYYY-MM-DD');
+const formatDateForFilterRequest = (dateMoment: moment$Moment) => getFormattedStringFromMomentUsingEuropeanGlyphs(dateMoment, 'YYYY-MM-DD');
```
paschalidi, line 24: This follows my previous [comment](https://github.com/dhis2/capture-app/pull/534/files#r409663542).

If we have the default value on the function signature now  we can simply use `getFormattedStringFromMomentUsingEuropeanGlyphs ` instead of `formatDateForFilterRequest ` and this way we have a function less to care about.

src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/epics/getConvertedNewSingleEvent.js
```diff
@@ -1,9 +1,10 @@
...
+import { getFormattedStringFromMomentUsingEuropeanGlyphs } from 'capture-core-utils/date';
 import convertDataEntryToClientValues from '../../../../DataEntry/common/convertDataEntryToClientValues';
 import { convertValue as convertToServerValue } from '../../../../../converters/clientToServer';
 import { convertMainEventClientToServer } from '../../../../../events/mainConverters';
-import RenderFoundation from '../../../../../metaData/RenderFoundation/RenderFoundation';
+import { RenderFoundation } from '../../../../../metaData';
```
paschalidi, line 7: Thumbs up for using the index!! We should do more of that! 👍

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/exposedHelpers/getRelationshipNewTei.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import uuid from 'd2-utilizr/src/uuid';
-import moment from 'capture-core-utils/moment/momentResolver';
+import { moment } from 'capture-core-utils/moment';
```
paschalidi, line 3: looks like this is not being used (?) Can we delete it if it is not used?

src/core_modules/capture-core/components/Pages/ViewEvent/EventDetailsSection/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -2,6 +2,7 @@
 import { ActionsObservable } from 'redux-observable';
 import { batchActions } from 'redux-batched-actions';
 import { moment } from 'capture-core-utils/moment';
```
paschalidi, line 4: looks like this is not used here as well. Since we touched this file in this PR can we pretty please delete it?

src/core_modules/capture-core/rulesEngineActionsCreator/converters/momentConverter.js
```diff
@@ -1,7 +1,8 @@
 // @flow
 /* eslint-disable class-methods-use-this */
 import moment from 'moment';
-import type { IMomentConverter } from '../RulesEngine/rulesEngine.types';
+import { getFormattedStringFromMomentUsingEuropeanGlyphs } from 'capture-core-utils/date';
+import type { IMomentConverter } from 'capture-core-utils/RulesEngine/rulesEngine.types';
 
 const momentFormat = 'YYYY-MM-DD';
```
paschalidi, line 7: so if we decide to change the signature and add the default values we wont need to pass this value here any more 👯

src/core_modules/capture-core-utils/date/date.utils.js
```diff
@@ -1,17 +1,25 @@
...
+    const europeanMoment = momentDate.clone();
```
paschalidi, line 17: I would do this is this was possible. 
```suggestion
    const europeanMoment = momentDate.clone().locale('en');
```

src/core_modules/capture-core-utils/date/date.utils.js
JoakimSM: When we started using moment in our app, there was an issue with moment causing the global locale setting not be preserved without this workaround (i.e. the global moment locale would always be 'en' even though we set it to something else). Had a quick look at our app code now and since the direct node module import is used in some places previously causing problems, the issue is most likely resolved. But to keep our code  consistent (for the most part) and be able to verify that the issue is resolved, let's keep the import like it is for now and do this update for the entire app through a maintenance task.

src/core_modules/capture-core-utils/date/date.utils.js
JoakimSM: Agreed, deleted them all. The reason the RulesEngineDateUtils was deleted in the first place  was because it was incorrect.

src/core_modules/capture-core/components/Pages/EditEvent/DataEntry/epics/saveEditSingleEvent.epics.js
JoakimSM: see comment from earlier.

src/core_modules/capture-core-utils/date/date.utils.js
JoakimSM: 👍

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
JoakimSM: see comment from earlier.

src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/epics/getConvertedNewSingleEvent.js
JoakimSM: Definitely. I'm continuesly updating the ones I know I have created indexes for when I touch the files. Possible maintenance tasks in the future.

src/core_modules/capture-core-utils/date/date.utils.js
JoakimSM: :+1 Agreed

src/core_modules/capture-core-utils/date/date.utils.js
JoakimSM: I'm a bit hesitant about this one. I can see your point and the benefits, thanks for pointing them out so we at least can make a conscious decision here. The disadvantage to adding defaults is that flow no longer will warn you that you're forgetting arguments if you intended to use something else than the defaults.

I'll meet you in the middle: No default for the momentDate, but I'm adding the default for the format. Hope that's ok. We can have a heated discussion later ;)

src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/epics/getConvertedNewSingleEvent.js
JoakimSM: Caused by adding the default for the momentDate parameter in getFormattedStringFromMomentUsingEuropeanGlyphs function. I will keep the parameter without default so still need this here.

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/exposedHelpers/getRelationshipNewTei.js
JoakimSM: Caused by adding the default for the momentDate parameter in getFormattedStringFromMomentUsingEuropeanGlyphs function. I will keep the parameter without default so still need this here.

src/core_modules/capture-core/components/Pages/ViewEvent/EventDetailsSection/EditEventDataEntry/editEventDataEntry.epics.js
JoakimSM: Caused by adding the default for the momentDate parameter in getFormattedStringFromMomentUsingEuropeanGlyphs function. I will keep the parameter without default so still need this here.

src/core_modules/capture-core/rulesEngineActionsCreator/converters/momentConverter.js
JoakimSM: 👍

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventsQueryArgsBuilder/filterConverters/dateConverter.js
JoakimSM: 👍

src/core_modules/capture-core-utils/date/index.js
JoakimSM: 👍 I believe my eslint wasn't up and running when I committed this.

JoakimSM:
Thanks for the thorough review @paschalidi. I've made some updates, please see my comments for details.

paschalidi:
Wanna maybe try "Squash and merge" ? This way we will only have to cherry pick one commit.

PR #565, opened 2020-04-20
chore: writes first tests on run rules for single event

paschalidi:
Hey @JoakimSM I am opening this PR which contains only the first test that I am adding. 

I intent to open one more after we review this which will clean up some unused code and do small refinements. Just FYI. 

So this one is create a safety net by testing the `runRulesForSingleEvent()` agains our own data. Therefore later on when I do certain changed then we will know if something is broken or not. 

Of course there will be more tests coming covering other parts of the RulesEngine class but there will be  following this PR.

### DIFF
--- a/src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
+++ b/src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
@@ -59,7 +59,9 @@ export default class RulesEngine {
         const dateUtils = getDateUtils(momentConverter);
         const variableService = new VariableService(valueProcessor.processValue, dateUtils);
         this.executionService = getExecutionService(onTranslate, variableService, dateUtils);
-        this.onProcessRulesEffects = getRulesEffectsProcessor(this.executionService.convertDataToBaseOutputValue, outputRulesConverterObject);
+        this.onProcessRulesEffects = getRulesEffectsProcessor(
+            this.executionService.convertDataToBaseOutputValue,
+            outputRulesConverterObject);
     }
 
     executeRules(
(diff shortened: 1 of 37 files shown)


src/core_modules/capture-core/components/D2Form/field/validators.js
```diff
@@ -1,6 +1,6 @@
 // @flow
-import isArray from 'd2-utilizr/src/isArray';
-import isString from 'd2-utilizr/src/isString';
+import isArray from 'd2-utilizr/lib/isArray';
```
JoakimSM, line 2: 🏆 Thanks for fixing all these

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
```diff
@@ -59,10 +59,7 @@ export default class RulesEngine {
         const dateUtils = getDateUtils(momentConverter);
         const variableService = new VariableService(valueProcessor.processValue, dateUtils);
         this.executionService = getExecutionService(onTranslate, variableService, dateUtils);
-        this.onProcessRulesEffects = getRulesEffectsProcessor(
-            this.executionService.convertDataToBaseOutputValue,
-            outputRulesConverterObject,
-        );
+        this.onProcessRulesEffects = getRulesEffectsProcessor(this.executionService.convertDataToBaseOutputValue, outputRulesConverterObject);
```
JoakimSM, line 62: You would probably get an eslint max-len error for this one?

paschalidi: yes you right! Now is fixed 👍 Thanks

PR #566, opened 2020-04-20
refactor: refactors on runRulesForSingleEvent and removes unused code.

paschalidi:
This one is based on https://github.com/dhis2/capture-app/pull/565

What we are doing here is 
* cleaning up code that is never used on the files that I touched or read through. I hope you agree we don't need to have code that is never used. : ) 
* removed variables that we pass in functions that are never used such as `onTranslate` or `formId`
* try to make the code more readable for the developer by explicitly showing to the reader which variables I am passing in functions instead of passing object properties that are not used. see [here](https://github.com/dhis2/capture-app/compare/cp/writes-first-tests-on-runRulesForSingleEvent...cp/refactors-on-runRulesForSingleEvent?expand=1#diff-91f832c4094c3bd7b657665929a0d27cR103) 

I am trying to split the PRs to as small as possible so that it is easy for you to review em. Le me know if I can do anything to make your life easier and your reviewing less effortless.

### DIFF
--- a/src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
+++ b/src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
@@ -3,13 +3,9 @@ import log from 'loglevel';
 import { RulesEngine, processTypes } from 'capture-core-utils/RulesEngine';
 import type
 {
-    OptionSets,
     ProgramRulesContainer,
     DataElement as DataElementForRulesEngine,
     EventData,
-    OrgUnit,
-    EventMain,
-    EventValues,
 } from 'capture-core-utils/RulesEngine/rulesEngine.types';
 
 import { errorCreator } from 'capture-core-utils';
@@ -22,11 +18,6 @@ import {
 } from '../metaData';
 import constantsStore from '../metaDataMemoryStores/constants/constants.store';
 import optionSetsStore from '../metaDataMemoryStores/optionSets/optionSets.store';
-
-export type EventContainer = {
-    main: EventMain,
-    values: EventValues
-};
 
 const errorMessages = {
     PROGRAM_NOT_FOUND: 'Program not found in loadAndExecuteRulesForEvent',
@@ -42,6 +33,8 @@ function getProgramRulesContainer(program: Program, foundation: RenderFoundation
     const programRules = [...mainProgramRules, ...foundationProgramRules];
 
     const constants = constantsStore.get();
+
+    console.log(constants);
     return {
         programRulesVariables,
         programRules,
@@ -50,7 +43,7 @@ function getProgramRulesContainer(program: Program, foundation: RenderFoundation
 }
 
 function getTrackerDataElements(trackerProgram: TrackerProgram): Array<DataElement> {
-    const elements = Array.from(trackerProgram.stages.values())
+    return Array.from(trackerProgram.stages.values())
         .reduce((accElements, stage) => {
             const stageElements = Array.from(stage.stageForm.sections.values())
                 .reduce((accStageElements, section) =>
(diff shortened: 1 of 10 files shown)


src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -94,35 +70,11 @@ function getDataElements(program: Program) {
     return getRulesEngineDataElementsAsObject(dataElements);
 }
 
-function runRulesEngine(
```
JoakimSM, line 97: I would prefer to keep this function to enclose the quirks of the executeRules function and easily see what data we need to supply when running rules for a single event. But what I would do though is to change the param list, using a maximum of 3 parameters (and annotate it with flow to make it clear what we need to provide).

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -23,50 +18,30 @@ import {
...
-
-    const constants = constantsStore.get();
-    return {
-        programRulesVariables,
-        programRules,
-        constants,
-    };
-}
 
+// TODO why when in our code at the moment we dont have any trackerPrograms we still keep this code?
```
JoakimSM, line 28: 💯 You're right, we should remove the tracker part here.

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -94,35 +70,11 @@ function getDataElements(program: Program) {
     return getRulesEngineDataElementsAsObject(dataElements);
 }
 
-function runRulesEngine(
-    rulesEngine: RulesEngine,
-    programRulesContainer: ProgramRulesContainer,
-    dataElementsInProgram: { [elementId: string]: DataElementForRulesEngine },
-    orgUnit: OrgUnit,
-    optionSets: ?OptionSets,
-    currentEventData: ?EventData | {} = {},
```
JoakimSM, line 103: It is important to keep this default value ({}). If you don't supply a value some event context variables will be missing when running the rules.

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
paschalidi: I agree we should be having functions less parameters to make it easier for the reader to follow what is going on and I agree we should annotate them for the same reason. 

When we make abstractions we want them to be deep so that we can hide complexity from the reader. 

For example if you have a complicated computation that you dont wanna the reader to see then you have it in a function and the user can continue reading the code without having to understand the complex part of the code. 

In my eyes this function sets the default value, annotates the inputs and calls a function. This abstraction is flat and doesnt provide much of functionality. We are now prompting the reader to another function which simply calls yet another function and if the reader wanted follow the pipeline it becomes challenging and increases the cognitive load more. 
To help the reader follow the code and our logic we could handle the quirks and the annotation including the default values and the `nulls` in the previous level without having to introduce one level deeper. 

Does this makes sense to you?

JoakimSM: Your principle is sound and I’ll admit that my stand here is very subjective and situational. To me, the code is more readable with the additional function because, as mentioned, the annotated param list and you also separate the concern of building the quirky arguments for the external function. To me this is more important than preventing adding another level here.

paschalidi: One reason of this refactoring is to make function calls with less variables so that the quirkiness is avoided. We are not there yet but we go there step by step. I agree that it is weird to have 10 variables passed in a function while some of there are `null` etc etc. 

what if I add the same annotations [here](https://github.com/dhis2/capture-app/pull/566/files/cabfad2b9e925886cd9f65814923b88019baf32b#diff-91f832c4094c3bd7b657665929a0d27cR74) @JoakimSM. Would that make the code more readable for you as well?

JoakimSM: Of course, if you're refactoring the rulesEngine, you could expose additional functions like executeRulesForSingleEvent, executeRulesForTEI and so on, that internally in the rulesEngine calls the generic executeRules function. That makes sense, but then maybe we should expose functions for every use case and this doesn't make the rules engine that "generic" anymore. What we're doing here is just creating this helper function and using it internally in the app instead.

JoakimSM: We talked it through and decided to add the helper methods to the rules engine.

PR #597, opened 2020-04-27
refactor: on RulesEngine class 03

paschalidi:
hey Joakim good morning! 

Opening this PR which continues what we have started with the RulesEngine. There will be few PRs following probably 3 of the same size which will finalise the refactoring. 
There are few touch points I would like to stress so that reviewing this gets easier.

I am focusing here only in simplifying the signature of the RulesEngine class. That means I mostly move around functions that dont have to exist as parameters on the RulesEngine class. One of those is for instance `dateUtils`. Instead of passing this as an argument to the functions I now import it to the files that is used. After these changes we can call `RulesEngine` class without giving it any parameters which is simplifies the code for the developer who wants to use it.

Another change is that now `executionService` is returning the effects instead of returning some effects that will go through a processor to give us the final effects. This way I aim to hide this implementation detail from the user who is calling the  `executeRules.getEffects` function.

NOTE: We have talked about having the same changes for TEI and separating the functions and having one for TEI and one  for the SingleEvents and this is still the case but will come after this PR since this is the order of the changes I have made in git and I also wanna add tests to the TEI before changing it around :)  

Hope the description helps and makes sense ☮️

### DIFF
--- a/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
+++ b/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
@@ -604,7 +604,7 @@ export default function getExecutionService(variableService) {
      * @param {*} optionSets all optionsets(matedata)
      * @param {*} flag execution flags
      */
-    return (programRulesContainer, executingEvent, evs, allDataElements, allTrackedEntityAttributes, selectedEntity, selectedEnrollment, selectedOrgUnit, optionSets, processType, flag) => {
+    function getEffects (programRulesContainer, executingEvent, evs, allDataElements, allTrackedEntityAttributes, selectedEntity, selectedEnrollment, selectedOrgUnit, optionSets, processType, flag) {
         const { programRules } = programRulesContainer;
         if (!programRules) {
             return null;
@@ -655,4 +655,5 @@ export default function getExecutionService(variableService) {
 
         return processRulesEffects(effects, processType, allDataElements, allTrackedEntityAttributes);
     };
+    return { getEffects }
 }
--- a/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
+++ b/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import log from 'loglevel';
 import isDefined from 'd2-utilizr/lib/isDefined';
 import isString from 'd2-utilizr/lib/isString';
@@ -8,7 +9,6 @@ import getDateUtils from '../dateUtils/dateUtils';
 import momentConverter from '../converters/momentConverter';
 import getRulesEffectsProcessor from '../rulesEffectsProcessor/rulesEffectsProcessor';
 import rulesEffectsValueConverter from '../converters/rulesEffectsValueConverter';
-
 
 export default function getExecutionService(variableService) {
     const dateUtils = getDateUtils(momentConverter);
@@ -591,7 +591,6 @@ export default function getExecutionService(variableService) {
 
         return effect;
     };
-
 
     /**
      *
--- a/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
+++ b/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
@@ -608,7 +608,7 @@ export default function getExecutionService(variableService, dateUtils) {
      */
     const internalExecuteRules = (programRulesContainer, executingEvent, evs, allDataElements, allTrackedEntityAttributes, selectedEntity, selectedEnrollment, selectedOrgUnit, optionSets, flag) => {
         const { programRules } = programRulesContainer;
-        if (programRules.length === 0) {
+        if (programRules.length === 0 || !programRules) {
             return null;
         }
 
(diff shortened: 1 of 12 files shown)


src/core_modules/capture-core/rulesEngineActionsCreator/rulesEngineActionsCreator.js
```diff
@@ -30,18 +31,29 @@ export function getRulesActionsForEvent(
...
-        rulesEngine,
-        program,
-        foundation,
-        orgUnit,
-        currentEventData,
-        allEventsData,
-    );
-    return getRulesActions(rulesEffects, foundation, formId);
+    const data = prepareEventData(program, foundation);
+    if (data) {
```
paschalidi, line 38: This might look a bit strange and I agree it is. I will move this piece of code as it used to be hidden in a function in the following PR. It was mostly an experiment here :) Apologies

src/core_modules/capture-core/rulesEngineActionsCreator/converters/inputValueConverter.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 /* eslint-disable class-methods-use-this */
-import { moment } from 'capture-core-utils/moment';
-import type { IConvertInputRulesValue } from 'capture-core-utils/RulesEngine/rulesEngine.types';
+import { moment } from '../../../capture-core-utils/moment';
```
paschalidi, line 3: can we keep those changes since my IDE throws always an error when in the path we use aliases?

paschalidi:
@JoakimSM 👋 

Thanks for looking into this!

You are right the complexity the dependency injection is adding here is not great. 👍 Thanks for pointing this out. As a result this PR has a different scope and is much smaller to review. 

The changes that are being introduced here are:

1. effects are being processed and returned from the executionService by a single function `getEffects`. That way we dont have to return 2 function from the executionService but one that does both the execution and the evaluation of the effects that we will then apply on the UI.

2. Some eslint changes that I caught and some unused code. These should not be here I admit but since I am touching the files and I find it that it is strange if I dont apply the changes.

src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js
```diff
@@ -646,15 +645,10 @@ export default function getExecutionService(variableService, dateUtils) {
...
 
-    return {
-        executeRules(programRulesContainer, executingEvent, evs, allDataElements, allTrackedEntityAttributes, selectedEntity, selectedEnrollment, selectedOrgUnit, optionSets, flags) {
-            if (!programRulesContainer.programRules) {
-                return null;
-            }
-            return internalExecuteRules(programRulesContainer, executingEvent, evs, allDataElements, allTrackedEntityAttributes, selectedEntity, selectedEnrollment, selectedOrgUnit, optionSets, flags);
-        },
-        convertDataToBaseOutputValue: convertRuleEffectDataToOutputBaseValue,
+        const processRulesEffects = getRulesEffectsProcessor(convertRuleEffectDataToOutputBaseValue, rulesEffectsValueConverter);
```
paschalidi, line 649: I am also realising that moving this inside the execution service is wrong since effects dont have anything to do with executionService really.

JoakimSM: I agree, I would prefer to keep the effects processing separate.

src/core_modules/capture-core/rulesEngineActionsCreator/converters/inputValueConverter.js
JoakimSM: what's wrong with your IDE? 😁 What's the error? It's no biggie, we can change it.

paschalidi: problem is with eslint. Dont you get the error as well?
image

src/core_modules/capture-core-utils/RulesEngine/rulesEffectsProcessor/rulesEffectsProcessor.js
```diff
@@ -259,22 +260,18 @@ export default function getRulesEffectsProcessor(
         effects: Array<ProgramRuleEffect>,
         processType: $Values<typeof processTypes>,
         dataElements: ?DataElements,
-        trackedEntityAttributes: ?TrackedEntityAttributes) {
+        trackedEntityAttributes: ?TrackedEntityAttributes): ?Array<OutputEffect> {
```
paschalidi, line 263: to annotate this I had to switch the filter and the map functions bellow.

paschalidi: What if we first filter for `mapActionsToProcessor[action]` and then also filter in the end for truth values?

This way we can avoid the pain of `FlowFixMe` and also have annotated the return of the function `mapProcessTypeToIdentifierName`

JoakimSM: If that for some reason helps flow, I guess that's ok.

PR #617, opened 2020-05-04
chore: adds tests 04

paschalidi:
based on https://github.com/dhis2/capture-app/pull/597

Hey @JoakimSM here is another bit regarding the RulesEngine. 

This PR 
1. Tests are added to validate that we dont break the functionality of the existing code.
2. The [events object](https://github.com/dhis2/capture-app/pull/617/files#diff-954329364f78f8c8195f13957aef04feL73) is moved out from the `executeRules` function to the preparation phase of the `runRulesForSingleEvent` and its own function called `getEventsData`. This way when someone sees in the class `RulesEngine` they only see how the effects are being generated.

I hope the above make sense. Let me know your thoughts! ☮️


src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -51,19 +38,34 @@ function getRulesEngineDataElementsAsObject(
...
-    }
-
     if (program instanceof EventProgram) {
         dataElements = getEventDataElements(program);
     }
 
     return getRulesEngineDataElementsAsObject(dataElements);
 }
+
+// todo do we need to separate events by stage at this point when we dont have stages in the product?
```
paschalidi, line 48: @JoakimSM can you please have a look on this comment?

JoakimSM: Agreed, it's redundant. Let's keep the byStage object though, so we don't break anything.

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -51,19 +38,34 @@ function getRulesEngineDataElementsAsObject(
...
+            }
+            return accEventsByStage;
+        }, {});
+
+        return { all: eventsData, byStage: eventsDataByStage };
+    }
+    return null;
+}
+
 function prepare(
```
JoakimSM, line 65: I like the added abstraction 🏆

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
paschalidi: agreed.

src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForSingleEvent.js
```diff
@@ -51,19 +38,26 @@ function getRulesEngineDataElementsAsObject(
...
     }
 
     return getRulesEngineDataElementsAsObject(dataElements);
 }
+
+// historically this function used to take care of event by stage and
+// after some refactoring `byStage` is still here to avoid introducing bugs
+function getEventsData(eventsData: ?EventsData) {
+    if (eventsData && eventsData.length > 0) {
+        return { all: eventsData, byStage: {} };
```
JoakimSM, line 52: We still need event data in the byStage object to be sure to not break anything because some rules variables rely on data from here (if you haven't change anything fundamentally in the rules engine). Looking at the old code I can see that the logic for creating the byStage object was earlier in the executeRules function of the RulesEngine. That's the reason why it was generically written (and we will need to reintroduce that code when we get to events for TEIs, but maybe not in the same place).

paschalidi: True it was before written in the in the RulesEngine. 

I can see that when we call the [`rulesEngine.executeRules`](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/rulesEngineActionsCreator/runRulesForTEI.js#L72) from the `runRulesForTEI.js` the events we are passing the events as `null`. 

So you are right we might have to reintroduce it when the TEI has also events but as for now do you see any advantages of having this logic in the `RulesEngine`? 


Could you please help me understand what are you suggesting??

JoakimSM: We still need data in the byStage object. We have two options:
1. We continue what you've started and add data to the byStage object in runRulesForSingleEvent (where it currently sits). We worry about the tracker part later (we need the approarch we had earlier for the tracker part)
2. Use the generic approach in the RulesEngine we had earlier. This will always create the correct byStage object for us based on the eventsdata. The calling function won't have to worry about the byStage part, but the code, when we only have one event, will be a tiny bit more complex then it needs to be.

Curiuos on which approach you will choose 🙂 If anything is unclear, we can discuss on Slack.

paschalidi: After our discussion I realised I was off removing this bit of code here. 

I thought there is no use case in our application at the moment for `byStage` which after our discussions on slack I realised is not the case. 

went on and added the functionality as it used to be in the `RulesEngine` before. 

I would suggest we keep this logic here and when we have the case that a TEI has events we can address the issue then. 

Is that fine with you @JoakimSM ?

paschalidi:
Same https://github.com/dhis2/capture-app/pull/644#issuecomment-631363596

PR #630, opened 2020-05-05
refactor: on execution service 10

paschalidi:
Hey @JoakimSM hope you having a good day over there in Oslo! 🇳🇴 
This is the last bit of the refactoring. 

## What is this PR about?
So in this PR there are few things that take place. 

The most important thing is that now the `executionService.js ` is now doing only execution of the functions by getting three arguments in the so called `executeExpression` function. The first argument is the `D2Functions` and the second argument is the `expression`. The third argument is a function that logs errors. 

That said you can now see that I have moved all the logic regarding the rules and the effects to the class `RuleEngine`. Which means that the generating the effects and then processing the effects is now part of the RulesEngine class and not of the executionService. This way, the `executionService.js` is agnostic to the rest of our application.

Also you will see that instead of having this massing `if else` statement in execution service we have now extracted the functions into its own file and we pass it as an argument to the `executeExpression` function. This way the function and the executeExpression are not bounded to each other. You can imagine for example that someone else could have used the `executeExpression` with their own functions if they wanted to.

Lastly you can see that I annotated with `flowjs` ❤️ 

Please give me your thoughts and comments on the changes and in case something is unclear I would love to go over it!


paschalidi:
Need to draft this found a bug

paschalidi:
@JoakimSM this must be now ready for you to have a look. @Bekkalizer feel also free to have a look. I am happy to explain what is going on and the reasoning behind the changes here. 

Also the advantages that one might have by going that way.

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
```diff
@@ -16,80 +17,380 @@ import type {
...
+    dataElements: ?DataElements,
+    trackedEntityAttributes: ?TrackedEntityAttributes
+  )=> ?OutputEffects;
+
+  constructor() {
+      const valueProcessor = new ValueProcessor(inputValueConverter);
+
+      this.dateUtils = getDateUtils(momentConverter);
+      this.variableService = new VariableService(valueProcessor.processValue);
+      this.processRulesEffects = getRulesEffectsProcessor(
```
JoakimSM, line 334: Can we make dateUtils, variableService and processRulesEffects static?

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
```diff
@@ -16,80 +17,380 @@ import type {
...
+          convertRuleEffectDataToOutputBaseValue,
+          rulesEffectsValueConverter,
+      );
+
+      this.generateEffects = (variablesHash, programRules, dataElements, trackedEntityAttributes) => {
```
JoakimSM, line 339: Is there a good reason why generateEffects is defined in the constructor like this? Could we make it a static (depends on the comment above) or a prototype method?

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
```diff
@@ -16,80 +17,380 @@ import type {
...
+          const dhisFunctions = d2Functions(this.dateUtils, this.variableService, variablesHash);
+          const effects = getEffects(dhisFunctions, programRules, dataElements, trackedEntityAttributes, variablesHash);
+          updateVariableHashWhenActionIsAssignValue(effects, variablesHash);
```
JoakimSM, line 342: You have rules depending on the result of other rules (assign action) so updating the variableHash after getting all the effects wouldn't work I think. Maybe I'm missing something though, but this seems suspicious.

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
paschalidi: Yeah thanks for pointing out. I havent realised!

src/core_modules/capture-core-utils/RulesEngine/RulesEngine.js
paschalidi: About this. Lets discuss it. One thing is that I have written tests for this case, and I have also tested it manually after importing the COVID rules but that doesnt mean that it cannot be broken.

Basically if you see [here](https://github.com/dhis2/capture-app/pull/630/files#diff-9c101046a67bb3c0f64cf485a7fa0fddL645) these 3 lines what they do is:
1. the rule is being checked if it is effective.
2. then you iterate though the `programRuleActions` and as a result you will have the `ruleEffects `

While this is fine if you take a closer look in the line [576 of the `executionService`](https://github.com/dhis2/capture-app/blob/cp%2Frefactors-rules-engine%2F09-writes-tests-on-functions/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js#L576) you will see that this `if` statement there has a little to do with the getting the effects. What it basically does is updating the `variablesHash` if there is an action called `ASSIGN`.  
In my point of view, the lines from [576](https://github.com/dhis2/capture-app/blob/cp%2Frefactors-rules-engine%2F09-writes-tests-on-functions/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js#L576) to [588](https://github.com/dhis2/capture-app/blob/cp%2Frefactors-rules-engine%2F09-writes-tests-on-functions/src/core_modules/capture-core-utils/RulesEngine/executionService/executionService.js#L588) have nothing to do with the function that is being called `getEffects`. 

What I did is I took these lines out of there and made it its own function.

So now instead of having a function that returns some effects _and_  updates the `variablesHash` when action is equal to `ASSIGN`, you now first get all the effects from the function called `getEffects` and then you iterate through the effects[1] and if there is an effect with an action that is equal to `ASSIGN` then you update the `variablesHash`

[1] so yeah now we are iterating though the effects twice instead of once but we have separated the concepts and we made it easier for the reader to understand what is going on and in which order things are taking place. I dont wanna argue about performance here since if you ask me it needs to be millions of effects for this to be a concern.


Does this explains a little bit better why I have done what I have done over there?

JoakimSM: Don't get me wrong, I like the separation of the concepts.

> So now instead of having a function that returns some effects and updates the variablesHash when action is equal to ASSIGN, you now first get all the effects from the function called getEffects and then you iterate through the effects[1] and if there is an effect with an action that is equal to ASSIGN then you update the variablesHash.

The problem I see is that one program rule in the chain could rely on an updated value in the variableHash from an earlier rule in the chain. Say you have two rules. The first one assigns a value to a variable. In the second rule you're checking the value of this variable in the expression. If the variableHash is updated post execution (after getEffects) the variable value would be wrong when running the second rule.

paschalidi: AH!

paschalidi: Might be true what you saying. How can we test that scenario. That's the question.

paschalidi:
@JoakimSM I corrected the point you mentioned. Thanks again for the review.

In the meantime I realised there were few other things that needed to be fixed. Check the commit history to see what I mean ;)

paschalidi:
Look here: https://github.com/dhis2/capture-app/pull/644#issuecomment-631363596

I intent to open a PR that will fix all the `replace`ments.

PR #644, opened 2020-05-07
fix: Syntax for d2:hasValue() function at 'Tracker Capture' and 'Capture' are not aligned.  11 

paschalidi:
fixes: [ticket](https://jira.dhis2.org/browse/DHIS2-8503)

Hey @JoakimSM after all the refactoring finally some fixes! 

## How is this PR solving our bug?
Here  we can see that we want the function to be called with `d2:hasValue( #{age} )`. 
According to the order that the execution of the code takes place ( and used to take before the refactoring ) what happens is that the `#{age}` is being replaced with its value if the value exists in the `variablesHash` and then the execution of the `d2:hasValue` takes place. 

That means that when the function is being called we dont have anymore to look of the age exists in the `variablesHash` but what we are doing now - since the variable has been already replaced with its value - is to check if the variable is truthy and return accordingly. 

You can also see that the tests have been updated to mirror this change :) 

Questions and thoughts are more than welcome!


Bekkalizer:
The solution here would almost work, and I think there is one problem with it. You can test to verify, but the following my concern:
When a numeric field or a boolean field does not have a value, the variables hash will contain either "0" or "false". This would not be interpreted correctly in this case I would think.

The TC rules engine was updated so that the input to specific functions was not evaluated/translated. Have a look here:
https://github.com/dhis2/d2-tracker/blob/5222c622bd65a4e4760454eb23d9fc53458ff6b3/src/dhis2.angular.services.js#L1967

paschalidi:
@Bekkalizer that makes sense. will have a closer look at it and cover the cases that you recommended with tests.

paschalidi:
okay I looked into in what you suggested. 

The initial values of Boolean and Numeric fields are `''`. I can see that this might bring or even bugs later on. One way to avoid those is to cover with test the initial cases of all the functions. This way if someone in the future changes the initial values from `''` to be something else our tests will fail and therefore we will know that something broke.

Before we find a good way to solve this problem lets make sure we are on the same page about functionality @Bekkalizer. I have talked with Yuri already about it but just to make sure once more. 

When _any_ field has a value that is set by the user (even if the cases value is 0 for Numeric fields or `false` for Boolean fields) then the `hasValue` function returns `true` as following:
image

In the case that the values is not set by the user then the `hasValue` would be returning `false`.

paschalidi:
This has now merged changed of the branch that was based on it. That is usual when you are chaining PRs with branches. I merged the changes just after having it approved. When all the PRs are approved then these will be merged on the opposite order (from last to first) till we reach `master`.

paschalidi:
https://github.com/dhis2/capture-app/pull/644#issuecomment-631357966

What is this saying here is that we replace 4 times the `{` left bracket but only once the right bracket. A solution to this would be something between the lines of `string.replace(/[A#CV]{/g, "'")`.

PR #665, opened 2020-05-15
refactor: 2. add context for metadata loading [TECH-344]

JoakimSM:
The second part of [TECH-344](https://jira.dhis2.org/browse/TECH-344)

This part adds a context to the metadata loader. The context consist of the the query function from the data engine and some other dependencies that will be used in the loading logic coming in the following PRs.

The context was added primarily to avoid passing the query function around to functions deep in the call stack, the storage controllers were added as a result of planning ahead (want to get rid of the global variable, using the React context instead). The objectStores might have been a little bit unnecessary, but I don't think the objectStores adds any noteable complexity. The benefit, I believe, is that you should be able to test the entire metadata loading process by mocking the context only.

Let me know your thoughts!


src/core_modules/capture-core/metaDataStoreLoaders/baseLoader/loadWithContext.js
```diff
@@ -0,0 +1,15 @@
+// @flow
```
paschalidi, line 1: Why dont we have this being exported from the `/baseLoader/metaDataLoader.js` and we create another file? 

Instead of exporting then `loadMetaData` from the `/baseLoader/metaDataLoader.js` we would export something like `loadMetaDataWithContext`. Is there any advantage by adding this file?

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
```diff
@@ -0,0 +1,63 @@
...
+) => {
+    context = {
+        onQueryApi,
+        storageController,
+        parentStorageController,
+        storeNames,
+        parentStoreNames,
+    };
+    await callback();
+    context = null;
```
paschalidi, line 55: Could you please explain what is the reason to assign `null` here?

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+import { StorageController } from 'capture-core-utils/storage';
+import type { QueryApiFn } from '../loader.types';
+
+let context;
+
+type StoreNames = {
```
paschalidi, line 7: Could we also move those types on a separate file? I think its a good think to be doing and we end up having more clear files in the end. 

What do you think on that?

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
```diff
@@ -0,0 +1,63 @@
...
+    REDUX_PERSIST: string,
+};
+
+type ParentStoreNames = {
+    USER_CACHES: string,
+    STATUS: string,
+    SYSTEM_SETTINGS: string,
+};
+
+type Input = {
```
paschalidi, line 30: What about `ContextInput` instead of `Input` ?

src/core_modules/capture-core/metaDataStoreLoaders/context/index.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+export { provideContext, getContext } from './context';
```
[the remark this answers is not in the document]
JoakimSM: sorry, you will have to look in the next PR

src/core_modules/capture-core/metaDataStoreLoaders/index.js
```diff
@@ -1,3 +1,3 @@
 // @flow
-export { default as loadMetaData } from './baseLoader/metaDataLoader';
+export { loadWithContext as loadMetaData } from './baseLoader';
```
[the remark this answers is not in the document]
JoakimSM: Thumbs up, absolutely agree. Renaming.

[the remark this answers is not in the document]
JoakimSM: On second thought, I will rename the file loadWithContext to loadMetaData and loadMetaData to loadMetaDataInternal and expose the loadMetaData function from loadMetaData.

Temporary will use filename loadMetaDataMain for loadMetaData to avoid some merge trouble and fix it in the last PR.

src/core_modules/capture-core/metaDataStoreLoaders/baseLoader/loadWithContext.js
JoakimSM: I will rename the files (loadWithContext -> loadMetaData, loadMetaData -> loadMetaDataInternal). The advantage is separation of concerns, keeping the context importing and building separate.

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
JoakimSM: If we don't, the context will still retain the object references after the function returns (because the context variable is part of the outer module scope).

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
JoakimSM: Sure. What I've been doing is extracting them to a new file when the types are required in multiple files. But I agree it's cleaner if we always keep them in a separate file.

paschalidi: That makes sense. Lets then leave them here since its not being reused

src/core_modules/capture-core/metaDataStoreLoaders/context/context.js
paschalidi: Thanks for sharing the knowledge.

src/core_modules/capture-core/metaDataStoreLoaders/index.js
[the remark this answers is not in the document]
paschalidi: This will be followed up in the last PR do understand you well?

src/core_modules/capture-core/metaDataStoreLoaders/baseLoader/loadWithContext.js
paschalidi: That makes sense

src/core_modules/capture-core/metaDataStoreLoaders/index.js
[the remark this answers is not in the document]
JoakimSM: yes

PR #675, opened 2020-05-16
fix: 6. improve metadata fetching [TECH-344]

JoakimSM:
Last part of [TECH-344](https://jira.dhis2.org/browse/TECH-344)

Another quite large one, sorry about that. The main objective of this PR is to ensure that all metadata can somehow be updated. Currently we only have the program version and option set version to work with. For metadata that is not tied to a program (like constants for example) the metadata will be refetched when a program version is increased, no matter which one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "capture-app",
   "homepage": ".",
-  "version": "34.1.0",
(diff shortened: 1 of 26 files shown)


JoakimSM:
Closing this until we settle on the previous ones.

PR #709, opened 2020-05-25
refactor: 7. cleaning up initialization [TECH-344]

JoakimSM:
Part seven of [TECH-344](https://jira.dhis2.org/browse/TECH-344)

Mostly moving/renaming files. Moving AppLoader to the components folder.

Should address:
1. [#664(comment)](https://github.com/dhis2/capture-app/pull/664#discussion_r426420886)
2. [#664(comment)](https://github.com/dhis2/capture-app/pull/664#discussion_r426465595)


src/components/App/index.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+export { default as App } from './App.component';
```
paschalidi, line 2: Great  👍

src/epics/trackerCapture.epics.js
```diff
@@ -185,7 +185,7 @@ import {
 import { saveAssigneeEpic } from 'capture-core/components/Pages/ViewEvent/RightColumn/AssigneeSection';
 
 import { loadCoreEpic } from 'capture-core/init';
-import { triggerLoadCoreEpic, loadAppEpic, loadCoreFailedEpic } from '../init/entry.epics';
+import { triggerLoadCoreEpic, loadAppEpic, loadCoreFailedEpic } from '../entry';
```
paschalidi, line 188: 👍 zuper!

src/locales/en/translations.json
```diff
@@ -1,5 +1,8 @@
 {
-    "Please provide a valid date": "",
+    "Choose one or more dates...": "",
```
paschalidi, line 2: Will have to ask again, how this is populated here?  I mean is it appearing when you run `yarn start`?

src/entry/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+export { triggerLoadCoreEpic, loadAppEpic, loadCoreFailedEpic } from './entry.epics';
+export { actionTypes } from './entry.actions';
```
paschalidi, line 3: I am gonna raise this argument here just to discuss it and see where we stand. It's basically out of the scope of this PR. 
I also know in the whole project we are exporting all the action types as `actionTypes` and when multiple of them are in the same file then we rename them as for example we do [here](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js). 

My question is what if we would export them with a unique name so that we would be able to search around the project using their unique names? 
Now if I would like to search some types it might be difficult to find them. Another thing is that one `actionTypes` might have been renamed in one place differently than in an another one using `as`. I am worried that using different names might end up a little bit tricky in the long term. 

Again just food for thought. Let me know what you think

src/locales/en/translations.json
JoakimSM: yes, yarn start and build. The file was modified now because by moving files around the translation order changed (I think)

src/entry/index.js
JoakimSM: Thanks for bringing this up. I've been swaying back and forth over this. I think I'm leaning the same direction as you in hindsight. Added the topic to our next Tracker technical discussion board meeting.

paschalidi: Great stuff thanks a lot! Let me know if I could help with that :)

PR #710, opened 2020-05-25
refactor: 8. Importing using created index files [TECH-344] 

JoakimSM:
Part eight of [TECH-344](https://jira.dhis2.org/browse/TECH-344)

This basically uses the created index files for importing.

Should resolve:
1. [#664(comment)](https://github.com/dhis2/capture-app/pull/664#discussion_r425883746)
2. [#664(comment)](https://github.com/dhis2/capture-app/pull/664#discussion_r425888854)


JoakimSM:
Thanks for the feedback @paschalidi.

I was about to do the same for the other files in the loading masks folder, but I saw so many problems and I want to postpone it for another issue (some I didn't want keep, others needed a huge refactor etc).

paschalidi:
I am not suggesting to do the refactor now but only to export them from the index for now and keep the refactoring for later step?

Is that not sensible?

JoakimSM:
Fair enough 🙂 I'll do it as long as you don't expect refactor of the exported files 😉

paschalidi:
Thanks a lot 🙇 !

JoakimSM:
Exporting all LoadingMasks from the index file and using them. Please have a look.

PR #711, opened 2020-05-26
refactor: moving the engine to capture-core

paschalidi:
Hey @JoakimSM as we discussed I am opening this last PR regarding the RulesEngine. 🎉
Please have a look and let me know your thoughts! 

### What is this PR about?
Moves the RulesEngine files from the `capture-core-utils` to `capture-core`. 
Also I followed the rule of the imports we talked about. 

_If there is any import that is happening outside of the module - in this case `RulesEngine` or `RulesEngineActionsCreator` - it is being imported from the `index` file of the module._

As we discussed and correct me if I am wrong we would like to be doing this so that the `index` would act as the interface for the rest of the app. 

### What is this PR changing?
There are no functional changes. Only moving the files and correcting the imports. However a lot of files are moved so it would be great if we can create a ticket for this so that it can be tested from the QA. Shall I create one or how is this being done in the past??

The file structure is also up to discussion. Now the structure of the file is capturing the whole feature. That means everything related to the `rules` is in the `rules` file. My inspiration for this was the `events` and the `enrolment` folder.

```
    .
    ├── ...
    ├── capture-core                   
    │   ├── rules                                    # all the topics related to the rules
    │        └──── actionsCreator      # all the topics related to the actionsCreator
    │        └──── engine                    # all the topics related to the engine
    ...
```

On another point, I am deleting the `rulesEngine/rulesEngine.middleware.js` since it is not being used at the moment if I am not wrong.


src/core_modules/capture-core-utils/RulesEngine/index.js
```diff
@@ -1,4 +0,0 @@
-// @flow
```
paschalidi, line 1: This file has simply moved is not removed. You will see it later. :)

src/core_modules/capture-core/components/DataEntry/actions/dataEntryLoadEdit.actions_old.js
```diff
@@ -8,7 +8,7 @@ import { convertValue } from '../../../converters/clientToForm';
 import { errorCreator } from 'capture-core-utils';
 import { actionCreator } from '../../../actions/actions.utils';
 import { addFormData } from '../../D2Form/actions/form.actions';
-import { getRulesActionsOnLoad, getRulesActionsOnLoadForSingleNewEvent } from '../../../rulesEngineActionsCreator/rulesEngineActionsCreatorForEvent';
+import { getRulesActionsOnLoad, getRulesActionsOnLoadForSingleNewEvent } from '../../../rules/actionsCreator';
```
paschalidi, line 11: this file looks like is not being used since this imports dont exist and it didnt also before this PR.

Would you like to introduce me to the history of this @JoakimSM?

src/core_modules/capture-core/rules/actionsCreator/runRulesForTEI.js
```diff
@@ -1,17 +1,17 @@
...
+import constantsStore from '../../metaDataMemoryStores/constants/constants.store';
+import optionSetsStore from '../../metaDataMemoryStores/optionSets/optionSets.store';
 
 import type {
     ProgramRulesContainer,
     TrackedEntityAttribute as TrackedEntityAttributeForRulesEngine,
     Enrollment,
     TEIValues,
-} from '../../capture-core-utils/RulesEngine/rulesEngine.types';
+} from '../engine/rulesEngine.types';
```
paschalidi, line 14: lets discuss about those, since it is not clear to me if these imports are from the same module or form a different module.

To me it looks the whole `rules` is its own module. What do you think?

src/core_modules/capture-core/components/DataEntry/actions/dataEntryLoadEdit.actions_old.js
paschalidi: Gonna go on and delete this one while doing the lgtm fixes :) I will follow up on this.

PR #713, opened 2020-05-28
refactor: moves engine to capture-core

paschalidi:
Check this back port Joakim. 

I still intent to create that ticket for QA so that we make sure this change didnt break anything.


paschalidi:
Ticket has been created https://jira.dhis2.org/browse/DHIS2-8945

PR #721, opened 2020-06-04
chore: readding eslint-import-resolver-webpack

JoakimSM:
Hey @paschalidi 

I'm readding the `eslint-import-resolver-webpack` dependency. Without it the import statements using aliases for `capture-core-utils`, `capture-core` and so on throws an eslint error.


paschalidi:
AH! Thanks for that. Didnt realise I have broken it. Apologies! You can also ask me to fix those things that I break :)

PR #722, opened 2020-06-05
fix: DHIS2-8648 force ltr for header bar

JoakimSM:


src/components/App/AppContents.component.js
WalterTwijukye, line 1: Is this supposed to replace the current one ?

Still fighting to handle this 

Cannot read properties of undefined (reading 'map')
TypeError: Cannot read properties of undefined (reading 'map')
    at http://localhost:3000/static/js/bundle.js:41522:74
    at updateMemo (http://localhost:3000/static/js/bundle.js:106405:23)
    at Object.useMemo (http://localhost:3000/static/js/bundle.js:106821:20)
    at useMemo (http://localhost:3000/static/js/bundle.js:122544:25)
    at HeaderBar (http://localhost:3000/static/js/bundle.js:41520:62)
    at renderWithHooks (http://localhost:3000/static/js/bundle.js:105856:22)
    at updateFunctionComponent (http://localhost:3000/static/js/bundle.js:107809:24)

PR #738, opened 2020-06-08
 refactor: DHIS2-8922 implements a shared component locked selector 

paschalidi:
Hey @JoakimSM ! 👋 Check this out 👨‍💻 

### What is this PR about?
After our discussion I went on and put together a component called LockedSelector. 
There is no new code to it I simply reused to code you have written. The only difference is that the epics and the reducer logic is now shared. This way when we would like to create a new page as I did with the `search` we could now reuse the component. 

I have thoroughly testing this manually and made sure the functionality is the same as it used to be. To avoid that I have missed something therefore I will create a ticket so that QA can also test since they always doing a good work on testing the features. 

```
The scenarios I have tested are: 

* Open the main page
1. and select both orgUnit and program from the LockedSelector
2. while orgUnit is preselected from the url and select the program
3. while program is preselected from the url and select the orgUnit
4. and click the New button
5. select orgUnit and program and click start again.
6. select orgUnit and program and click start New Event.
7. from the url add wrong program
8. from the url add wrong orgUnit

* Open the NewEvent page
1. and select both orgUnit and program from the LockedSelector
2. while orgUnit is preselected from the url and select the program
3. while program is preselected from the url and select the orgUnit
4. and click the cancel button on the bottom
5. select orgUnit and program and click start again.
6. select orgUnit and program and click start New Event.
7. while orgUnit and program are selected try to edit the form and then click start again you should see a modal.
8. from the url add wrong program
9. from the url add wrong orgUnit

* Open the viewEvent page
1. from the url
2. and then click start again.
3. from clicking to an event on the table.
4. try to edit the event and then click start again you should see a modal. 
5. from the url with wrong id.
```
Let me know if I missed anything! When we have cypress I will test those scenarios there as well ;) 

Two last points:
1. I am suggesting we discuss it first with a meeting before you jump into reviewing this. 
2. Lets discuss if we really want to have this `activePage` reducer. I think its valuable to get this straight. In this PR you can see how it is being used. As you said it might be that we wont like to have it at all or that we want to have it only for the loading and the errors. Lets take the time in this and have an open discussion! 

Please let me know your thoughts on this!


paschalidi:
@JoakimSM as we talked I have looked into the things you reported the other day. Let me know if you would like to go through this together to speed up the process ;)

paschalidi:
@JoakimSM I looked into the comments we have discussed. Have a final look :)

PR #744, opened 2020-06-09
[DHIS2-8332] fix: built in variable V{orgunit code} 

paschalidi:
Hey @JoakimSM this ticket got returned back since the previous fix was showing the name instead of the code. Ticket: https://jira.dhis2.org/browse/DHIS2-8332

### What is this PR about? 
When data is being fetched I am adding the `code` to redux as well. This way the code is now available whenever the rules engine runs. I have updated the tests as well. 

Last thing is that while I was looking to fix that I found a file that is not used. I removed it from the codebase. 

That is all I am doing here. Please have a look and then we can go on and back-port


paschalidi:
@JoakimSM good point thanks a lot! I have fixed the bug you spotted. Wanna have another look?

PR #761, opened 2020-06-17
refactor: moves components with dependencies to the `capture-core` components

paschalidi:
DRAFT


paschalidi:
closes in favour of opening the same PR from branch that has the latest changes of the current chain. This is because of summer time break.

PR #762, opened 2020-06-17
chore: TECH-401 update version of rxjs to v6

paschalidi:
Ticket: https://jira.dhis2.org/browse/TECH-401

@JoakimSM I had some extra time in between the things so I worked on updating rxjs to v6. The rxjs team is already preparing the alpha of the version 7. I think you might agree that we wont want to be left that back behind. :) 

### What is this PR doing? 
I updated the `rxjs` and `redux-observable` to their latest stable version. This came with the update of their flow types. 

#### `rxjs`
One good that came with the `rxjs` update we got rid of many `FlowFixMe` comments that would throw an error when we used `action$.ofType`. Also in the version 6 of `rxjs` we start using `pipe`. Basically nothing changed in how observables would work the only thing that is changed is some syntax. You can read more on pipe [here](https://rxjs.dev/guide/v6/pipeable-operators). To update to `rxjs` version 6 I have followed the instruction from the [ official migration guide](https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md)
#### `redux-obervables`
Updating `redux-obervables` came with the redundancy of the function `getState` now we can access the store in the `epics` by using the `store.value`. 

### Testing
I have tested manually all the functionality and everything seem to work as expected but please have a look your self in case I have missed or miss understood anything ;) And as always bring your comments in 👍


src/core_modules/capture-core/components/Connectivity/connectivity.epics.js
```diff
@@ -1,23 +1,25 @@
...
+import { ofType } from 'redux-observable';
+import { map, filter } from 'rxjs/operators';
+
 import { getEventListOnReconnect, goingOnlineExecuted, batchActionTypes } from './connectivity.actions';
 
 const OFFLINE_STATUS_CHANGED = 'Offline/STATUS_CHANGED';
 
 export const goingOnlineEpic = (action$: InputObservable, store: ReduxStore) =>
-
-    // $FlowFixMe[prop-missing] automated comment
```
paschalidi, line 9: One good that came with the rxjs update we got rid of all those comments

src/core_modules/capture-core/cleanUp/cleanUp.js
```diff
@@ -11,7 +11,7 @@ function getActionToCleanUpInLoadingList(state: ReduxState) {
 
 export default function cleanUp(store: ReduxStore) {
     const cleanUpActions = [
-        getActionToCleanUpInLoadingList(store.getState()),
```
paschalidi, line 14: updating `redux-obervables` came with the redundancy of the function `getState` now we can access the store in the `epics` by using the `store.value`

PR #775, opened 2020-06-20
01 fix: TECH-399 suppressed all flow errors 

paschalidi:
Hey Joakim this is the PR that suppresses all the flow errors. The ticket is https://jira.dhis2.org/browse/TECH-399

So now we on 0 errors  🎉 (excluding the packages we import tho). That means we can then start fixing errors we are having and more important we wont be introducing more errors accidentally.

### What is happening in this PR and why there are 432 files changes 😨 

I have used [this](https://github.com/facebook/flow/tree/master/packages/flow-dev-tools) tool made by the `flow` team. Its not perfect but did the job. 

Basically what I did is that I run two commands. 
The first one automatically removed all the suppressions that were not necessary. Lets say because that at some point a lib was throwing an error which we suppressed however after updating few versions either of `flow` itself or of the lib the errors has been fixed. 
```
../flow/tool remove-comments --bin node_modules/.bin/flow --check check .
```

The second one adds comments to all the errors that `flow check`  reports. 
```
../flow/tool add-comments --all --comment "automated comment" --bin node_modules/.bin/flow -c check ./src

```

Bring your thoughts and concerns in! Have a good day :)


.flowconfig
```diff
@@ -10,8 +10,6 @@ flow/
 src/core_modules/capture-core/flow/
 
 [options]
-suppress_comment= \\(.\\|\n\\)*\\$FlowSuppress
```
paschalidi, line 13: looks like according to this we dont need those https://github.com/facebook/flow/issues/8403#issuecomment-644942810

PR #800, opened 2020-06-29
01 feat: DHIS2-8924 Implements the SearchDomainSelector component

paschalidi:
Hey 👋  @JoakimSM this is closing the https://jira.dhis2.org/browse/DHIS2-8924

So basically not much happening here. Just adding a new component the `SearchDomainSelector` one.


src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
```diff
@@ -0,0 +1,53 @@
...
+const mapStateToProps = (state: ReduxState): PropsFromRedux => {
+    const { currentSelections, activePage } = state;
+
+    const trackedEntityTypesWithCorrelatedPrograms =
+      [...programCollection.values()]
+          .filter(program => program instanceof TrackerProgram)
+          .reduce((acc, {
+              id: programId,
+              name: programName,
+              trackedEntityType: { id: trackedEntityTypeId, name: trackedEntityTypeName },
```
paschalidi, line 18: I am getting a flow error here that I dont like much. Can you please help me fix it @JoakimSM?

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
```diff
@@ -0,0 +1,53 @@
...
+import { SearchPage as SearchPageComponent } from './SearchPage.component';
+import { programCollection } from '../../../metaDataMemoryStores';
+import { TrackerProgram } from '../../../metaData/Program';
+import { withErrorMessageHandler, withLoadingIndicator } from '../../../HOC';
+import type { PropsFromRedux } from './SearchPage.types';
+
+const mapStateToProps = (state: ReduxState): PropsFromRedux => {
+    const { currentSelections, activePage } = state;
+
+    const trackedEntityTypesWithCorrelatedPrograms =
```
JoakimSM, line 12: Maybe we should use memoization on this? For the solution I think I prefer React useMemo now that we're using hooks, but reselect would work also.

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
JoakimSM: Hmm, flow doesn't pick up the result from the filter function always being a tracker program. This is a recurring problem, i.e. I don't have a good solution for now.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -0,0 +1,103 @@
...
+                        />
+                    }
+                >
+                    <div className={classes.searchRow}>
+                        <div className={classes.searchRowTitle}>Search for</div>
+                        <div className={classes.searchRowSelectElement}>
+                            <SingleSelect
+                                onChange={({ selected }) => { choseSelected(selected); }}
+                                selected={selectedOption}
+                                empty={<div className={classes.customEmpty}>Custom empty component</div>}
```
JoakimSM, line 73: Since there is no optimization inside the SingleSelect component I guess it's ok to add the jsx inline like this for now.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -0,0 +1,103 @@
...
+                            >
+                                {
+                                    Object.values(trackedEntityTypesWithCorrelatedPrograms)
```
JoakimSM, line 76: I think I would prefer to use React useMemo for this options calculation (I can see that the SingleSelect component doesn't use any kind of optimization (pure component, useMemo etc) so it wouldn't help to avoid reconciliation, but at least building the array with SingleSelectOptions would only happen once.

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
paschalidi: We want  to memoize it in the container or in the component? Can we use `useMemo` in the container?

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: So that is not a blocker for you right?

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -0,0 +1,103 @@
...
+    },
+    searchRowSelectElement: {
+        width: '100%',
+        marginRight: 8,
+    },
+});
+
+
+const Index = ({ classes, trackedEntityTypesWithCorrelatedPrograms, preselectedProgram }: Props) => {
+    const [selectedOption, choseSelected] = useState(preselectedProgram);
```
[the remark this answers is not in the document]
paschalidi: Yea!

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
paschalidi: Hm yeah. Strange `flow`. Can you explain to me -coz I dont get it- how is it that a `programmCollection` which is typed as ```Map<string,Program>``` can end up being `EventProgram` or `TrackerProgram`?

paschalidi:
@JoakimSM looked into the comments. Thanks for looking into it 👍 One comment we need to discuss about https://github.com/dhis2/capture-app/pull/800#discussion_r449435834

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
JoakimSM: `EventProgram` and `TrackerProgram` is a child class of `Program`. The collection contains programs of some kind.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
JoakimSM: perfect!

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
paschalidi: Got it thanks!

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
paschalidi: @JoakimSM isn't it enough that we are using `useMemo` on the `trackedEntityTypesWithCorrelatedPrograms` here? 
https://github.com/dhis2/capture-app/pull/800/files#diff-cd18d8025db724faaf2850624427ce57R71

paschalidi: I am already preparing a branch that will include this improvement. Let me know if this is fine with you :)

PR #802, opened 2020-06-30
fix: DHIS2-7593 year selection in date picker

JoakimSM:
JIRA-issue: [DHIS2-7593](https://jira.dhis2.org/browse/DHIS2-7593)

This PR enables a year selector in the date picker. I've forked the underlying react-infinite-calendar be to able to make it work.  This component is far from perfect and there are many things I would like to change, but @Bekkalizer and I decided this is what we will do at this point in time. In the future, when a date picker is built in the dhis2 ui library, we will git rid of this component (hopefully sooner than later).


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "capture-app",
   "homepage": ".",
   "version": "35.0.0",
-  "cacheVersion": "1",
+  "cacheVersion": "2",
```
JoakimSM, line 5: I had to add some data to the cache. Increasing the cache version ensures the indexedDB cache is rebuilt.

src/core_modules/capture-core/components/FormFields/DateAndTime/D2Date/D2DateCalendar.component.js
```diff
@@ -140,6 +138,7 @@ class D2DateCalendar extends Component<Props> {
             <div
                 className={classes.container}
             >
+                { /* $FlowFixMe */ }
```
JoakimSM, line 141: After one of the flow js upgrades, flow has started complaining about "missing" props. These props aren't really missing, but rather passed on through our "passOnProps" object. I don't want to annotate props for every level where they are just passed on, but not used (a "interface" level where every prop is listed could make sense). Suppressing these errors for now, we should upgrade flow at some point and try to find a viable solution to this problem.

src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storePrograms.js
```diff
@@ -92,9 +92,9 @@ const fieldsParam = 'id,version,displayName,displayShortName,description,program
 'organisationUnits[id,displayName],' +
 'userRoles[id,displayName],' +
 'programStages[id,access,displayName,description,executionDateLabel,formType,featureType,validationStrategy,enableUserAssignment,dataEntryForm[id,htmlCode],' +
-'programStageSections[id,displayName,sortOrder,dataElements[id]],programStageDataElements[compulsory,displayInReports,renderOptionsAsRadio,renderType[*],' +
+'programStageSections[id,displayName,sortOrder,dataElements[id]],programStageDataElements[compulsory,displayInReports,renderOptionsAsRadio,allowFutureDate,renderType[*],' +
```
JoakimSM, line 95: allowFutureDate was missing after the metadata load rewrite.

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
```diff
@@ -165,32 +166,37 @@ class UIDate extends React.Component<Props, State> {
                     maxWidth,
                 }}
             >
-                { /* // $FlowSuppress */}
+                { /* // $FlowFixMe */}
```
JoakimSM, line 169: After one of the flow js upgrades, flow has started complaining about "missing" props. These props aren't really missing, but rather passed on through our "passOnProps" object. I don't want to annotate props for every level where they are just passed on, but not used (a "interface" level where every prop is listed could make sense). Suppressing these errors for now, we should upgrade flow at some point and try to find a viable solution to this problem.

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
```diff
@@ -1,7 +1,5 @@
 // @flow
```
JoakimSM, line 1: I know there is some code in this file that should have been refactored, but leaving that for another time. I removed the eslint errors and flow errors at least. This component should probably be moved to capture core (because of the DateCalendar), but leaving that to the issue @paschalidi has already started working on.

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
paschalidi: yeah that makes sense this is a long discussion and you right we need to find a way that flow works in our favour here.

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
paschalidi: yeah exactly. I am gonna be opening a PR soon. Kind low on the priority but I have done 80 percent of the work there

.eslintrc
```diff
@@ -52,6 +52,7 @@
             2,
             {
                 "order": [
+                    "static-variables",
```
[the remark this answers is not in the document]
JoakimSM: Yes, removing 👍

JoakimSM:
> From what I understood tho we want the users to be able to select years in the calendar instead of just scrolling back for 5 minutes to get to year 1985 for example. And this PR fixes that.

Correct.

> To be clear, my only suggestion here is that we could avoid forking the library so that we wont be left with the burden to maintain it a copy of an existing library. react-infinite-calendar has its own ecosystem and is a software that is being maintained, which is a good thing for us since it saves us a ton of effort and time.

**The forking was a last resort to get the behavoiur we needed. I wouldn't have forked if it wasn't absolutely necessary.**

The 'react-infinite-calendar' was never made for being shown as a popup as a supplement to a field like we do in the Capture app. The problem with using the non-forked version is: The same callback is provided to the app whether we select year or date in the calendar (and the callback only gets a selected date as argument), so there is no way of knowing (for certain) whether the user selected a date or a year (and this is part of a fundamental problem to the calendar, but won't go into details on this here). In the app we need to know what was selected, because we want to automatically close the popup when a date is selected (but not the year).

PR #812, opened 2020-07-03
fix: TECH-399 adds flow typed

paschalidi:
Hey morning! This is based on the https://github.com/dhis2/capture-app/pull/775 I split the work so that it is a little bit easier to review.

### What is this PR about?
This PR commits all the `flow-typed` from external libraries. As I explained yesterday my inspiration for this is (spectrum)[https://github.com/withspectrum/spectrum]. They also use flow to static type check. 
#### Do we *really* need to add it? 
The only reason is that we want to run the command `flow check` during the CI/CD and this means `flow` would throw erros if there are no types for external libs we using. More over adding the files is not much of a burden for us since they are not making our app slower or larger in size since these files are not bundled. As I said I am open to discuss it. So please bring your thoughts in!


paschalidi:
wrong base. should have based on the `cp/lgtm/04-flow`

PR #816, opened 2020-07-03
fix: DHIS2-9067 extend year range

JoakimSM:


paschalidi:
Lets merge this first so that the build wont fail https://github.com/dhis2/capture-app/pull/819. 
What you think?

JoakimSM:
Yeah, fixed it temporarily. Getting [#819](https://github.com/dhis2/capture-app/pull/819) in first is best, so waiting for that one before merging.

PR #817, opened 2020-07-03
02 feat: DHIS2-8926 renders Input fields depending on how many unique attributes a selection has 

paschalidi:
https://jira.dhis2.org/browse/DHIS2-8926

Hey @JoakimSM. This is based on the https://github.com/dhis2/capture-app/pull/800

Basically adds the functionality the ticket describes. Does not execute any search, simply renders the input fields


src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -28,24 +41,52 @@ const getStyles = (theme: Theme) => ({
...
+    trackedEntityTypesWithCorrelatedPrograms,
+    preselectedProgram,
+    availableSearchOptions,
+    addFormIdToReduxStore,
+    forms,
+}: Props) => {
     const [selectedOption, setSelected] = useState(preselectedProgram);
 
+    // dan abramov suggest to stringify https://twitter.com/dan_abramov/status/1104414469629898754?lang=en
+    const stringifyPrograms = JSON.stringify(availableSearchOptions);
```
JoakimSM, line 73: Ties to a comment from the previous PR regarding memoization of objects from the container (and also one from this PR): If `reselect` is used for `availableSearchOptions` this isn't necessary. As far as I can understand the contents of `availableSearchOptions` will never change and it would be beneficial to keep the same object reference to avoid this 'workaround' (as well as doing the calculation for `availableSearchOptions` once).

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
```diff
@@ -40,15 +63,33 @@ const mapStateToProps = (state: ReduxState) => {
             programs.find(({ programId }) => programId === currentSelections.programId))
         .filter(program => program)[0];
 
+    const availableSearchOptions = Object.values(trackedEntityTypesWithCorrelatedPrograms)
```
JoakimSM, line 66: memoization?

src/core_modules/capture-core/components/Pages/Search/SearchPage.types.js
```diff
@@ -1,22 +1,30 @@
 // @flow
+import RenderFoundation from '../../../metaData/RenderFoundation/RenderFoundation';
```
JoakimSM, line 2:
```suggestion
import { RenderFoundation } from '../../../metaData';
```

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -80,20 +121,69 @@ const Index = ({ classes, trackedEntityTypesWithCorrelatedPrograms, preselectedP
...
+                                        classes.divider,
                                         trackedEntityTypesWithCorrelatedPrograms,
                                     ])
                                 }
                             </SingleSelect>
                         </div>
                     </div>
                 </Section>
+
+                {
```
JoakimSM, line 141: Maybe extract the unique search part into its own component to make it a bit easier on the eyes? (Also applies for the SearchDomainSelector related code)

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: This will be reoccurring in all the current PRs since I havent used `reselect` Just `useMemo`. But is valid point. Can we make a 'refactoring' as a last PRs where we deal with the `reselect` and memoization?

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: yeah valid and I agree. Both are happening in later PRs. You will see it :)

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
paschalidi: https://github.com/dhis2/capture-app/pull/817#discussion_r460847646

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: I am creating a new branch where this is gonna be included. 

However there is a little thing we need to puzzle out. `useEffect` will thrown a warning if we dont add each dependency it is used inside its callback. We are using the `availableSearchOptions` in the callback to add the `formId` in the redux store and therefore `useEffect` wont allow us to skip adding it in the array even if the object is memoized.

Silencing the linter is not recommended by the react team and therefore they suggest this workaround. This is why I ended up with it. I know it is not that intuitive but I am happy to find a better solution with your help

JoakimSM: You will add the `availableSearchOptions` as a dependency to `useEffect`. `availableSearchOptions` will never change if it is memoized, so the effect will never retrigger because of `availableSearchOptions`.

paschalidi: I see. Would give it a try :)

PR #822, opened 2020-07-03
03 feat: DHIS2-8925 direct lookup using unique identifiers

paschalidi:
Hey 👋 this is about the https://jira.dhis2.org/browse/DHIS2-8925

### What is this PR about?
There is functionality added regarding the direct lookup. 
Basically the user can now search using a unique identifier. 
In case the lookup is successful the user is prompted to the Tracker Capture app. 
In case there is no results found there is a popup coming up.

There is validation in the form where I used the same technique that has already been implemented for the `TeiSearch` functionality.

### What is this PR not about? 
The texts needs to be worked out with translations and all. I will be adding a new ticket that will take care of this holistically in this feature.


src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.epics.js
```diff
@@ -0,0 +1,86 @@
...
+import { getTrackedEntityInstances } from '../../../../trackedEntityInstances/trackedEntityInstanceRequests';
+import {
+    getTrackedEntityTypeThrowIfNotFound,
+    getTrackerProgramThrowIfNotFound,
+} from '../../../../metaData';
+import { actionCreator } from '../../../../actions/actions.utils';
+
+const trackerCaptureAppUrl = () => (process.env.REACT_APP_TRACKER_CAPTURE_APP_PATH || '..').replace(/\/$/, '');
+
+export const onScopeProgramFindUsingUniqueIdentifierEpic = (action$: InputObservable, store: ReduxStore) =>
```
paschalidi, line 15: I have created here two epics. One for the lookup where the scope is `TrackedEntityType` and one for when the scope is simply a `Program`.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -58,27 +67,80 @@ const getStyles = (theme: Theme) => ({
     },
 });
 
+const SearchSelection =
```
paschalidi, line 70: This component is being abstracted to make the code more readable. We can discuss if we would like to move it in a separate file if you feel it doesn't belong here.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -91,88 +153,42 @@ const Index = ({
...
-                                    className={classes.searchDomainSelectorSection}
-                                    header={
-                                        <SectionHeaderSimple
-                                            containerStyle={{ paddingLeft: 8, borderBottom: '1px solid #ECEFF1' }}
-                                            title={i18n.t('Search {{name}}', { name })}
-                                        />
-                                    }
+                    searchStatus === searchPageStatus.NO_RESULTS &&
+                    <Modal position="middle">
+                        <ModalTitle>Empty results</ModalTitle>
```
paschalidi, line 171: We will have to make a PR where we correct the text in the Search page as a whole. When I am developing this feature most of the people who are responsible for the text were away.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -91,88 +153,42 @@ const Index = ({
...
+                                    Search Again
+                                </Button>
+                            </ButtonStrip>
+                        </ModalActions>
+                    </Modal>
                 }
+
+                {
+                    searchStatus === searchPageStatus.ERROR &&
+                    <h1 style={{ color: 'red' }}>{searchResultsErrorMessage}</h1>
```
paschalidi, line 189: The idea of this state is that when there is a network error or unknown error of some sort we can let the user know. In the next PR I am styling this further.

src/core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import { createReducerDescription } from '../../trackerRedux/trackerReducer';
+import { searchPageActionTypes } from '../../components/Pages/Search/SearchPage.container';
+
+export const searchPageStatus = {
```
paschalidi, line 5: I am using the term "Status" here to signify the state of the page. 

In this case a status is like a machine state.
image

That means that the Search page would render specific content given the statuses.

src/core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription.js
```diff
@@ -0,0 +1,30 @@
...
+        searchStatus: searchPageStatus.LOADING,
+    }),
+    [searchPageActionTypes.SEARCH_RESULTS_EMPTY]: state => ({
+        ...state,
+        searchStatus: searchPageStatus.NO_RESULTS,
+    }),
+    [searchPageActionTypes.SEARCH_RESULTS_ERROR]: state => ({
+        ...state,
+        searchStatus: searchPageStatus.ERROR,
+        errorMessage: 'There is an error please be aware.',
```
paschalidi, line 24: The text here needs to be worked out as well.

src/core_modules/capture-core/components/D2Form/actions/form.actions.js
```diff
@@ -5,6 +5,6 @@ export const actionTypes = {
     ADD_FORM_DATA: 'AddFormData',
 };
 
-export function addFormData(formId: string, formValues: Object) {
+export function addFormData(formId: string, formValues: $Shape<{||}> = {}) {
```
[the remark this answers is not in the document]
paschalidi: That is true. I was playing around with this and forgot to return it back to its original shape. Will do now.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -0,0 +1,39 @@
...
+const mapDispatchToProps = (dispatch: ReduxDispatch): DispatchersFromRedux => {
+    const {
+        VIA_UNIQUE_ID_ON_SCOPE_TRACKED_ENTITY_TYPE_SEARCH,
+        VIA_UNIQUE_ID_ON_SCOPE_PROGRAM_SEARCH,
+    } = searchPageActionTypes;
+
+    return {
+        searchViaUniqueIdOnScopeTrackedEntityType: ({ trackedEntityTypeId, formId }) => {
+            dispatch(
+                actionCreator(VIA_UNIQUE_ID_ON_SCOPE_TRACKED_ENTITY_TYPE_SEARCH)({ trackedEntityTypeId, formId }),
```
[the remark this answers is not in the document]
paschalidi: Yeah I dont have a strong opinion on this either. It felt when I was writing this that I could have it in one file since they are not being reused. Also sometimes it feels that I end up having many tabs open and all belong to the same component. Like the container file the component file the types and the actions. So I thought maybe I could try it out here and see how you also feel about this. 

However later on I am refactoring this. This will be on the ninth PR of this chain. Is this a blocker for you?

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const mapDispatchToProps = (dispatch: ReduxDispatch): DispatchersFromRedux => {
+    const {
+        VIA_UNIQUE_ID_ON_SCOPE_TRACKED_ENTITY_TYPE_SEARCH,
+        VIA_UNIQUE_ID_ON_SCOPE_PROGRAM_SEARCH,
+    } = searchPageActionTypes;
+
+    return {
+        searchViaUniqueIdOnScopeTrackedEntityType: ({ trackedEntityTypeId, formId }) => {
```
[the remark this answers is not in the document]
paschalidi: Valid point will apply the changes in this PR.

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: @JoakimSM would you mind if I take a note and do this is a later PR to avoid conflicts?

src/core_modules/capture-core/components/D2Form/actions/form.actions.js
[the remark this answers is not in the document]
paschalidi: Flow is a weird thing. This should throw an error since as you say I am explicitly saying here that the only shape of the form values that can be given is `{}`. However no erros being thrown when this function is used with somthing else than `{}`. 

Can you explain why maybe? Just curious here.

src/core_modules/capture-core/components/Pages/Search/SearchPage.container.js
```diff
@@ -67,29 +84,37 @@ const mapStateToProps = (state: ReduxState): PropsFromRedux => {
...
 };
 
 const mapDispatchToProps = (dispatch: ReduxDispatch): DispatchersFromRedux => ({
     addFormIdToReduxStore: (formId) => { dispatch(addFormData(formId)); },
+    closeModal: () => { dispatch(actionCreator(searchPageActionTypes.MODAL_CLOSE)()); },
 });
 
-export const SearchPage = connect<Props, OwnProps, _, _, _, _>(mapStateToProps, mapDispatchToProps)(withLoadingIndicator()(withErrorMessageHandler()(SearchPageComponent)));
+export const SearchPage =
+  compose(
```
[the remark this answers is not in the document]
paschalidi: yeah this looks cute. thanks :)

src/core_modules/capture-core/components/D2Form/actions/form.actions.js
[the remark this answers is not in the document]
JoakimSM: If I add an object literal with a property as an argument to the caller (and keep the $Shape<{||}), Flow will complain. But yeah, absolutelty agree that Flow is hard to get sometimes.

PR #828, opened 2020-07-07
06 chore: DHIS2-9195 configures CI/CD for cypress and writes e2e test

paschalidi:
👋 @JoakimSM after this PR `cypress` will be on! 

This is the ticket https://jira.dhis2.org/browse/DHIS2-9195

### What is this ticket about?
The configuration of the e2e tests is being finalised. That means two things. 
1. there is a github action now that is set up to run the e2e tests in parallel. 
2. there is a `cypress.json` which configures some basic cypress behaviour.

Apart from the configuration the most common use cases of the search page and the `LockedSelector` is being covered.

### How can someone run the tests?
The tests either run in the CI/CD or locally. The tests run against a backend server. 

In the CI/CD this server is the https://debug.dhis2.org/ca-test-2.34 
When running the tests locally cypress will execute the `yarn start` and will start the application locally. Running it locally means that the `env.development` variables will be used to run the tests. The tests would most likely pass will your local database as well, in case they dont please change the `env.development` to the https://debug.dhis2.org/ca-test-2.34 while running the tests :) 

Let me know your thoughts and suggestions on this!!


.github/workflows/cypress-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Upload artifacts
+        if: failure()
+        uses: actions/upload-artifact@v1
+        with:
+          name: failed-screenshots
+          path: cypress/screenshots
+    env:
+      CI: true
+      REACT_APP_DHIS2_BASE_URL: 'https://debug.dhis2.org/2.34dev/'
```
paschalidi, line 27: here we pass the environment variables for the test. 


Phil explained how to set up an instance with some seeded data. The plan here would be to set up an instance just for us that we would call somehow `https://capture-app-test.dhis2.org/2.34dev/`

.github/workflows/cypress-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Upload artifacts
+        if: failure()
+        uses: actions/upload-artifact@v1
+        with:
+          name: failed-screenshots
+          path: cypress/screenshots
+    env:
+      CI: true
+      CYPRESS_RECORD_KEY: '6b0bce0d-a4e8-417b-bbee-9157cbe9a999'
```
paschalidi, line 35: cypress provides us with a dashboard view. there you can see failing tests etc etc. this key is given by them and is not a "secret" therefore I am adding it here.

.github/workflows/cypress-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      REACT_APP_DHIS2_BASE_URL: 'https://debug.dhis2.org/ca-test-2.34'
```
paschalidi, line 36: this instance I have created. Is an instance that simulates the latest of the version we are having with the seeded data our databases usually have at this time.

in https://debug.dhis2.org/ you can see the instance. I have left a message there so that people dont mess with the data.

One problem with this approach is that if the data from the database are changed it might cause our tests to fail. In these cases which should be rare we can spin up a new instance with all the data reseted.

cypress/integration/LockedSelector.feature
```diff
@@ -0,0 +1,119 @@
+Feature: In the main page, use the LockedSelector to navigate
```
paschalidi, line 1: Feature holds a general description of the feature that is tested. Is only for us developers to describe what we test in this feature. Basically, `cypress` encourages us to categorise the tests by features.

cypress/integration/LockedSelector.feature
```diff
@@ -0,0 +1,119 @@
...
+
+  # Main page
+
+  Scenario: Notifying that you need to select org unit and program to get started
```
paschalidi, line 5: In each feature we are having multiple scenarios. It has nothing to do with the tests itself. It can be changed any time and nothing will break.

cypress/integration/LockedSelector.feature
```diff
@@ -0,0 +1,119 @@
...
+    Given you are in the main page with no selections made
```
paschalidi, line 6: Till now it might be clear that we are using cypress for the browser manipulation.
However these `.feature` files are part of the cucumber library. Is very human readable and is what the frontend team suggested for us to use.

Now cucumber comes with some keywords. Apart from the `Feature` and `Scenario` we also having the `Given`, `When`, `Then` and the `And` keywords. 

These four are used to describe actions that will take place on the browser itself. There is always a mapping between the text "you are in the main page with no selections made" and a function in the `js` file. You can find this here https://github.com/dhis2/capture-app/pull/828/files#diff-8fe75cae3f49f1e59a01b231043215b7R1

cypress/plugins/ignore-x-frame-headers/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
```
paschalidi, line 1: This is a plugin this is suggested as a workaround for the issue https://github.com/cypress-io/cypress/issues/4220

We would like to have it removed when cypress fixes this issue. 

The issue basically is that if you navigate to a different origin while doing a cypress test cypress wont allow this. 
An example is for us a navigation from `localhost` to the tracker capture app when a find is successful.

guilryder: I'm the author of the "Ignore X-Frame headers" extension copy-pasted in this commit.

I'm happy that the extension is useful to this project, but FYI the extension is not in the public domain and the [source code](https://github.com/guilryder/chrome-extensions/tree/master/xframe_ignore) has a [license](https://github.com/guilryder/chrome-extensions/blob/master/license.txt).

The extension was GPLv3-licensed when the commit was submitted, which **does not seem directly compatible** with the [BSD license of this project](https://github.com/dhis2/capture-app/blob/master/LICENSE).

Since then I've switched the extension to the MIT license upon request by a reader of https://github.com/cypress-io/cypress/issues/4220.

I think the new MIT license is compatible with BSD, but this **doesn't remove the requirement for proper attribution**.

paschalidi: Hey @guilryder thanks for writing this extensions. Its been very helpful :) What would be the steps to follow in order to achieve proper attribution while using your extension?

guilryder: There's no strict rule, but copying the license file along with the code is enough.

Some projects also reference all the licensed libraries they use in their about/credit page.

[Reference 1](https://opensource.stackexchange.com/questions/5828/where-would-i-give-credit-to-an-mit-licensed-project-in-an-ios-app) - [Reference 2](https://opensource.stackexchange.com/questions/6838/how-to-properly-re-license-code-released-under-the-mit-license)

PR #839, opened 2020-07-13
07 feat: DHIS2-9196 adds collapsable buttons

paschalidi:
Hey 👋 the ticket is https://jira.dhis2.org/browse/DHIS2-9196

Not much going on here simply the collapsibility is being added in the forms.


src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.component.js
```diff
@@ -45,14 +45,45 @@ const Index = ({
...
+    useEffect(() => {
+        setExpandedFormId(null);
+    },
+    [selectedSearchScopeId],
+    );
+
+    const sortedSearchGroup =
+      selectedSearchScopeId ?
+          availableSearchOptions[selectedSearchScopeId].searchGroups
+              .sort(({ unique: xBoolean }, { unique: yBoolean }) => {
```
JoakimSM, line 68: I would prefer the sorting being done in our preconstructed metadata objects instead. If not we would need to duplicate this code to the relationship search.

paschalidi: yeah that is very valid. In the next PR you will see it. I am doing it as part of the refactoring when I basically should have done it here. Apologies the chain was big and sometimes keep track of everything at the same time not so easy

paschalidi: Forget what I said I can move it here if you feel this is easier for you to review. Just let me know :)

JoakimSM: It's fine to switch between PR's, dont worry about it. Will continue the conversion in the next PR.

PR #843, opened 2020-07-27
refactor: moves components with dependencies to the `capture-core` components

paschalidi:


paschalidi:
basing from branch `cp/search/09-show-search-results-table` however it looks as if the differences are ones from the branch `master`. Will close to investigate further and open it again.

PR #853, opened 2020-07-30
13 chore: annotates components correctly

paschalidi:
👋 hello again 

### What is this PR about?

I like flow and I think is useful but I guess you would agree with me that sometimes is a pain. In this case I realised that I wont throw errors for the components. In other words, lets say we wanna use the `SearchForm` component. I would expect flow to throw suggestions and errors about the props the component takes. For example since I am annotating strictly using the `{| ... |}` I would expect it to throw an error when I am passing an extra component. In another scenario, I would expect it to throw an error when a prop that is annotated as required and is not given when you use the component. 

Luckily flow is our friend and this PR fixes basically this small problem so that we can benefit from flow as much as we could. Even when we wrapping with HOCs. 

Let me know if you would like anything to be done in a different way ;)


src/core_modules/capture-core/components/Pages/Search/SearchDomainSelector/index.js
```diff
@@ -1 +1,2 @@
+// @flow
```
paschalidi, line 1: without this here basically flow wont work :P took me a while to realise this bit

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -84,5 +87,8 @@ const mapDispatchToProps = (dispatch: ReduxDispatch): DispatchersFromRedux => ({
 });
 
 
-export const SearchForm =
-  connect<Props, OwnProps, _, _, _, _>(mapStateToProps, mapDispatchToProps)(SearchFormComponent);
+export const SearchForm: ComponentType<OwnProps> =
+  compose(
+      connect<Props, OwnProps & CssClasses, _, _, _, _>(mapStateToProps, mapDispatchToProps),
+      withStyles(getStyles),
```
paschalidi, line 93: Happy to discuss this if you find it strange. To me it looks fine but I can see why one might not like it

src/core_modules/capture-core/components/Pages/Search/SearchDomainSelector/SearchDomainSelector.js
```diff
@@ -85,4 +86,4 @@ export const Index =
       </Section>
       );
 
-export const SearchDomainSelector = withStyles(styles)(Index);
+export const SearchDomainSelector: ComponentType<OwnProps> = withStyles(styles)(Index);
```
paschalidi, line 89: Without this bit here flow will lose the props that we the component has been annotated with before wrapped wit the HOC. And that goes for any HOC.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
JoakimSM: Hehe, you're right, I don't like it. Some of the reasons:

1. The component file should export when thing only, the component.
2. I feel adding styles to the component shouldn't be a concern for the container file, at least not when getStyles is defined in the component file in the first place. It feels right to keep this internal to the component file.
3. We're changing the code to accommodate for flow.

JoakimSM (commented):
Good stuff adding the ComponentType for the HOCs 👍 

Please see my comment regarding one of the changes.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
paschalidi: I see your points. Fair 👍

paschalidi:
@JoakimSM check this out when you free :)

JoakimSM (approved):
LGTM 👍

PR #854, opened 2020-07-30
12 fix: DHIS2-9140 adds correct texts and translations

paschalidi:
Hey @JoakimSM this is another small ticket about text and i18n https://jira.dhis2.org/browse/DHIS2-9140

Have a look and let me know :)


src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -150,7 +152,7 @@ const Index = ({
                             data-test="dhis2-capture-general-purpose-error-mesage"
                             className={classes.generalPurposeErrorMessage}
                         >
-                            {generalPurposeErrorMessage}
+                            {i18n.t(generalPurposeErrorMessage)}
```
JoakimSM, line 155: This won't work, you need to use i18n.t where the actual text is defined.

JoakimSM (commented):
One comment for you @paschalidi then this should be ok 👍

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
paschalidi: Dont see the reason to have this stored in redux. I removed it from there.

paschalidi:
@JoakimSM perhaps you can go on another round on this one

JoakimSM (approved):
LGTM 👍

PR #872, opened 2020-08-06
15 DHIS2-9192 cleans up form on search page unount

paschalidi:
Hey 👋 👋 https://jira.dhis2.org/browse/DHIS2-9192

### What is this PR about?
Maybe I will start with a note this time :) I wanted to say that this ticket originally was about removing the form data from redux on unmount. Later on we decided not to go with it after discussing some implementation limitations. 

However I had already implemented that functionality - the removal of the data on the unmount - and removed it later. While I was doing this I touched different files and since I try to follow the Boy Scout Rule aka "Leave the code cleaner than you found it" I annotated some of the files or changed small bits to follow our conventions. Hope this is fine with you! 

Other than that what happens here is pretty simple. We clean the store from the form data on Search page unmount.


src/core_modules/capture-core/components/D2Form/D2Form.component.js
```diff
@@ -1,37 +1,23 @@
...
-    formHorizontal: boolean,
-};
-
-class D2Form extends React.PureComponent<Props> {
+class D2Form extends React.PureComponent<PropsForPureComponent> {
     name: string;
     sectionInstances: Map<string, Object>;
 
-    constructor(props: Props) {
+    constructor(props: PropsForPureComponent) {
```
paschalidi, line 20: Originally this ticket was about removing the form data from redux on unmount. Later on we decided not to go with it after discussing some implementation limitations. However I touched this file and I took a moment to annotate it. Hope this is fine for this PR.

src/core_modules/capture-core/components/D2Form/D2Form.component.js
```diff
@@ -118,7 +104,7 @@ class D2Form extends React.PureComponent<Props> {
     )
     renderVertical = (section: Section, passOnProps: any, classes: Object) => (
         <div
-            className={this.props.formFoundation.customForm ? classes.containerCustomForm : classes.container}
+            className={classes.containerCustomForm}
```
paschalidi, line 107: there is not much on the container class and I removed it

src/core_modules/capture-core/components/D2Form/D2Form.component.js
```diff
@@ -135,21 +121,41 @@ class D2Form extends React.PureComponent<Props> {
...
 
         return (
             <React.Fragment>
-                {sections}
+                {
+                    isFormInReduxStore
+                        ?
+                        sections
+                        :
+                        log.error(
```
paschalidi, line 143: the component will throw an error when the user wont add the form id into the redux store. this way we let the user know of what they need to do to use this component in the future

src/core_modules/capture-core/components/D2Form/field/Components/OrgUnitField/orgUnitFieldForForms.epics.js
```diff
@@ -19,7 +19,7 @@ import { set as setStoreRoots } from '../../../../FormFields/New/Fields/OrgUnitF
 const FILTER_RETRIEVE_ERROR = 'Filter form field org units failed';
 
 const isAddFormData = (action: Object, formId: string) =>
-    action.type === formActionTypes.ADD_FORM_DATA && action.payload.formId === formId;
+    action.type === formActionTypes.FORM_DATA_ADD && action.payload.formId === formId;
```
paschalidi, line 22: touched the file so I thought to rename this to fit our convention

src/core_modules/capture-core/components/D2Form/index.js
```diff
@@ -1,4 +1,3 @@
 // @flow
-
 export { asyncHandlerActionTypes, asyncUpdateFieldEpic } from './asyncHandlerHOC';
-
+export { D2Form } from './D2Form.container';
```
paschalidi, line 3: thought of exporting the file from the index as we decided we would like to be doing

src/core_modules/capture-core/components/D2Form/D2Form.types.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import type { RenderFoundation } from '../../metaData';
```
paschalidi, line 3: remember we had at some point this tool to complain about some imports that they were unused. This is fixing it 🎉 
I prepared a PR to fix _all_ of them 👍 
https://github.com/dhis2/capture-app/pull/873

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.component.js
```diff
@@ -38,21 +38,53 @@ export const getStyles = (theme: Theme) => ({
...
+            });
+        // we remove the data on unmount to clean the store
+        return () => searchGroupForSelectedScope
+            .forEach(({ formId, searchForm }) => {
+                removeFormDataFromReduxStore(formId);
+
+                Array.from(searchForm.sections.entries())
+                    .map(entry => entry[1])
+                    .forEach(({ id }) => {
+                        removeFormDataFromReduxStore(`${formId}-${id}`);
```
JoakimSM, line 61: I would suggest that we call `removeFormDataFromReduxStore` only once for each form (and deal with the sections in the reducer)

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -86,6 +84,7 @@ const mapDispatchToProps = (dispatch: ReduxDispatch): DispatchersFromRedux => ({
         dispatch(actionCreator(searchPageActionTypes.CURRENT_SEARCH_INFO_SAVE)({ searchScopeType, searchScopeId, formId, currentSearchTerms }));
     },
     addFormIdToReduxStore: (formId) => { dispatch(addFormData(formId)); },
+    removeFormDataFromReduxStore: (formId) => { dispatch(addFormData(formId)); },
```
JoakimSM, line 87: Don't you think we can completely remove the relevant object properties from the redux store? (using the removeFormData action you've created)

JoakimSM (commented):
Thanks for the PR @paschalidi 👍 

I've made some comments. Please have a look.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.component.js
paschalidi: Now is being called once. The logic is handled in the container. Perhaps we can avoid logic into the reducer part of the application when it can be handled into the containers.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
paschalidi: I think there was a bug. I believe now the key is deleted completely :) 

Check that the keys from the object are removed when the form 'unmounts'
image
image

paschalidi:
This is ready for another round @JoakimSM

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.component.js
JoakimSM: When I'm dispatching the action `removeFormData` I expect all form data for the form to be wiped, that includes all sections of that form. If we want to remove section data only we could create a separate action for that (not needed at the moment). Doesn't that make sense?

paschalidi: Yeah I might understand why this is confusing. And I agree it is confusing..

What this iteration does is not removing each one of the fields individually, but basically it constructs some ids (see image). Then it triggers the `removeFormData` to remove these formIds from the store.So in other worlds line 61 will remove the object `searchForm-${id}-#MAIN#` from the store.

I agree this is confusing, but at that time I thought it is necessary. I will find a better way of doing that now. Thanks for raising this

image

JoakimSM (approved):

PR #877, opened 2020-08-13
18 feat: DHIS2-9258 styles and adds general error message 

paschalidi:
👋 this is the https://jira.dhis2.org/browse/DHIS2-9258
After our design doc review the text has changed. 

### What is this PR about? 
Changes the error text and styles it according to the design


src/core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription.js
```diff
@@ -32,7 +32,7 @@ export const searchPageDesc = createReducerDescription({
     [searchPageActionTypes.SEARCH_RESULTS_ERROR_VIEW]: state => ({
         ...state,
         searchStatus: searchPageStatus.ERROR,
-        generalPurposeErrorMessage: 'There is a problem with this search, please change the search terms or try again later. For more details open the Console tab of the Developer tools',
+        generalPurposeErrorMessage: 'Your search returned too many results. Try changing the search terms or search with more attributes to narrow down the results.',
```
JoakimSM, line 35: So, we now have a general error message that states that there is a problem with the number of results returned? What if there is another problem? If we don't check the actual error message that is returned from the api, we can't be this specific.

paschalidi: Sorry I copied the wrong text here. My bad :|

JoakimSM (approved):
👍

PR #895, opened 2020-08-24
feat: DHIS2-9294 remove F_VIEW_USER authority prerequisite for user assignment

JoakimSM:
- using new userLookup endpoint in search field
- using new assigned user display name from events endpoint
- username data element is storing and showing the username only.


Bekkalizer (approved):
Looks good!

PR #904, opened 2020-08-31
fix: DHIS2-9347 Align assigned user field in row view

paschalidi:
Hey @JoakimSM a short one.

This will fix the bug from this https://jira.dhis2.org/browse/DHIS2-9347 ticket.

### DIFF
--- a/src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/Assignee/Assignee.component.js
+++ b/src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/Assignee/Assignee.component.js
@@ -11,8 +11,9 @@ const getStyles = () => ({
         alignItems: 'center',
         padding: 8,
     },
-    containerHorizontal: {
+    containerVertical: {
         display: 'flex',
+        flexDirection: 'column',
         margin: 8,
     },
     label: {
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/components/Pages/NewEvent/DataEntry/Assignee/Assignee.component.js
```diff
@@ -21,31 +27,25 @@ const getStyles = () => ({
...
-
+    const { classes, orientation, ...passOnProps } = props;
     return (
-        <div
-            className={classes.container}
+        <div className={classes[classNames({
+            container: orientation === 'horizontal',
+            containerVertical: orientation === 'vertical',
+        })]}
         >
```
JoakimSM, line 40: This felt unnecessary complex and a bit weird since we can't concat different classes this way. I would just use the ternary operator (if you feel the urge to use classNames, use it in the regular way). I'm adding this comment only, but this applies to all usage of classNames in this file.

paschalidi: Yeah was complex for me as well for sure. Will follow what you suggested. :)

paschalidi:
@JoakimSM did as you suggested. Thanks. Somehow I got myself into this complex `className` situation.

JoakimSM (approved):
👍

PR #909, opened 2020-09-01
chore: removes unused directory cache

paschalidi:
I had some time and while I was looking into the codebase found this folder that is not being used anymore. 

Please double check I am not mistaken @JoakimSM :)

This has no real ticket its more of maintenance work. I can ofc create one :)


JoakimSM (approved):
LGTM 👍

PR #929, opened 2020-09-08
fix: DHIS2-9347 Align assigned user field in row view

paschalidi:
v33 back port for the https://jira.dhis2.org/browse/DHIS2-9347


JoakimSM (approved):
LGTM 👍

PR #943, opened 2020-09-15
19 DHIS2-9245 View active enrollment 

paschalidi:
@JoakimSM this works the ticket https://jira.dhis2.org/browse/DHIS2-9245.

You can test this by creating users in the tracker capture app and then completing their enrolment. 
If you run this against the https://debug.dhis2.org/ca-test-2.35 I would search for `Arran` in the Child Program. There you can see that this Arran has previously active enrolments.


src/core_modules/capture-core/components/Pages/Search/SearchResults/SearchResults.component.js
```diff
@@ -112,3 +150,5 @@ export const SearchResultsComponent = ({
         </div>
     </>);
 };
+
+export const SearchResultsComponent: ComponentType<Props> = withStyles(getStyles)(SearchResultsIndex);
```
paschalidi, line 154: Moved this here as we discussed in the past

JoakimSM (commented):
Thanks for the PR @paschalidi 

Couple of points here. I can see that the we've moved away from the `CardList` and `CardListItem` being generic components to only support our current tei search use case. This is fair enough from a 'don't do premature abstractions' standpoint, but could we rename the files and remove unnecessary component props to reflect this? (Could be a different PR of course).

I can see that the search is now failing in the relationship part of the app (both in the actual search and the duplicate search in the registration form). Could you have a look at this?

Also, if you navigate back to the main page after doing a search, when you reenter the search page, the results are still displayed.

paschalidi:
Thanks for looking into this and bringing your comments in Joakim. 

> This is fair enough from a 'don't do premature abstractions' standpoint, but could we rename the files and remove unnecessary component props to reflect this?

> I can see that the search is now failing in the relationship part of the app (both in the actual search and the duplicate search in the registration form). Could you have a look at this?

Strange. Will def have a look

paschalidi:
Please lets move this converstation in the new PR. Apologies for this unexpected closed PR.

PR #944, opened 2020-09-16
20 fix: DHIS2-9246 removes total pages from cature app

paschalidi:
As we discussed this piece of work could have been done alone. However was easier for me to simply do them together. 
Tickets: https://jira.dhis2.org/browse/DHIS2-9246 https://jira.dhis2.org/browse/DHIS2-8705 

Your thoughts on this one would be appreciated Joakim


src/core_modules/capture-core/components/Pagination/withDefaultNavigation.js
```diff
@@ -73,19 +67,11 @@ const getNavigation = (InnerComponent: React.ComponentType<any>) =>
                     <IconButton
                         data-test={'dhis2-capture-search-pagination-next-page'}
                         onClick={this.handleNextButtonClick}
-                        disabled={currentPage >= Math.ceil(rowsCount / rowsPerPage)}
```
paschalidi, line 76: Well this is a bit problematic. Perhaps you have some ideas how to achieve the same.

src/core_modules/capture-core/trackedEntityInstances/trackedEntityInstanceRequests.js
```diff
@@ -42,10 +42,10 @@ async function convertToClientTei(apiTei: ApiTrackedEntityInstance, attributes:
     };
 }
 
-export async function getTrackedEntityInstances(queryParams: ?Object, attributes: Array<DataElments>) {
+export async function getTrackedEntityInstances(queryParams: Object, attributes: Array<DataElments>) {
     const api = getApi();
     const apiRes = await api
-        .get('trackedEntityInstances', { ...queryParams, totalPages: true });
+        .get('trackedEntityInstances', queryParams);
```
paschalidi, line 48: Looked to me this is always an object.

JoakimSM (commented):
Thanks for the PR @paschalidi 

Replied to your question regarding the disabling of the button.

I could not get the event list pager to work. I also think we should display the page instead of records now in the pager component because in the event list the pager is updated before the data is retrieved (and it also makes sense when we need to show empty results). Let's discuss on slack.

src/core_modules/capture-core/components/CardList/CardList.component.js
```diff
@@ -9,7 +9,7 @@ import type { CardDataElementsInformation, SearchResultItem } from '../Pages/Sea
 type OwnProps = $ReadOnly<{|
     dataElements: CardDataElementsInformation,
     items: Array<SearchResultItem>,
-    currentSearchScopeProgramName?: string,
+    currentSearchScopeName?: string,
```
paschalidi, line 12: Renamed this since sometimes is TEType and not program

src/core_modules/capture-core/components/Pages/MainPage/EventsList/Pagination/ListPagination.container.js
```diff
@@ -0,0 +1,24 @@
+// @flow
```
paschalidi, line 1: _This comment applies for the whole file_

Replaced this src/core_modules/capture-core/components/Pages/MainPage/EventsList/Pagination/withData.js
with a container. Looked to me that this change moves this component closer to our style.

src/core_modules/capture-core/components/Pages/MainPage/EventsList/Pagination/ListPagination.component.js
```diff
@@ -0,0 +1,33 @@
+// @flow
```
paschalidi, line 1: _This comment applies for the whole file_

Simply renamed the `/capture-core/components/Pages/MainPage/EventsList/Pagination/ListPagination.js` and exporting it now as const instead of default. No real functional changes.

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/GeneralOutput/WarningsSection/SearchGroupDuplicate/searchGroupDuplicate.epics.js
```diff
@@ -71,17 +71,32 @@ export const loadSearchGroupDuplicatesForReviewEpic = (action$: InputObservable,
...
+                }) => {
+                    const isPaginationButtonDisabled = Boolean(nextPageSearchResults.length === 0);
+                    return duplicatesForReviewRetrievalSuccess(
+                        searchResults,
+                        {
+                            currentPage: pagingData.currentPage,
+                            isPaginationButtonDisabled,
+                        },
+                    );
+                }),
```
paschalidi, line 99: Using `forkJoin` we can perform multiple requests at a time. We get the Observable back when they are completed. 

I think this is handling gracefully the limitation we face with the backend at the moment. We dont ask the user to click the next paging button to show them an empty page but we do it for them under the hood.

I applied the changes with the forkJoin in all the places that search takes place.

src/core_modules/capture-core/components/Pagination/withDefaultNavigation.js
paschalidi: I am passing the `isPaginationButtonDisabled` for the Pagination. This way the one who is using the component can feed it with this property accordingly or not.

paschalidi:
Hey @JoakimSM. This is now ready for another round. 

Couple of changes I did are: 

1. added a prop to the Pagination component which is responsible for disabling the next button.
2. fixed the issue you found in the EventList which the next page wasnt working as expected.
3. added some tests in the search page to cover the case where the pagination is disabled.

Please as always dont hesitate to share your thoughts.

src/core_modules/capture-core/components/Pagination/withDefaultNavigation.js
```diff
@@ -19,9 +19,8 @@ const styles = (theme: Theme) => ({
 });
 
 type Props = {
-    rowsPerPage: number,
+    isPaginationButtonDisabled: boolean,
```
JoakimSM, line 22: NIT: could we go with a more precise name here?

paschalidi: you mean such as `isPaginationNextButtonDisabled`?

src/core_modules/capture-core/components/TeiSearch/epics/teiSearch.epics.js
```diff
@@ -79,15 +79,27 @@ const searchTei = (state: ReduxState, searchId: string, formId: string, searchGr
...
-            formId,
-            searchGroupId,
-            searchId,
-        ),
-    )
-        .catch(() => searchTeiFailed(formId, searchGroupId, searchId));
+
+    return forkJoin({
+        first: getTrackedEntityInstances(queryArgs, attributes),
+        second: getTrackedEntityInstances({ ...queryArgs, filter: filters, page: queryArgs.page + 1 }, attributes),
```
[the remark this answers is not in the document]
paschalidi: > (when rows received equals rowsPerPage)

Perhaps explain to me what you mean here a little bit further.

src/core_modules/capture-core/components/Pagination/withDefaultNavigation.js
JoakimSM: Yes. Suggestion: `nextPageButtonDisabled`

paschalidi: You mind if wit is `isNextPageButtonDisabled `

JoakimSM: `isNextPageButtonDisabled` sounds more like a function to me...

paschalidi: Fair 👌  There is a convention by some in which, `is` at the beginning of a variable name indicates boolean.

I will go with your suggestion to make it clear for both of us

src/core_modules/capture-core/components/TeiSearch/epics/teiSearch.epics.js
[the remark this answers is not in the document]
JoakimSM: Yes. Let's do one request only (for the current page). Based on the results we know:
- If received rows is less than rows per page, there is no next page.
- If received rows equals rows per page, we can't be sure if there will be any rows on the next page. Let's keep the button active, and accept that the next page might contain no rows for now.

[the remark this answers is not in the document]
paschalidi: Yeah I understand, that can work. 

It is perhaps a little bit complicated to get this straight. I will try to explain why and perhaps we can figure out a good way to go forward on this together.

At the moment, we are reusing the same epics for 
a) when a user clicks the search button
b) when the user clicks the next or previous or any pagination button.

For example imagine a the user that is on the search page. They search for TEI _but_ there is absolutely no result returned. What happens then is that we display this modal giving them a clue of their situation.

This is a solvable problem obviously but the solutions might be a little bit strange. 

Do we want now to differentiate between the fetches that we do from the search button and the search we do from the next pagination button?

Or what is your take here?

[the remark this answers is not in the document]
JoakimSM: Looking at the code and provided you would like to keep the searchStatus, I think the best bet is to handle it in the epic (searchViaAttributesStream function) based on which page you're on.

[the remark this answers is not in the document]
paschalidi: Will give it a shot tomorrow. Thanks

paschalidi:
@JoakimSM look at my last two commits https://github.com/dhis2/capture-app/pull/944/commits/0ed20c57c1cdd0759594fe470e50f8c1fef9b24d and https://github.com/dhis2/capture-app/pull/944/commits/55126a627b685ba17fe9ba6b2a60f1c119dad172 handled your suggestions there. Now the user will se an empty page when there are no results returned _and_ they have triggered the search from the pagination. This is from my point of view ready for another review. Let me know your thoughts on this please 👍

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.epics.js
```diff
@@ -42,12 +43,24 @@ const getFiltersForAttributesSearchQuery = formValues =>
...
+                return showSuccessResultsViewOnSearchPage(
+                    searchResults,
+                    { currentPage: pagingData.currentPage, nextPageButtonDisabled },
+                );
             }
+
+            if (searchResults.length === 0 && triggeredFrom === PAGINATION) {
+                return showSuccessResultsViewOnSearchPage(
+                    [],
+                    { currentPage: pagingData.currentPage, nextPageButtonDisabled: true },
```
JoakimSM, line 60: One thing I've learned is that it is better to store only what is needed, i.e. save a common state that can be derived from and do the computations when rendering. The `nextPageButtonDisabled` could be derived from state you already have. 

In the same vein, you could derive from the currentPage if you would like to show an empty page or the messages.

Furthermore, the above leads to why I don't like the search status. Is it really needed or could the search status be derived from other state? Is it just going to be a pain to keep the search status up to date and prevent it from going stale?

This is just friendly advice for the future ;) I would appreciate it if we could compute the nextPageButtonDisabled instead of storing it though (lets keep the search status and the other solutions for now)

paschalidi: Hey I agree on that you dont wanna be storing _all_ the information and things could derive. 

The state approach I am taking in my eyes ( which are of course biased in this case )  helps the reader understand what to expect from the `SearchPage.component.js` when they start reading it. 

I am removing the `nextPageButtonDisabled` from the state. Can we please not be absolute on the `seachStatus` and give it a go to see how this works in the longer run?

paschalidi: @JoakimSM also seeing the code again a little bit. And I want to stress out that we mostly have the same opinion on the topic here. I also strongly agree that things that shouldnt be stored on redux shouldnt be there. However I am just gonna raise another point because I rushed into answering the message before and hadnt though through it well... 

The next button derives from two values at the moment. For example
```
const nextPageButtonDisabled = Boolean(searchResults.length < pagingData.rowsPerPage);
```
the values that used to derive the state of the button are, 
a) the length of the results on the response and 
b) the `rowsPerPage` from the response as well.

Now I have stopped storing the `rowsPerPage` in the reducers. (exception to this is the working reducer list coz there is needed).

So we can store the rowsPerPage and then derive the status of the button but then from this point of view isnt the `rowsPerPage` that isnt really needed since its only used to derive the button state?

paschalidi: Or do you mean there is another way to derive the button state that I havent considered?

JoakimSM: Since we don't allow for changing rows per page, we should just set this in the component itself and pass it as a argument when initiating the search.

paschalidi: Which component?

JoakimSM: > Can we please not be absolute on the seachStatus and give it a go to see how this works in the longer run?

Yes, we're not changing this now. Regarding the readability, you could compute the status in the component.

paschalidi: > Regarding the readability, you could compute the status in the component.

Yeah I am happy to understand your point of view on this one and perhaps find a way that both of us like to go forward

JoakimSM: > Which component?
In the lowest common ancestor to where rows per page is needed :)

paschalidi: I still dont think I get that fully. 

At the moment the `pageSize` is hardcoded to number 5. And this is happening usually in the epics. 
The places that this happens at the moment is: 
1. `searchGroupDuplicate.epics`
2. `searchForm.epics`
3. `teiSearch.epics`

What I understand _( and perhaps I am wrong )_ is that now, we would like the number of the pages to be defined on the component level. We could then pass the number alongside with other information in the action creator and then into the epics. This way we can pass _also_ the number 5 into the `Pagination` component that is defined on those 3 places defined above?

JoakimSM (approved):
Approving, remaining [comment](https://github.com/dhis2/capture-app/pull/944#discussion_r500851782) handled in different PR.

PR #945, opened 2020-09-16
21 fix: bug on search with range values

paschalidi:
While implementing some features I found out about a bug. 

### What is this bug about?
The bug basically is that when you fetch the `trackedEntityInstances` with some range values you need to add specific filters that include `ge:${value}:le${value}` in the request. 
Also you need to display correctly the form value when this is not a string but an object `{from:"string", to:"string}`. 

Fixed this also in the TEI search realm since displaying the object part of the bug was present there.

Lastly wrote some tests to make sure this wont break again in some near future. The tests cover the SearchPage and the NewEvent page search with range values


JoakimSM (approved):
LGTM

PR #951, opened 2020-09-21
fix: element types flow related errors

paschalidi:
There is no ticket for this yet. I can gladly create one if you want @JoakimSM.
### Motivation
When I added the automated `$FlowFixMe` comments I realised there were quite many comments generated related to the `dataElementTypes`. This PR fixes the majority of these comments. We do that by defining `dataElementTypes` as object instead of building it with a `reduce`. This way we can better annotate it and use this annotation around the project.

### What this PR does?
1. imports always `elementTypes` as `dataElementTypes`. 
2. removes all unused `$FlowFixMe` comments
3. changes `$Values<typeof dataElementTypes>` to `$Keys<typeof dataElementTypes>`. There is a slight but very important difference between those two. I think `$Keys` is what we want to use here. 

Please let me know your thoughts on all of those.


src/index.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import 'capture-core-utils/extensions/asyncForEachArray';
 import 'capture-core-utils/extensions/arrayToHashMap';
-// $FlowFixMe autogenerated content
+// $FlowFixMe
```
paschalidi, line 5: dunno why I had to add this again?

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/eventFiltersInterface/toEventQueryCriteriaConverter/optionSet.js
```diff
@@ -11,9 +9,7 @@ import type {
     OptionSetFilterData,
 } from '../../../../EventsList/eventList.types';
 
-const stringifyNumber = (rawValue: number) => {
-    rawValue.toString();
-};
+const stringifyNumber = (rawValue: number) => rawValue.toString();
```
paschalidi, line 12: Looks like this was not returning a value before. Perhaps that was intended but I doubt it. Can you also check Joakim that I am not breaking anything here?

src/core_modules/capture-core/metaData/DataElement/dataElementTypes.js
```diff
@@ -0,0 +1,37 @@
+// @flow
```
paschalidi, line 1: renamed this file to dataElementTypes since this is the exported value

JoakimSM (approved):
Thanks @paschalidi!

After a flow update this problem occured. I must say I liked it better defined as an array and then processed, but I agree with we should accommodate for flow in this particular instance.

PR #970, opened 2020-09-22
19 feat: DHIS2-9245 View active enrollment 

paschalidi:
The PR https://github.com/dhis2/capture-app/pull/943 because I have merged all the previous one in the chain and it could base itself to master.

Repeating the description here:
@JoakimSM this works the ticket https://jira.dhis2.org/browse/DHIS2-9245.

You can test this by creating users in the tracker capture app and then completing their enrolment.
If you run this against the https://debug.dhis2.org/ca-test-2.35 I would search for Arran in the Child Program. There you can see that this Arran has previously active enrolments.


paschalidi:
I also move your message here Joakim after you reviewed https://github.com/dhis2/capture-app/pull/943

The message after your review is: 
Thanks for the PR @paschalidi

Couple of points here. I can see that the we've moved away from the CardList and CardListItem being generic components to only support our current tei search use case. This is fair enough from a 'don't do premature abstractions' standpoint, but could we rename the files and remove unnecessary component props to reflect this? (Could be a different PR of course).

I can see that the search is now failing in the relationship part of the app (both in the actual search and the duplicate search in the registration form). Could you have a look at this?

Also, if you navigate back to the main page after doing a search, when you reenter the search page, the results are still displayed.

paschalidi:
@JoakimSM I addressed your comments. 
Specifically: 
>This is fair enough from a 'don't do premature abstractions' standpoint, but could we rename the files and remove unnecessary component props to reflect this? (Could be a different PR of course).

I am addressing this in a later PR.

> I can see that the search is now failing in the relationship part of the app (both in the actual search and the duplicate search in the registration form). Could you have a look at this?

I am addressing this in a later PR as we have discussed in the past.

> Also, if you navigate back to the main page after doing a search, when you reenter the search page, the results are still displayed.

This is fixed.

Could you please have another look on this one?

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -76,6 +76,8 @@ const Index = ({
         if (!preselectedProgram.value) {
             showInitialSearchPage();
         }
+
+        return () => showInitialSearchPage();
```
paschalidi, line 80: This will solve the issue you mentioned. The one that when you come back to the back from the main page you still see the results displayed.

JoakimSM (approved):

PR #977, opened 2020-09-24
fix: alerts reported from lgtm

paschalidi:
@JoakimSM another one with zero functional changes. I believe this will be the last time you hear from lgtm. 
At least for now hehe


JoakimSM (approved):
Good work @paschalidi

PR #978, opened 2020-09-24
fix: test that failed unexpectedly during CI/CD

paschalidi:
@JoakimSM there is this test that sometimes failed and sometimes not. 

This is because it can happen that the request clean up sometimes doesnt take place. The user is then added in the database. Next time the test runs we have a duplicate on the person. In this case the tests passes only if the machine is quick enough to press the create user button before the server responds with possible duplicates. 

To solve this I am registering a user we already know exists in the database. And by changing the order the machine is typing. So now the machine first types the name and last name which is used to detect duplicates and then the dates. This way hopefully there will be always time the detection of the duplicates

I hope this will fix the issue.


JoakimSM (approved):
LGTM 👍 Still wonder why it fails, but this is a good workaround.

PR #982, opened 2020-09-25
chore: upgrades from `@dhis2/ui-core` and `@dhis2/ui-widgets` to `@dhis2/ui`

paschalidi:
Hey Joakim 👋 

### What is this PR about?
No functional changes again. Simply updating to v5 of the ui library. 
There were some breaking changes in the `SingleSelector` so I took care of them. If you like you could see all the breaking changes [here](https://github.com/dhis2/ui/releases?after=v5.0.6).

### Why is this Part 1?

This is part of a three chain PR with this one being the first. The second is https://github.com/dhis2/capture-app/pull/990 and the third is https://github.com/dhis2/capture-app/pull/992.

Updating to v5 is something we can only benefit from. So I took the chance and put this together. However updating to v5 had breaking changes. This is the reason this is part of a chain. The second PR basically is about the search domain selector and its new design by Joe. Would be double the update to the new design and then merge this PR into master. 

The third PR, which is about the new button design is also part of this chain because the same reason. Meaning I needed to have the new domain selector into place before working on the new find button concept


src/core_modules/capture-core/components/Pages/Search/SearchDomainSelector/SearchDomainSelector.component.js
```diff
@@ -50,8 +50,8 @@ export const Index =
               <div className={classes.searchRowTitle}>{ i18n.t('Find results from') }</div>
               <div className={classes.searchRowSelectElement} style={{ marginRight: 8 }}>
                   <SingleSelect
-                      onChange={({ selected }) => { onSelect(selected); }}
-                      selected={selectedSearchScope}
+                      onChange={(selected) => { onSelect(selected); }}
```
paschalidi, line 53: There were some breaking changes that lead me to this change

JoakimSM (approved):
LGTM

PR #984, opened 2020-09-29
fix: API breaking changes to v35 

paschalidi:
Hey this is according the breaking changes happened in the v35 https://github.com/dhis2/notes-backend/blob/master/platform/35/changelog/user_property_transformer.md


JoakimSM:
Closing this because we need to deal with this problem differently. see: [PR](https://github.com/dhis2/capture-app/pull/985)

PR #985, opened 2020-09-29
fix: use userLookup in filter and bugfix

JoakimSM:
- use userLookup endpoint in assignee filter
- view / working list name can now be filled in


paschalidi (approved):
Thanks for looking into this 👍

PR #992, opened 2020-10-02
feat: DHIS2-9675 Redesign Find button to match the new concept

paschalidi:
https://jira.dhis2.org/browse/DHIS2-9675

### What is this PR doing?

Adds the dropdown menu on the Find button of the LockedSelector.

image

The button basically has two states. 
a) when there is _no_ Tracker program selected you can only see a button. 
b) when there is  a Tracker program selected then there is a dropdown button. The dropdown has two options as we discussed in the concept. 

You can find the concept [here](https://dhis2.github.io/design-specs/capture-context/capture-context-test/?01_blank).

### DIFF
--- a/src/core_modules/capture-core/components/LockedSelector/QuickSelector/ActionButtons.component.js
+++ b/src/core_modules/capture-core/components/LockedSelector/QuickSelector/ActionButtons.component.js
@@ -1,9 +1,9 @@
 // @flow
-import React, { type ComponentType } from 'react';
+import React, { type ComponentType, useMemo } from 'react';
 import { withStyles } from '@material-ui/core/styles';
 import i18n from '@dhis2/d2-i18n';
-import { Button, colors } from '@dhis2/ui';
-import { TrackerProgram } from '../../../metaData';
+import { Button, colors, DropdownButton, FlyoutMenu, MenuItem } from '@dhis2/ui';
+import { getProgramFromProgramIdThrowIfNotFound, TrackerProgram } from '../../../metaData';
 
 const styles = ({ typography }) => ({
     marginLeft: {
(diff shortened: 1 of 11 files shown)


paschalidi:
@cooper-joe perhaps you wanna have a look here for the design as well :)

cooper-joe:
The `Find` button usage look good 👍 

The menu that opens from the button is right-aligned, is it possible to make it left-aligned instead?
image

paschalidi:
will look into it Joe thanks

src/core_modules/capture-core/components/LockedSelector/QuickSelector/QuickSelector.component.js
```diff
@@ -119,8 +120,9 @@ class QuickSelector extends Component<Props> {
                         <ActionButtons
                             selectedProgramId={this.props.selectedProgramId}
                             onStartAgainClick={this.props.onStartAgain}
-                            onFindClick={this.props.onClickFind}
-                            onNewClick={this.props.onClickNew}
+                            onFindClick={this.props.onFindClick}
```
paschalidi, line 123: Changed this bit coz I neglected it on some previous PR. So was inconsistent on this component now.

src/core_modules/capture-core/components/LockedSelector/QuickSelector/ActionButtons.component.js
```diff
@@ -26,13 +26,41 @@ type Props = $ReadOnly<{|
...
+    TRACKER_PROGRAM: 'TRACKER_PROGRAM',
+    EVENT_PROGRAM: 'EVENT_PROGRAM',
+};
+
+function useProgramInfo(programId) {
+    const [programName, setName] = useState('');
+    const [programType, setType] = useState('');
+    const [trackedEntityName, setTrackedEntityName] = useState('');
+
+    useEffect(() => {
```
JoakimSM, line 43: Why do we use `useEffect` here and why are we saving state here? Wouldn't `useMemo` make more sense?

JoakimSM (commented):
One comment, other than that looks good.

src/core_modules/capture-core/components/LockedSelector/QuickSelector/ActionButtons.component.js
paschalidi: Yeah good point. Do you know any differences in those two concepts?

JoakimSM: `useEffects` is for side effect and runs after every render. More here: [useEffect](https://reactjs.org/docs/hooks-effect.html)

JoakimSM (approved):

PR #995, opened 2020-10-06
fix: DHIS2-9709 rules engine assigned values of type boolean and true only

JoakimSM:
fix for [JIRA-issue](https://jira.dhis2.org/browse/DHIS2-9709)

### DIFF
--- a/src/core_modules/capture-core/rules/engine/converters/rulesEffectsValueConverter.js
+++ b/src/core_modules/capture-core/rules/engine/converters/rulesEffectsValueConverter.js
@@ -22,23 +22,17 @@ class RulesValueConverter implements IConvertOutputRulesEffectsValue {
     convertEmail(value: string): string {
         return value;
     }
-    convertBoolean(value: boolean): boolean {
-        return value;
+    convertBoolean(value: boolean): string {
+        return value ? 'true' : 'false';
     }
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/rules/engine/converters/rulesEffectsValueConverter.js
```diff
@@ -70,9 +64,6 @@ class RulesValueConverter implements IConvertOutputRulesEffectsValue {
         return value;
     }
     convertAge(value: string): ?Object {
```
paschalidi, line 66:
```suggestion
    convertAge(value: string): Object {
```

paschalidi: Looks to me that this will always return an object now. Perhaps this is unnecessary since `Object === any`

paschalidi (approved):
Tested locally looks good to me. 

I have only one comment that is not a blocker for me but perhaps you wanna have a look into it.
The tests you wrote helped me understand the problem, thanks. Perhaps would be useful to write tests for the case of the date as well? Its up to you though if you would like to do so or not.

Irrelevant but I have run the tests locally and they have failed. Do you maybe have any clue why?

PR #996, opened 2020-10-06
fix: DHIS2-9709 rules engine assigned values of type boolean and true only

JoakimSM:
fix for [JIRA-issue](https://jira.dhis2.org/browse/DHIS2-9709)


paschalidi (approved):
Looks good.

PR #1008, opened 2020-10-09
refactor: shares the pageSize using context

paschalidi:
Hey there 👋 

I am opening a series of 3 PR regarding the context change we discussed [here](https://github.com/dhis2/capture-app/pull/944#pullrequestreview-504655030). 

### What is this PR doing?

I only write e2e tests. For two reasons. One is that is good to have them. Two is that will help us later make sure that nothing broke on the way.

### Why is this a chain of PRs?

As you probably gathered already this is a chain of three PR. Why is this a chain? While I was trying to get to use context for passing down this prop instead of using redux. I got myself into starting making some component hooks in order to try out the `useContext` hook. Then I got carried a bit away and annotated some classes that I touched. Keep in mind I annotate only the classes that will then be touched from the context refactoring.

So now we have three steps to reach our final goal use `Context`. 
1. is this PR we are seeing now.
2. https://github.com/dhis2/capture-app/pull/1009 is dealing with annotations, exporting components as `const` and converting classes to functional components
3. https://github.com/dhis2/capture-app/pull/1010 does the actual context refactoring

To answer the question "Why is this a chain". I believe splitting them the way I described above will make the work easier for you. Focusing on one topic at a time while reviewing you can give me better feedback and as a result have a more concrete outcome.


JoakimSM (approved):
LGTM 👍

PR #1010, opened 2020-10-09
refactor: [pageSize via context part 3] shares the pageSize using context

paschalidi:
👋  @JoakimSM 

### What is this PR about?
As being said in the overview here https://github.com/dhis2/capture-app/pull/1008#issue-500441377 this PR deals with sharing the pageSize using context. 

_I use context only when it is needed._ That means only when the prop has to be passed down to several components. To avoid that I use context. You will see that in one case I am passing it down as prop in some component.

Please let me know your thoughts. 👍 

To be honest with you, and this is only my opinion, I dont have the feeling that the solution I wrote here simplifies the code. Perhaps there is much room for improvement in the solution I have provided. But if I am being honest as I see it _now_, refactoring into using context seems to complicate things than the opposite. I say this because as you will see the arguments of many functions are being increased by one and passing the `pageSize` around to derive the button state and also to use it to fetch the next 5 items looks a bit cumbersome. 

I know we are saying redux should not store these kind of values but in my eyes we were doing this before already. We used stored the number `5` in the store and used it to show the text `x of 5` in the Pagination. 

Anyway sorry for talking so much, you can look for yourself. Please let me know your thoughts


JoakimSM (approved):
Looks good to me 👍 

> To be honest with you, and this is only my opinion, I dont have the feeling that the solution I wrote here simplifies the code. Perhaps there is much room for improvement in the solution I have provided. But if I am being honest as I see it now, refactoring into using context seems to complicate things than the opposite. I say this because as you will see the arguments of many functions are being increased by one and passing the pageSize around to derive the button state and also to use it to fetch the next 5 items looks a bit cumbersome.

Still think this is the best approach here, but I can see the point your making about storing the page size.

My original point still stands though. We should be careful of storing to specific data, it's better to more generic and do the computations when rendering. Or else we might end up in a situation where we need to remember to do a lot of computions when storing and/or we will be saving unnecessary data. Hence, it would be better to save the page size instead of nextPageButtonDisabled, because if something else needed to know the page size you would find yourself in a sitation where you would need to save more data because nextPageButtonDisabled doesn't help. I think the page status is a better example because I think you probably could compute this when rendering instead of saving it (Didn't really dig into it though so might be something I'm missing).

Anyway, good job on this PR 👍

paschalidi:
> I think the page status is a better example because I think you probably could compute this when rendering instead of saving it (Didn't really dig into it though so might be something I'm missing).

I have been thinking about it. Lets discuss it in person at some point :)

PR #1015, opened 2020-10-12
feat: have a loo into `redux-toolkit

paschalidi:
Hey @JoakimSM for an overview of the toolkit have a look [here](https://redux-toolkit.js.org/). As I said it is basically a package written from the current redux maintainer. An article that I read quickly and I liked is [this one](https://blog.logrocket.com/smarter-redux-with-redux-toolkit/). Nothing special just an overview. :) 

To sum it up the advantages I see using this are: 
1.  Better type definitions of the redux state. 
2. Mutate state directly using `immer`
3. Reduce boilerplate of the action creators. We could even use one file for the actions and the reducers sine now action creators are only one line long each 🤯 

Again I dont think this is crusial change. I am only creating this PR as a suggestion and to have your point of view on this. I think it can bring benefits and improve the overal health of the codebase if used but it is not that we _need_ to go with it.


src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -92,12 +91,12 @@ const mapDispatchToProps = (dispatch: ReduxDispatch, { searchGroupsForSelectedSc
         const currentSearchTerms =
           collectCurrentSearchTerms(searchGroupsForSelectedScope, formsValues);
 
-        dispatch(actionCreator(searchPageActionTypes.CURRENT_SEARCH_INFO_SAVE)(
-            { searchScopeType,
-                searchScopeId,
-                formId,
-                currentSearchTerms,
-            }));
+        dispatch(saveCurrentSearchInfo({
```
paschalidi, line 94: This I forgot to change on one my previous PR. Apologies

paschalidi, line 94: This I forgot to change on one my previous PR. Apologies

src/core_modules/capture-core/components/Pages/Search/SearchPage.actions.js
```diff
@@ -1,48 +1,21 @@
...
+export const searchViaAttributesOnScopeProgram = createAction('search/searchViaAttributesOnScopeProgram');
 
-export const searchViaUniqueIdOnScopeProgram = ({ programId, formId }) =>
-    actionCreator(searchPageActionTypes.VIA_UNIQUE_ID_ON_SCOPE_PROGRAM_SEARCH)({ programId, formId });
+export const showInitialViewOnSearchPage = createAction('search/showInitialViewOnSearchPage');
+export const showLoadingViewOnSearchPage = createAction('search/showLoadingViewOnSearchPage');
+export const showSuccessResultsViewOnSearchPage = createAction('search/showSuccessResultsViewOnSearchPage');
+export const showEmptyResultsViewOnSearchPage = createAction('search/showEmptyResultsViewOnSearchPage');
+export const showErrorViewOnSearchPage = createAction('search/showErrorViewOnSearchPage');
+export const saveCurrentSearchInfo = createAction('search/saveCurrentSearchInfo');
```
paschalidi, line 18: So basically now we dont have to write strings as types. To refer the type of an action you simply refer the searchViaUniqueIdOnScopeProgram which is the result of the createAction redux-toolkit provides.

This has some advantages to what plain the redux solution provides. As for instance boilderplate.

src/core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription.js
```diff
@@ -1,6 +1,13 @@
...
-import { searchPageActionTypes } from '../../components/Pages/Search/SearchPage.actions';
+import { createReducer } from '@reduxjs/toolkit';
+import {
+    saveCurrentSearchInfo,
+    showEmptyResultsViewOnSearchPage,
+    showErrorViewOnSearchPage,
+    showInitialViewOnSearchPage,
+    showLoadingViewOnSearchPage,
+    showSuccessResultsViewOnSearchPage,
+} from '../../components/Pages/Search/SearchPage.actions';
```
paschalidi, line 10: Again here we are only passing the action creator functions. We will use them to add cases to our reducers.

src/core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription.js
```diff
@@ -10,39 +17,54 @@ export const searchPageStatus = {
...
+                state.searchStatus = searchPageStatus.ERROR;
+            })
+        .addCase(
+            saveCurrentSearchInfo,
+            (
+                state,
+                { payload: { searchScopeType, searchScopeId, formId, currentSearchTerms } },
+            ) => {
+                state.currentSearchInfo = { searchScopeType, searchScopeId, formId, currentSearchTerms };
+            });
```
paschalidi, line 68: So there are couple of improvements here. 

First one is annotation. Now we can annotate the state on the component and have it annotated when we use it later on either in epics or containers.

Second is `Immer` `redux-toolkit` comes with `Immer` as enchantment. That means we will always have immutable state without having to do all the destruction with es6. Meaning the `...state` and so on. 

We can now simply assign variables and `Immer` will do the resty for us under the hood. 🎉

src/store/getStore.js
```diff
@@ -43,24 +40,24 @@ export function getStore(
...
-    );
 
+    const store = toolkitConfigureStore({
+        reducer: enableBatching(offlineEnhanceReducer(rootReducer)),
+        middleware,
+        devTools: true,
+        enhancers: defaultEnhancers => [offlineEnhanceStore, ...defaultEnhancers],
+    });
+
+    // this line needs to be executed after the configuration of the store.
```
paschalidi, line 60: here is merely configuration. Nothing has changed.

paschalidi (commented):
Hey I made some comments regarding the toolkit in order to give you my thoughts on the topics.

PR #1048, opened 2020-10-26
DHIS2-9793 Text is not aligned correctly

paschalidi:
👋 Joakim this solves the https://jira.dhis2.org/browse/DHIS2-9793

If you have a look in the ticker you will see that the bug is described pretty well there from Geetha.

Basically this ticket fixes this for the different inputs that had the same behaviour.

### DIFF
--- a/src/core_modules/capture-core/components/D2Form/field/buildField.module.css
+++ b/src/core_modules/capture-core/components/D2Form/field/buildField.module.css
@@ -11,11 +11,11 @@
 }
 
 .ageLabel {
-    padding-top: 27px;
+    padding-top: 13px;
 }
 
 .coordinateLabel {
-    padding-top: 27px;
+    padding-top: 13px;
 }
 
 .polygonLabel {
@@ -23,11 +23,11 @@
 }
 
 .dateLabel {
-    padding-top: 32px;
+    padding-top: 13px;
 }
 
 .dateTimeLabel {
-    padding-top: 27px;
+    padding-top: 13px;
 }
 
 .imageLabel {
--- a/src/core_modules/capture-core/components/D2Form/field/buildField.module.css
+++ b/src/core_modules/capture-core/components/D2Form/field/buildField.module.css
@@ -23,7 +23,7 @@
 }
 
 .dateLabel {
-    padding-top: 13px;
+    padding-top: 32px;
 }
 
 .dateTimeLabel {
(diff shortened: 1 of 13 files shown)


src/core_modules/capture-ui/AgeField/AgeField.component.js
```diff
@@ -211,11 +210,8 @@ class D2AgeField extends Component<Props> {
             onParseDate,
             ...passOnProps
         } = this.props;
-        const dateInputContainerClass = classNames(
-            { [defaultClasses.ageDateInputContainerHorizontal]: !isVertical },
-        );
         return (
-            <div className={dateInputContainerClass}>
```
paschalidi, line 218: I removed these classes because as far as I could see, both `ageDateInputContainerHorizontal` and `dateInputContainerClass` seemed either empty classes or werent present in the css file.

src/core_modules/capture-ui/DateAndTimeFields/DateRangeField/DateRangeField.component.js
```diff
@@ -100,29 +100,24 @@ class DateRangeField extends React.Component<Props> {
         const toValue = value && value.to ? value.to : '';
         return (
             <div className={defaultClasses.container}>
-                <div className={defaultClasses.inputContainer}>
```
paschalidi, line 103: There is no class `dateInputContainerClass` in the css file

JoakimSM:
@paschalidi I can see that row view is broken here and that's the reason why for instance `isVertical` is never used. Did some digging and its because of this PR: https://github.com/dhis2/capture-app/commit/c3b5b5a7917fe4a847a2d5d524790c2b57b3e544# specifically because formHorizontal is extracted from the props here: https://github.com/dhis2/capture-app/blob/c3b5b5a7917fe4a847a2d5d524790c2b57b3e544/src/core_modules/capture-core/components/D2Form/D2Form.component.js#L130. Fix that and you will see that `isVertical` is actually used.

src/core_modules/capture-core/components/D2Form/field/buildField.module.css
```diff
@@ -1,5 +1,5 @@
 .booleanLabel {
-    padding-top: 6px;
+    padding-top: 13px;
```
paschalidi, line 2: image
`booleanLabel` is responsible for what we see in the image. I also tried to align them

paschalidi:
@JoakimSM took a look on what you mentioned. I reverted the changes. Can you give this one more look?

paschalidi:
image
image

JoakimSM (approved):
perfect, thanks

PR #1053, opened 2020-10-26
[DHIS2-8695] Refactor working lists #5 - clean interfaces

JoakimSM:
[DHIS2-9754 (Subtask of DHIS2-8695)](https://jira.dhis2.org/browse/DHIS2-9754)

Main purpose is to make a good interface for the "ListView" and "WorkingLists". I'm mostly renaming props and some complexity is left for later PRs. The interface specifications are found in the `workinglists.types` and `listView.types` files.


src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
```diff
@@ -461,16 +460,27 @@ export const workingListsContextDesc = createReducerDescription({
     The meaning is slightly changed though, having a context now implies that a request for events was done for this context,
     not that events was successfully retrieved for this context.
     */
```
paschalidi, line 462: Is this comment still above the right reducer case?

paschalidi (approved):
Really great work on splitting this up! I really mean it, great work 🙇

paschalidi:
@JoakimSM forgot to say on the review but can you have a look why the types are failing and perhaps why lgtm gives a new alert?

src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
JoakimSM: Good catch. Rephrased a bit, because it applies to both reducer cases here.

JoakimSM:
@paschalidi Types should be fixed in next PR. LGTM complaints because of some flow complexity (correct code) that is removed in a later PR (tried to use generics, but figured out it wasn't worth the hazzle at this point)

src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
paschalidi: Sorry I think I didnt explain myself well. 

I think that now the comment is above an action called `TEMPLATES_FETCH` while as far as I understand the comments refers to the action with name `LIST_VIEW_INIT`. 

If this comment applies to the whole reducer perhaps then the best place to have it is above the line 
```
export const workingListsContextDesc = createReducerDescription({
```

src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
```diff
@@ -457,20 +456,31 @@ export const workingListsColumnsOrderDesc = createReducerDescription({
 
 export const workingListsContextDesc = createReducerDescription({
     /*
-    Setting context on init (not on success anymore) because it makes sense for the loading effect.
-    The meaning is slightly changed though, having a context now implies that a request for events was done for this context,
-    not that events was successfully retrieved for this context.
+    Setting context on fetch/init (not on success anymore) because it makes sense for the loading effect.
```
paschalidi, line 459:
```suggestion
    Setting context on templates fetch and list view init (not on success anymore) because it makes sense for the loading effect.
```

paschalidi: Since it applies for both reducers but not for the third we can be a little bit more explicit here.

PR #1055, opened 2020-10-26
[DHIS2-8695] TEI working lists #7 - move common actions to WorkingListsCommon

JoakimSM:
[DHIS2-9802 (Subtask of DHIS2-8695)](https://jira.dhis2.org/browse/DHIS2-9802)

Moving **most** actions to "workingListsCommon". The idea is to use same Redux actions for both the single event and tracked entity instance lists to get "free" state management. For most actions this probably makes sense, but there is no problem deviating now if different behaviour is required.


paschalidi (approved):
Feels great to review those small PRs. Thanks Joakim for the work

Issue #13, opened 2018-03-19
Implement "Selector" for program and orgunit


Bekkalizer:

Bekkalizer:
Moved to Jira

PR #1061, opened 2020-10-26
[DHIS2-8695] Refactor working lists #13 - use array as working lists datasource

JoakimSM:
[DHIS2-9757 (Subtask of DHIS2-8695)](https://jira.dhis2.org/browse/DHIS2-9757)

Use array as working lists datasource.


src/core_modules/capture-core/components/Pages/MainPage/EventWorkingLists/DataSourceSetup/EventWorkingListsDataSourceSetup.component.js
```diff
@@ -55,10 +56,17 @@ export const EventWorkingListsDataSourceSetup = ({
         columns,
     ]);
 
+    const dataSourceArray = useMemo(() =>
+        recordsOrder && recordsOrder
+            .map(id => dataSource[id]), [
+        dataSource,
+        recordsOrder,
+    ]);
```
paschalidi, line 64: Dont know what is your opinion here but I find that there will be more consistent if we returned an empty array in case records order is not defined. 

This way we know that this is _always_ an array but it can sometimes just be empty.

src/core_modules/capture-core/components/Pages/MainPage/EventWorkingLists/DataSourceSetup/eventWorkingListsDataSourceSetup.types.js
```diff
@@ -19,6 +24,6 @@ export type Props = {|
 export type EventWorkingListsDataSourceSetupOutputProps = {|
     ...RestProps,
     columns: EventWorkingListsColumnConfigs,
-    dataSource: DataSource,
+    dataSource?: DataSource,
```
paschalidi, line 27: Relevant to the other comment I am making. `DataSource` will then not be optional. But will be annotated a little bit different 

```
export type DataSource = Array<?DataSourceItem>;
```

I dont know your opinion on this but for me defining the value with _same type_  instead of `null` is more consistent. I guess I dont have any knockout arguments but perhaps is something we can think of.

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/workingLists.types.js
```diff
@@ -44,8 +45,6 @@ export type ColumnConfig = {
 
 export type ColumnConfigs = Array<ColumnConfig>;
 
-export type DataSource = { [id: string]: Object };
```
paschalidi, line 47: Good stuff reusing the one we already have 👍

paschalidi (commented):
Easy! Good work here :

src/core_modules/capture-core/components/Pages/MainPage/EventWorkingLists/DataSourceSetup/EventWorkingListsDataSourceSetup.component.js
JoakimSM: For me, it depends. Most of the time I agree with you, here it doesn't really matter, because the value should be defined after the data is fetched from the api.

src/core_modules/capture-core/components/Pages/MainPage/EventWorkingLists/DataSourceSetup/eventWorkingListsDataSourceSetup.types.js
JoakimSM: To be clear, It's never `null` here, it's either `undefined` or it has value. I don't feel it makes sense to change it in this instance.

src/core_modules/capture-core/components/Pages/MainPage/EventWorkingLists/DataSourceSetup/EventWorkingListsDataSourceSetup.component.js
JoakimSM: Also sometimes you will utilize this distinction, like I do in the offline lists in an upcoming PR.

paschalidi: Yep okay that makes sense to me as well

paschalidi (approved):
Hey thanks for following up on those here

PR #1083, opened 2020-10-28
01 feat: DHIS2-9647 Create empty registration page

paschalidi:
Ticket: https://jira.dhis2.org/browse/DHIS2-9647

Have a look Joakim


src/core_modules/capture-core/components/Pages/New/NewPage.component.js
```diff
@@ -0,0 +1,58 @@
...
+
+const NewPagePlain = ({
+    showMessageToSelectOrgUnitOnNewPage,
+    showDefaultViewOnNewPage,
+    classes,
+    currentOrgUnitId,
+    newPageStatus,
+}: Props) => {
+    useEffect(() => {
+        if (!currentOrgUnitId) {
```
JoakimSM, line 25: I know we discussed these statuses before, but this is good example. Here you could just check if there is currentOrgUnitId directly when the page is rendered instead of dispatching an action. If you want to, you could compute the status in this component instead of saving it in the Redux store.

JoakimSM (commented):
Made a comment regarding the statuses. Apart from that, looks good.

src/core_modules/capture-core/components/Pages/New/NewPage.component.js
paschalidi: Right. Perhaps we could look at an example where this pattern adds some value.

This technique shines and works very well with epics and asynchronous actions.
An example would be this: 

```
const searchViaAttributesStream = (queryArgs, attributes, triggeredFrom) => {
    const stream: Stream = from(getTrackedEntityInstances(queryArgs, attributes));
    return stream.pipe(
        map(({ trackedEntityInstanceContainers: searchResults, pagingData }) => {
            if (searchResults.length > 0) {
                return showSuccessResultsViewOnSearchPage(
                    searchResults,
                    pagingData.currentPage,
                );
            }
            return showEmptyResultsViewOnSearchPage();
        }),
        startWith(showLoadingViewOnSearchPage()),
        catchError(() => of(showErrorViewOnSearchPage())),
    );
};
```

Where I want the page to be in loading state when the user is triggering the request and I want to show an error when errors are being shown. 

I am gonna try to explain the biggest advantage of having this string, instead of `boolean` values or computed values. 

Basically when you have boolean values for `loading` and `error` you can be on a state that both `loading` and `error` can be true. Whereas in actuality there is no point for your page to be in that state. 

State machines is common pattern in computer science in general. Every arrow in the image is a user action and can take them to one state. 
image

Here are some other resources 
[1]https://www.youtube.com/watch?v=VU1NKX6Qkxc


What are some drawbacks you see?

JoakimSM: Thinking about this a bit, let's go with the solution you proposed 👍 I think the defined statuses earlier was what threw me off a bit.

JoakimSM (approved):

PR #1150, opened 2020-11-05
feat: DHIS2-9244 fallback search

paschalidi:
Fallback functionality is on this one. https://jira.dhis2.org/browse/DHIS2-9244
Testing is coming right after https://github.com/dhis2/capture-app/pull/1185

Hope my comments will help out with the review of this. If some things arent clear or you need anything ping me on slack :)


src/core_modules/capture-core/components/CardList/CardListItem.component.js
```diff
@@ -85,15 +85,15 @@ const deriveEnrollmentType =
           .map(({ status, lastUpdated }) => ({ status, lastUpdated }));
 
 
-      const { ACTIVE, CANCELLED, COMPLETED, NOT_ENROLLED } = enrollmentTypes;
+      const { ACTIVE, CANCELLED, COMPLETED } = enrollmentTypes;
```
paschalidi, line 88: Realised that there is no use case  for `NOT_ENROLLED`. 
This is because at the moment we are only fetching TEInstances from either a context of 
a) Program (Tracker & Event) or 
b) TEType. 

In case a) we have `ACTIVE`, `CANCELLED`, `COMPLETED`  and in case b) we have `DONT_SHOW_TAG`

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.container.js
```diff
@@ -88,18 +89,19 @@ const mapDispatchToProps = (dispatch: ReduxDispatch, { searchGroupsForSelectedSc
...
-                formId,
-                currentSearchTerms,
-            }));
+        dispatch(saveCurrentSearchInfo({
+            searchScopeType,
+            searchScopeId,
+            formId,
+            currentSearchTerms,
+        }));
+    },
```
paschalidi, line 101: Looks like I have forgotten to extract this. I am creating an action creator from this now

src/core_modules/capture-core/components/LockedSelector/LockedSelector.container.js
```diff
@@ -37,7 +38,11 @@ const mapDispatchToProps = (
         dispatch(setOrgUnitFromLockedSelector(id, orgUnit));
     },
     onSetProgramId: (id: string) => {
-        dispatch(setProgramIdFromLockedSelector(id));
+        dispatch(
+            batchActions([
+                setProgramIdFromLockedSelector(id),
+                ...customActionsOnProgramIdSet,
+            ], lockedSelectorBatchActionTypes.PROGRAM_ID_SET_BATCH));
```
paschalidi, line 45: Added the same pattern here so that the LockedSelector can basically trigger extra functionality when there is a set of the program id. 

This is useful on the callback to clean up some things.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.component.js
```diff
@@ -44,23 +44,54 @@ const useFormDataLifecycle = (
...
+            .forEach(({ formId, searchForm }) => {
+                const elements = searchForm.getElements();
+                const formValuesThatExistInTETypeSearchScope =
+                  ([...elements.values()].reduce((acc, { id: fieldId }) => {
+                      if (Object.keys(keptFallbackSearchFormValues).includes(fieldId)) {
+                          const fieldValue = keptFallbackSearchFormValues[fieldId];
+                          return { ...acc, [fieldId]: fieldValue };
+                      }
+                      return acc;
+                  }, {}));
```
paschalidi, line 62: This is a little bit complicated. At the moment its is own hook called `useFormDataLifecycle`. 

To explain what we basically doing here I need to explain some bits before. 

So one of those bits is that, when we fallback search this happens on a scope of a program and then broadens up to a scope of a TEType. When this happens basically we push to the `router` a new url.

However we want to keep the search terms that are relevant and have them pre-filed in the form. 

This is what we are doing here. When we instantiate the form we pre-fill it with the relevant data.

src/core_modules/capture-core/utils/url.js
```diff
@@ -10,3 +10,21 @@ export const urlArguments = (programId: string, orgUnitId: string): string => {
...
+    }
+
+    if (programId) {
+        argArray.push(`programId=${programId}`);
+    } else if (trackedEntityTypeId) {
+        argArray.push(`trackedEntityTypeId=${trackedEntityTypeId}`);
+    }
+
+    return argArray.join('&');
+};
```
paschalidi, line 30: I wanted to have your opinion on this approach before I go on and replace the function above everywhere in the app. If you think that the implementation described here can work I can go on and do the replacement on the next PR.

src/core_modules/capture-core/components/Pages/Search/SearchForm/SearchForm.epics.js
```diff
@@ -160,3 +162,94 @@ export const searchViaAttributesOnScopeTrackedEntityTypeEpic: Epic = (action$, s
...
+                const { formsValues, searchPage: { currentSearchInfo: { currentSearchTerms: searchTermsFromOriginalSearch } } } = store.value;
+
+                const { searchForm, formId: fallbackFormId } = deriveSearchFormInfo(availableSearchOptions[trackedEntityTypeId].searchGroups);
+                const fallbackFormValues = deriveFormValues(searchForm, formsValues[formId]);
+                const fallbackSearchTerms = deriveCurrentFallbackSearchTerms(searchTermsFromOriginalSearch, fallbackFormValues);
+
+                return concat(
+                    of(fallbackPushPage({ orgUnitId, trackedEntityTypeId })),
+                    of(fallbackSearch({ trackedEntityTypeId, fallbackFormValues, pageSize })),
+                    of(saveCurrentSearchInfo({
```
paschalidi, line 208: Here we are having three actions being dispatched. 
One that pushes to the new page. 
One that saves the current search information such us search terms. We do this because we want the form in the new page to be prepopulated with the same values the user has already typed.
And last we trigger the fallbackSearch which does the actual fetching of the users.

paschalidi:
Looks like I got a bug while demoing this. So will convert to draft

src/core_modules/capture-core/reducers/descriptions/currentSelections.reducerDescriptionGetter.js
```diff
@@ -213,5 +216,9 @@ export const getCurrentSelectionsReducerDesc = (appUpdaters: Updaters) => create
         categories: undefined,
         categoriesMeta: undefined,
     }),
+    [searchPageActionTypes.FALLBACK_SEARCH_COMPLETED]: (state, { payload: { trackedEntityTypeId } }) => ({
+        ...state,
+        trackedEntityTypeId,
+    }),
```
paschalidi, line 222: Thinking abou this again. It seems like using the url to get this information and not the redux store might be a better idea. What you think?

JoakimSM (approved):
Looks very good 👍 Good work

PR #1213, opened 2020-11-20
DHIS2-9785 Capture app crashes when there is no public access for one of the tracked entity type

paschalidi:
Check this out Joakim. It fixes the bug I believe.

Also I tried to annotate and followed eslint where it was saying that returning awaiting is redundant.

### DIFF
--- a/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
+++ b/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
@@ -155,7 +155,7 @@ function sortPrograms(programs: Array<EventProgram | TrackerProgram>) {
     return programs;
 }
 
-async function getBuiltPrograms(
+function getBuiltPrograms(
     cachedPrograms: Array<CachedProgram>,
     cachedOptionSets: Map<string, CachedOptionSet>,
     cachedRelationshipTypes: Array<CachedRelationshipType>,
@@ -165,10 +165,6 @@ async function getBuiltPrograms(
     trackedEntityTypeCollection: Map<string, TrackedEntityType>,
     locale: ?string,
 ) {
-    if (cachedPrograms.length <= 0) {
-        return null;
-    }
-
     const programFactory = new ProgramFactory(
         cachedOptionSets,
         cachedRelationshipTypes,
--- a/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
+++ b/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
@@ -2,15 +2,11 @@
 import { typeof StorageController } from 'capture-core-utils/storage';
 import type { ProgramRule, ProgramRuleVariable } from '../../rules/engine';
 import { ProgramFactory } from './factory';
-import type {
-    Program,
-    TrackedEntityType,
-} from '../../metaData';
+import type { TrackedEntityType, EventProgram, TrackerProgram } from '../../metaData';
 import { programCollection } from '../../metaDataMemoryStores';
 import getRulesAndVariablesFromProgramIndicators from './getRulesAndVariablesFromIndicators';
 import { getUserStorageController } from '../../storageControllers';
 import { userStores as stores } from '../../storageControllers/stores';
-
 import type {
     CachedProgram,
     CachedOptionSet,
@@ -63,16 +59,14 @@ async function getBuilderPrerequisites() {
     const cachedRelationshipTypesPromise = getRelationshipTypes(storageController, stores.RELATIONSHIP_TYPES);
     const cachedCategoriesPromise = getCategories(storageController, stores.CATEGORIES);
 
-    const values =
-        await Promise.all([
-            cachedProgramsPromise,
-            cachedProgramRulesVariables,
-            cachedProgramRules,
-            cachedProgramIndicatorsPromise,
-            cachedRelationshipTypesPromise,
-            cachedCategoriesPromise,
-        ]);
-    return values;
+    return Promise.all([
+        cachedProgramsPromise,
+        cachedProgramRulesVariables,
+        cachedProgramRules,
+        cachedProgramIndicatorsPromise,
+        cachedRelationshipTypesPromise,
+        cachedCategoriesPromise,
+    ]);
 }
 
 function addProgramVariables(d2ProgramRulesVariables: Array<ProgramRuleVariable>) {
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
```diff
@@ -63,16 +59,14 @@ async function getBuilderPrerequisites() {
...
-        ]);
-    return values;
+    return Promise.all([
+        cachedProgramsPromise,
+        cachedProgramRulesVariables,
+        cachedProgramRules,
+        cachedProgramIndicatorsPromise,
+        cachedRelationshipTypesPromise,
+        cachedCategoriesPromise,
+    ]);
```
paschalidi, line 69: Please correct me if I get this wrong here. I followed what eslint was suggesting. 

That basically was saying that in an async function awaiting for the promise to resolve and then return it is redundant since an async function always returns a promise anyways.

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
```diff
@@ -149,7 +143,7 @@ function addRulesAndVariablesFromProgramIndicators(cachedProgramIndicators: Arra
     });
 }
 
-function sortPrograms(programs: Array<Program>) {
+function sortPrograms(programs: Array<EventProgram | TrackerProgram>) {
```
paschalidi, line 146: Also not sure I understand the difference between `Program` and `EventProgram | TrackerProgram`. I changed it because flow was complaining that some functions dont exist in Program but exist on `EventProgram` or `TrackerProgram`. 

Perhaps you could explain the differences to me if you got a minute.

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
```diff
@@ -172,7 +166,7 @@ async function getBuiltPrograms(
     locale: ?string,
 ) {
     if (cachedPrograms.length <= 0) {
-        return [];
+        return null;
```
paschalidi, line 169: Changed that because flow was assuming that an empty array is the same as `Program`. 
We had this discussion the other day about `null` and `[]`. Seems that we could return `null` here and nothing much will change. 

What you think?

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
```diff
@@ -185,21 +179,24 @@ async function getBuiltPrograms(
...
-        const program = await programFactory.build(cachedProgram);
-        return program;
-    });
-    const programs = await Promise.all(promisePrograms);
-    return programs;
+    const promisePrograms = cachedPrograms
+        .filter(cachedProgram => (
+            cachedProgram.trackedEntityTypeId &&
+            trackedEntityTypeCollection.get(cachedProgram.trackedEntityTypeId)
+        ))
```
paschalidi, line 186: Here is basically what we discussed. Having it in the factory would result the class returning sometimes program as `undefined`. 

This is because 
```
    async build(cachedProgram: CachedProgram) {
        let program;
        if (cachedProgram.programType === 'WITHOUT_REGISTRATION') {
            program = new EventProgram((o) => {
                o.id = cachedProgram.id;
                o.access = cachedProgram.access;
                o.name = cachedProgram.displayName;
                o.shortName = cachedProgram.displayShortName;
                o.organisationUnits = cachedProgram.organisationUnits;
                o.categoryCombination = this._buildCategoryCombination(cachedProgram.categoryCombo);
            });
            const d2Stage = cachedProgram.programStages && cachedProgram.programStages[0];
            program.stage =
                await this.programStageFactory.build(
                    d2Stage,
                    program.id,
                );
        } else {
            program = new TrackerProgram((o) => {
                o.id = cachedProgram.id;
                o.access = cachedProgram.access;
                o.name = cachedProgram.displayName;
                o.shortName = cachedProgram.displayShortName;
                o.organisationUnits = cachedProgram.organisationUnits;
                // $FlowFixMe
                o.trackedEntityType = this.trackedEntityTypeCollection.get(cachedProgram.trackedEntityTypeId);
            });

            if (cachedProgram.programTrackedEntityAttributes) {
                program.searchGroups = await this.searchGroupFactory.build(
                    cachedProgram.programTrackedEntityAttributes,
                    cachedProgram.minAttributesRequiredToSearch,
                );

                // $FlowFixMe
                program.attributes = await this._buildProgramAttributes(cachedProgram.programTrackedEntityAttributes);
            }

            // $FlowFixMe
            await cachedProgram.programStages.asyncForEach(async (cachedProgramStage: CachedProgramStage) => {
                program.addStage(
                    await this.programStageFactory.build(
                        cachedProgramStage,
                        program.id,
                    ),
                );
            });

            program.enrollment = await this.enrollmentFactory.build(cachedProgram, program.searchGroups);
        }
        // $FlowFixMe
        program.icon = await ProgramFactory._buildProgramIcon(cachedProgram.style);

        return program;
    }

```

basically we would have to say return undefined when the program doesnt exist in `this.trackedEntityTypeCollection`. This wouldnt be very nice I believe. This is why I went with this alternative

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
JoakimSM: I would prefer keeping this one as `[]`. I tried changing it back and I didn't see any flow errors? You could also remove the entire check I think (probably the best option)

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
JoakimSM: Yes, I concur with the approach 👍

JoakimSM (approved):
Approving, but please look into my comment.

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
paschalidi: Will try it tomorrow thanks for looking into it

paschalidi: Right no no errors. I am removing the statement as you suggested.

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/programsBuilder.js
paschalidi: Right

PR #1219, opened 2020-11-24
feat: DHIS2-9276 #1 - Init tei working lists

JoakimSM:
First part of [DHIS2-9276](https://jira.dhis2.org/browse/DHIS2-9276)

The purpose here is to get an empty dummy list up an running. No fetching, no columns, no data and no working functionality in the list.


src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
```diff
@@ -152,7 +152,6 @@ export const workingListsTemplatesDesc = createReducerDescription({
             criteria,
             isDefault: undefined,
             notPreserved: true,
-            skipInitDuringAddProcedure: true,
```
JoakimSM, line 155: should have been removed earlier..

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/types/concept.types.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+type TeiRecord = {| [id: string]: any |};
+
+export type TeiRecords = {| [teiId: string]: TeiRecord |};
+
+export type TeiWorkingListsTemplate = {
```
JoakimSM, line 6: Can't do exact annotaion here.

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
```diff
@@ -0,0 +1,65 @@
...
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { TeiWorkingListsSetup } from '../Setup';
+import { useWorkingListsCommonStateManagement } from '../../WorkingListsCommon';
+import { useTrackerProgram } from '../../../../../hooks/useTrackerProgram';
+import { TEI_WORKING_LISTS_TYPE } from '../constants';
+import type { Props } from './teiWorkingListsReduxProvider.types';
+
+export const TeiWorkingListsReduxProvider = ({ storeId }: Props) => {
+    const programId = useSelector(({ currentSelections }) => currentSelections.programId);
```
paschalidi, line 11: Thats more of a thought than a suggestion to change the code. 

I am thinking if it makes any sense to create its own hook from this one since if I find myself writing the same selector over and over. What you think?

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
```diff
@@ -0,0 +1,65 @@
...
+        lastTransactionOnListDataRefresh: undefined,
+        listDataRefreshTimestamp: undefined,
+        initialViewConfig: undefined,
+        records: undefined,
+        recordsOrder: undefined,
+    };
+    // ---------------------------------------------------------------
+
+    return (
+        // $FlowFixMe Yep, dealing with this later
```
paschalidi, line 58: :) thanks for the note

src/core_modules/capture-core/hooks/useProgramInfo/index.js
```diff
@@ -1,3 +1,3 @@
 // @flow
-export { useProgram } from './useProgram';
+export { useProgramInfo } from './useProgramInfo';
 export { programTypes } from './programTypes.const';
```
paschalidi, line 3: I was thinking that at the moment we only have three types in total. 

1. TEType.
2. Event Program
3. Tracker Pogram

I have created [my own constant](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/metaData/helpers/getScopeFromScopeId.js#L6) to describe those. I think it would be better if we can align on this so that we have only one constant describing the types around the app. What you think?

paschalidi (approved):
Hey thanks for the small PR. I made some comments that arent blocking the PR but I also wanted to discuss those point with you. Perhaps we can do it in slack, can you please take a moment and look at them and drop me a message when you got some time?

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
JoakimSM: Hmm. Probably more work to import this one from somewhere else though, and very minor readability improvement. But I won't stop you doing it ;)

src/core_modules/capture-core/hooks/useProgramInfo/index.js
JoakimSM: Good point 👍 Could have been a different PR, but made a suggestion in the last commit in this one. Please have a look @paschalidi.

paschalidi: Yeah looks rad!

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
paschalidi: Yeah dont know if there is value here since this is indeed too small to do it anyway. But perhaps if in the future if I use it 100 times I might do it in the end.

PR #1228, opened 2020-11-26
feat: DHIS2-9808 Capture TEType id in the url

paschalidi:
https://jira.dhis2.org/browse/DHIS2-9808

Here I am adding the `trackedEntityTypeId` in the url and also in the redux store. Each time a user changes their selection from the `TrackedEntityTypeSelector` this will be reflected in the url.

## How to test

Try to use the `TrackedEntityTypeSelector` from either the Search or New page and see if the url is changing.


src/core_modules/capture-core/reducers/descriptions/app.reducerDescriptionGetter.js
```diff
@@ -124,10 +124,6 @@ export const getAppReducerDesc = (appUpdaters: Updaters) => createReducerDescrip
         return newState;
     },
 
-    [lockedSelectorActionTypes.SELECTIONS_FROM_URL_UPDATE]: (state, action) => ({
-        ...state,
-        page: action.payload.nextPage,
-    }),
```
paschalidi, line 130: Realised this is duplicated with one action 5 lines above it

src/core_modules/capture-core/reducers/descriptions/currentSelections.reducerDescriptionGetter.js
```diff
@@ -216,9 +217,20 @@ export const getCurrentSelectionsReducerDesc = (appUpdaters: Updaters) => create
         categories: undefined,
         categoriesMeta: undefined,
     }),
-    [searchPageActionTypes.FALLBACK_SEARCH_COMPLETED]: (state, { payload: { trackedEntityTypeId } }) => ({
+    [searchPageActionTypes.FALLBACK_SEARCH_COMPLETED]:
+      (state, { payload: { trackedEntityTypeId } }) => ({
+          ...state,
+          programId: undefined,
+          categories: undefined,
+          categoriesMeta: undefined,
```
paschalidi, line 225: Thought is a good idea to clean those when your selection is a `trackedEntityTypeId`

src/core_modules/capture-core/utils/url.js
```diff
@@ -1,30 +1,16 @@
...
-    }
-
     if (programId) {
         argArray.push(`programId=${programId}`);
     } else if (trackedEntityTypeId) {
         argArray.push(`trackedEntityTypeId=${trackedEntityTypeId}`);
     }
+    if (orgUnitId) {
+        argArray.push(`orgUnitId=${orgUnitId}`);
+    }
```
paschalidi, line 13: I didnt see a good reason why I was changing the order here so I reverted it.

src/core_modules/capture-core/components/TrackedEntityTypeSelector/TrackedEntityTypeSelector.component.js
```diff
@@ -60,11 +63,14 @@ const InfoOutlinedIconWithStyles = withStyles({
...
 export const TrackedEntityTypeSelectorPlain =
-  ({ classes, onSelect, selectedSearchScopeId }: Props) => {
+  ({ classes, onSelect }: Props) => {
       const trackedEntityTypesWithCorrelatedPrograms = useTrackedEntityTypesWithCorrelatedPrograms();
+      const dispatch = useDispatch();
+      const selectedSearchScopeId = useCurrentTrackedEntityTypeId();
 
       const handleSelectionChange = ({ selected }) => {
           onSelect(selected, scopeTypes.TRACKED_ENTITY_TYPE);
+          dispatch(setTrackedEntityTypeIdOnUrl({ trackedEntityTypeId: selected }));
```
JoakimSM, line 73: Not a big fan of putting this directly in the component file here. I have suggestion and you can tell my why that would not work: Remove the dispatch from here. Add a callback method in `NewPage.container` and call that from from https://github.com/dhis2/capture-app/blob/1b119d401fc52bec01368b5b43b98303bf7aefc2/src/core_modules/capture-core/components/Pages/New/NewPage.component.js#L74. For the `selectedSearchScopeId` on line 69 in this file, I would prefer receiving this from the parent component as you did earlier. 

Could be that I'm missing some reason why this would not work so please feel free to enlighten me 🙂

JoakimSM (commented):
Looks quite good. Made one comment, please have a look and correct me if I'm wrong 🙂

src/core_modules/capture-core/components/TrackedEntityTypeSelector/TrackedEntityTypeSelector.component.js
paschalidi: > Not a big fan of putting this directly in the component file here. I have suggestion and you can tell my why that would not work: Remove the dispatch from here. Add a callback method in NewPage.container and call that from from

I agree with what you say that this could be in a container. We had discussed before that we would like to keep separate the components from the containers, something I obviously dont do here. I will change that. 

One thing though. I think its more fitting for the `dispatch(setTrackedEntityTypeIdOnUrl({ trackedEntityTypeId: selected }));` to be part of the `TrackedEntityTypeSelector` module. As far as I can see we will always want the same redux action to be triggered here. Aka the `setTrackedEntityTypeIdOnUrl` action. 

So now you are right, this could also happen in the `NewPage.container` and be passed here as prop. However that indicates that we would like to to pass in different actions than the `setTrackedEntityTypeIdOnUrl` in the `TrackedEntityTypeSelector` component. I cant think of a use case for it. At least not with the current use of the `TrackedEntityTypeSelector`. We now use it in the search page and in the new page and in both cases we want the action to change the url and the action can be in both places the same. I would leave it as it is for now and if we find ourselves confronted with cases that the component needs to trigger extra actions then I would "open" it.

--- 

Now about the line 69 @JoakimSM in which file is it again?

JoakimSM: Aha, I didn't notice this file was reused. Alright, lets leave it as is for now then (apart from the container part).

JoakimSM (approved):

src/core_modules/capture-core/components/TrackedEntityTypeSelector/TrackedEntityTypeSelector.component.js
paschalidi: yes you right I will change the container part.

>  For the selectedSearchScopeId on line 69 in this file, I would prefer receiving this from the parent component as you did earlier.

Can you help me with that? Cant really find which `selectedSearchScopeId` you referring to

JoakimSM: Never mind, let's leave it as is 👍

PR #1231, opened 2020-11-26
refactor: exports components as constants and moves the files around

paschalidi:
@JoakimSM this is the first of the two part refactoring we talked about yesterday. In this one, the part 1, there are no functional changes. I am moving some files around and I am exporting most of the components I find as relevant as constants. I try to keep it into a minimum.

## Motivation
I move the Enrollment and the TrackedEntityType folders into the entry as you can see in the image since they are only being used inside that folder.

## Before & After comparison in a glance
<p float="left">
  
   
</p>


JoakimSM (approved):
LGTM

PR #1232, opened 2020-11-26
[part 02] refactors the Enrollment and TEI entries

paschalidi:
This is the part two of the https://github.com/dhis2/capture-app/pull/1231. 

Well, now arguably in this PR there are quite few things that happening. I will try to explain here the goal of this PR in my hopes to help you review it. 

The idea that I wanted to achieve is that I wanted to create two `DataEntries` one for the Enrollment registration and one for the Tei registration. I wanted these components to fulfil one requirement.

* be reusable out of the box. In the sense that someone can use them without having to trigger redux actions and write epics. The approach I am taking is that the components trigger these actions themselves since the actions are the same each time for our use cases.  

Every change I make in this PR is an attempt to achieve the above objective

Now to achive this two new components are created. These live inside the `DataEntries` and are called `EnrollmentRegistrationEntry` and `TeiRegistrationEntry`. Naming is obviously under discussion but what I try to communicate with the name name is that the component contains the `Registration` in the form of a `Data entry`. And this is possible either for `Enrollment` or for `TEI`


src/core_modules/capture-core/hooks/useCurrentOrgUnitInfo.js
```diff
@@ -1,10 +1,5 @@
 // @flow
 import { useSelector } from 'react-redux';
 
-export const useCurrentOrgUnitInfo = (): {|id: string, name: string, path: string|} =>
-    useSelector((
-        {
-            organisationUnitRoots: { captureRoots: { roots } },
-            currentSelections: { orgUnitId },
-        }) => roots.find(({ id }) => id === orgUnitId))
-    || {};
```
paschalidi, line 10: It looked to me this wasnt necessary so I went back on a simpler way of doing this

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/DataEntry/RegisterTeiDataEntry.container.js
```diff
@@ -1,23 +1,20 @@
...
-    ready: !state.newRelationshipRegisterTei.dataEntryIsLoading,
     error: state.newRelationshipRegisterTei.dataEntryError,
     programId: state.newRelationshipRegisterTei.programId,
 });
 
 const mapDispatchToProps = () => ({});
 
-// $FlowFixMe[missing-annot] automated comment
-export const RegisterTeiDataEntry = connect(mapStateToProps, mapDispatchToProps)(
-    withLoadingIndicator()(
```
paschalidi, line 18: With this change I am handling the loading on the DataEntry level. This is arguably almost the same so the change doesnt serve as an improvement. 

To understand why this change is here we will need to think like that. 
Now that the component itself triggers its initialisation, meaning adding the relevant information to redux we cant have the loading indicator depend on the redux state since it hasnt been yet initialised. Thus I move it on a lower level :)

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.epics.js
```diff
@@ -0,0 +1,42 @@
...
+export const startNewEnrollmentDataEntrySelfInitialisationEpic = (action$: InputObservable) =>
+    action$.pipe(
+        ofType(enrollmentRegistrationEntryActionTypes.TRACKER_PROGRAM_REGISTRATION_ENTRY_INITIALISATION_START),
+        pluck('payload'),
+        switchMap(({ selectedOrgUnitId, selectedScopeId: programId, dataEntryId, formFoundation }) => {
+            if (selectedOrgUnitId) {
+                let trackerProgram: ?TrackerProgram;
+                try {
+                    trackerProgram = getTrackerProgramThrowIfNotFound(programId);
+                } catch (error) {
```
JoakimSM, line 23: This is of course good practice, but if this fails then something is so seriously wrong that in my opinion it is ok to not catch it (I used to do exactly what you do here). I'm just saying that I will continue to be pragmatic when I'm using for example `getTrackerProgramThrowIfNotFound` and not spend time writing these. I won't complaint that you do it of course, but now you know why you won't see me do it :)

src/core_modules/capture-core/components/DataEntries/TeiRegistrationEntry/TeiRegistrationEntry.component.js
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+import { useScopeInfo } from '../../../hooks/useScopeInfo';
+import { scopeTypes } from '../../../metaData';
+import { TrackedEntityInstanceDataEntry } from '../TrackedEntityInstance';
+import { useCurrentOrgUnitInfo } from '../../../hooks/useCurrentOrgUnitInfo';
+import type { OwnProps } from './TeiRegistrationEntry.types';
+import { useRegistrationFormInfoForSelectedScope } from '../common/useRegistrationFormInfoForSelectedScope';
+
+
+export const TeiRegistrationEntryComponent = ({ selectedScopeId, id, ...rest }: OwnProps) => {
```
JoakimSM, line 11: `...rest` is fine, but I suggest we stick to `passOnProps` like the rest of the app.

JoakimSM (approved):
Looks good @paschalidi 👍

src/core_modules/capture-core/components/DataEntries/TeiRegistrationEntry/TeiRegistrationEntry.component.js
paschalidi: You right there isnt any reason from my side to  changed it.

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.epics.js
paschalidi: This is basically your code :) This is why it is there. I would prefer to remove it from here. We have the error boundary from the app platform now and it will catch it. You mind if I remove it?

JoakimSM: This is not a biggie for me so do whatever you feel is best here :)

PR #1260, opened 2020-12-03
feat: DHIS2-9652 with save buttons

paschalidi:
Hey here is the https://jira.dhis2.org/browse/DHIS2-9652👋  Again another not so small PR. Apologies I could see the best way to split this up. Will start with smaller chunks from the next one :) 

## What am I changing
What happens here is that we are reusing the components we created from the last two PRs and adding the buttons on the bottom. You can say it is this PR that the refactoring from the previous two PR comes together.

Now we are using the `EnrollmentRegistrationEntry` and `TeiRegistrationEntry` across the new registration page and the add relationship part of the application.

## How it looks
image


src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.types.js
```diff
@@ -1,5 +1,16 @@
...
+  enrollmentMetadata: RegistrationFormMetadata,
   selectedScopeId: string,
-  id: string
+  saveButtonText: string,
+  fieldOptions?: Object,
+  onSave: (dataEntryId: string, itemId: string, formFoundation: RenderFoundation) => void,
+  onPostProcessErrorMessage: Function,
+  onGetUnsavedAttributeValues: Function,
+  onUpdateField: Function,
+  onStartAsyncUpdateField: Function,
```
paschalidi, line 15: Here I am trying to explicitly define the properties this components needs in order to be rendered. 
In the future when someone will need to use it they can look in this signature

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.component.js
```diff
@@ -6,25 +6,66 @@ import { EnrollmentDataEntry } from '../Enrollment';
...
+                      {
+                          onSave &&
+                          <Button
+                              dataTest="dhis2-capture-create-and-link-button"
+                              primary
+                              onClick={onSave}
+                              className={classes.marginTop}
+                          >
+                              {saveButtonText}
+                          </Button>
```
paschalidi, line 58: Displaying the buttons only when there is an onSave function

src/core_modules/capture-core/components/InfoIconText/InfoIconText.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
paschalidi, line 1: image
I am creating this component since I find myself reusing it

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
```diff
@@ -1,30 +1,74 @@
...
+        {orgUnitName && <>{' '}{i18n.t('in')} <b>{orgUnitName}</b></>}.
+    </>);
+
+
+const translatedTextWithStylesForTei = (trackedEntityName, orgUnitName) =>
+    (<>
+        {i18n.t('Saving a {{trackedEntityName}}', { trackedEntityName })} <b>{i18n.t('without')}</b> {i18n.t('enrollment')}
+        {orgUnitName && <>{' '}{i18n.t('in')} <b>{orgUnitName}</b></>}.{' '}
+        {i18n.t('Enroll in a program by selecting a program from the top bar.')}
+    </>);
```
paschalidi, line 25: This text if you see on the design has some bolds and some not bolds. This is why I am hacking it like this.

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
```diff
@@ -1,30 +1,74 @@
...
+                        id={dataEntryId}
+                        selectedScopeId={selectedScopeId}
+                        enrollmentMetadata={registrationMetaData}
+                        saveButtonText={'Save new'}
+                        onSave={() => alert('onSave will save in the future')}
+                        onGetUnsavedAttributeValues={() => console.log('onGetUnsavedAttributeValues will be here in the future in the future')}
+                        onPostProcessErrorMessage={() => console.log('onPostProcessErrorMessage will be here in the future in the future')}
+                        onUpdateField={() => console.log('onUpdateField will be here in the future in the future')}
+                        onStartAsyncUpdateField={() => console.log('onStartAsyncUpdateField will be here in the future in the future')}
+                    />
```
paschalidi, line 48: Here I would say you will see the benefit of creating the `EnrollmentRegistrationEntry` and `TeiRegistrationEntry ` components.

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
JoakimSM: I know I've used these bold parts in the new single event page, but I've later realised the concatenations will cause problems, at least for rtl languages. I suggest we keep it nonbold for now? @Bekkalizer , any insights?

JoakimSM (approved):
Looks good @paschalidi 👍 please have a look at my comments before you merge.

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.types.js
paschalidi: Good point @JoakimSM. Yep for me the naming here is under discussion as well. 
What I am trying to do with the `OwnProps` is to have a type for properties that the user passes directly on the component when they use it.

Eg.
```
<Button type='one' size='lg'>
Text
</Button>
```
On the example above `OwnProps` would be the `type` and the `size`. However sometimes you might have the `ContainerProps` which are props that are passed on the container level. My only point is that I would want to differentiate between what I call `OwnProps` and the rest of props a component might have. This is because then flow can tell you what props you are supposed to pass in the component when you try to use it. 

Again I agree we would like to align on a name here. I stole my name from the redux world, look [here](https://react-redux.js.org/using-react-redux/connect-mapstate).
---

Now if we agree on communicating the same thing we can be creative with a name. What do you think, would you find it useful to differentiate between the props that we pass on the component and the props that the component takes from redux?

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.component.js
```diff
@@ -1,30 +1,70 @@
...
-};
+const EnrollmentRegistrationEntryPlain =
+  ({
+      id,
+      selectedScopeId,
+      enrollmentMetadata,
+      saveButtonText,
+      classes,
+      onSave,
+      ...rest
```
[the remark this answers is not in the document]
paschalidi: Right. I will make a PR in the end renaming all the `rest` to `onProps`.

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.types.js
JoakimSM: Thanks for explaining. Our use cases are a bit different. `OwnProps` is good for what you're trying to achieve here 👍

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
paschalidi: Does it make sense to make a bug ticket for this? Then I could merge this and we come back to this in isolation.

PR #1271, opened 2020-12-07
fix: DHIS2-10070 improves rules engine performance

paschalidi:
This works on this ticket https://jira.dhis2.org/browse/DHIS2-10070. 

Here is how much this would take in the previous version. If you would like me to explain in more detail how to get this view, please let me know and we can get into a call
image

This is the same action now 
image

### DIFF
--- a/src/core_modules/capture-core/rules/actionsCreator/inputHelpers.js
+++ b/src/core_modules/capture-core/rules/actionsCreator/inputHelpers.js
@@ -31,11 +31,10 @@ function convertFormValuesToClientValues(formValues: {[key: string]: any}, rende
 
 function getValidFormValues(formValues: { [key: string]: any }, fieldsValidation: { [key: string]: boolean }) {
     return Object.keys(formValues)
-        .reduce((accValidFormValues, key) => {
-            const isValid = fieldsValidation[key];
-            accValidFormValues[key] = isValid ? formValues[key] : null;
-            return accValidFormValues;
-        }, {});
+        .filter(key => fieldsValidation[key])
+        .reduce((accValidFormValues, key) => (
+            { ...accValidFormValues, [key]: formValues[key] }
+        ), {});
 }
 
 function getFieldsValidationForForm(sectionsFieldsUI: Object, formId: string) {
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/rules/actionsCreator/inputHelpers.js
```diff
@@ -31,11 +31,10 @@ function convertFormValuesToClientValues(formValues: {[key: string]: any}, rende
 
 function getValidFormValues(formValues: { [key: string]: any }, fieldsValidation: { [key: string]: boolean }) {
     return Object.keys(formValues)
-        .reduce((accValidFormValues, key) => {
-            const isValid = fieldsValidation[key];
-            accValidFormValues[key] = isValid ? formValues[key] : null;
-            return accValidFormValues;
-        }, {});
+        .filter(key => fieldsValidation[key])
```
paschalidi, line 34: What used to happen before, is that we iterated over all the form values and returned `null` when there isn't a value. I return now elements that have values. 

Please lets try to think if this will cause any problems in the ling term.

src/core_modules/capture-core/rules/actionsCreator/postProcessRulesEffects.js
```diff
@@ -116,12 +116,15 @@ function filterFieldsHideEffects(
...
     return elementIds
         .filter((elementId) => {
-            const element = foundation.getElement(elementId);
-            const dataCompulsory = element && element.compulsory;
+            const { compulsory: elementIsCompulsory } =
+                compulsoryElements.find(({ id }) => id === elementId) || {};
+
             const compulsoryEffect = makeCompulsoryEffects[elementId];
-            return !(dataCompulsory || compulsoryEffect);
+            return !(elementIsCompulsory || compulsoryEffect);
```
paschalidi, line 127: What we did here is look in _all_ the elements and get back the id with the `elementId`.  This specific action costs us time and made the application slower. 

Also is important to understand we care only about the `compulsory` elements. 

What I do now is filter the elements and keep only the ones that are `compulsory` which are usually few and the `find` in these the one with the `elementId` we are looking for.

src/core_modules/capture-core/rules/actionsCreator/inputHelpers.js
Bekkalizer: LGTM - I would test wether the default value assignment works as before. Check that a "Yes/No" data element left blank is evaluated to "No", or that a Numeric data element defaults to 0.

Bekkalizer (approved):
Looks good - one comment on a thing you can test functionally before we merge

PR #1297, opened 2020-12-10
chore: DHIS2-10111 adds `restrictToCaptureScope=true`

paschalidi:
@JoakimSM and @Bekkalizer hey. 

This is taking care of the https://jira.dhis2.org/browse/DHIS2-10111. 

## How to test this? 

1. In maintenance app for a program, lets say child program and select all units _but_ one.
image

2. in the users app go and create a user. Add to the user a capture org unit that is _not_ the one you have added to the child program. Also give the user Facility tracker access.
image

3. save the user.
4. clear the cache.
5. log in with the user you created. 

you can see in the network tab that the orgUnits for that program that have been fetched are empty. 
image

## How it used to be?

It would fetch all the org units a program has been assigned to
image

### DIFF
--- a/src/core_modules/capture-core/metaDataStoreLoaders/programs/queries/queryProgramsOutline.js
+++ b/src/core_modules/capture-core/metaDataStoreLoaders/programs/queries/queryProgramsOutline.js
@@ -5,6 +5,7 @@ export const queryProgramsOutline = async (): Promise<Array<Object>> => {
     const specification = {
         resource: 'programs',
         params: {
+            restrictToCaptureScope: true,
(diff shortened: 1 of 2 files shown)


JoakimSM (approved):
👍 Interestingly enough, your test write up describes an example where the program shouldn't be part of the response at all. Nothing wrong with the client code, but potential for improvement on the backend later.

paschalidi:
I know... I realised that myself.

ameenhere:
Just to add to this. The program now wont be part of the response at all if its configured as per your screenshots and aligning to your comments. It was like that for categoryOptions, but was missed for programs.

paschalidi:
Cool stuff @ameenhere thanks for the update

PR #1314, opened 2020-12-13
feat: [part 02] DHIS2-9637 adds new button and unifies the urls | extracting the page to a component

paschalidi:
👋 Hey Joakim. This might look a bit bloated but actually is much smaller than it looks. Most of the changes are here because I moved some files and therefore the imports have changed. :) 

### What is this PR about?
This is the second part of the three I was talking about [here](https://github.com/dhis2/capture-app/pull/1312). 

It basically doesnt have any functional changes. What I am trying to achieve here is to extract the `SingleEventRegistrationEntryComponent` in the `DataEntries` folder. 

The idea is that we can have three components in the data entries. 
1. one for the tracked entity instances
2. one for the tracker programs 
3. one for the event programs
image

Note please that this isnt functionally sane yet.


src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/NewEventDataEntryWrapper.component.js
```diff
@@ -0,0 +1,84 @@
+// @flow
```
paschalidi, line 1: This whole file isn't new. The original of it is called `NewEvent/DataEntryWrapper/NewEventDataEntryWrapper.component.js`.Github dont recognise this as a incremental change of the file. So I hope you knowing that this is not a complete new component that I built but a copied version of an existing one, will help you review this file.

src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/SingleEventRegistrationEntry.component.js
```diff
@@ -0,0 +1,32 @@
+// @flow
```
paschalidi, line 1: Same as before this file isnt a complete rewrite but only a copied version of the file `Pages/NewEvent/SelectionsComplete/SelectionsComplete.component.js`

JoakimSM (approved):
Looks good @paschalidi

PR #1317, opened 2020-12-14
feat: DHIS2-9276 #2.1 - api utils for epics

JoakimSM:

Main purpose here is to make methods/data from the app-runtime dataEngine available to the epics. You will see it in use in the next PR. Methods/data derived from the dataEngine will be available to the epics through a third parameter (added using the dependencies argument in `getStore.js`). I'm aware that we could potentially have made the dataEngine available through a global, but wanted to test this approach instead for a couple of reasons 1) The dataEngine is already available to the React components through the hook `useDataEngine` 2) Potential Jest tests for the epics should be easier to write when the dataEngine methods are made available through a param.

Sorry about all the changes in the `metaDataStoreLoaders` folder, could have been its own PR, but it was tied to changes I did so opted to do it here. The change here is that I'm passing in a function that queries for a single resource only, since querying for multiple resources are never used.


src/core_modules/capture-core/flow/app-runtime_v2.x.x.js
```diff
@@ -0,0 +1,97 @@
...
+    declare type FetchErrorTypeName = 'network' | 'unknown' | 'access' | 'aborted';
+    declare type FetchErrorDetails = { message?: string };
+
+    declare interface FetchErrorPayload {
+        type: FetchErrorTypeName,
+        details?: FetchErrorDetails,
+        message: string,
+    }
+
+    declare class FetchError extends Error implements FetchErrorPayload {
```
JoakimSM, line 43: LGTM made a complaint here, but I believe the code should be ok.

src/store/getStore.js
```diff
@@ -36,14 +35,14 @@ export function getStore(
...
-        effect: getEffectReconciler(onApiMutate),
+        effect: getEffectReconciler(apiUtils.mutate),
         persistCallback: onRehydrated,
         queue: queueConfig,
         persistOptions: getPersistOptions(),
     });
 
     const epicMiddleware = createEpicMiddleware({
-        dependencies: {},
+        dependencies: apiUtils,
```
JoakimSM, line 45: `apiUtils` made available to the epics through the third parameter.

src/components/AppLoader/AppLoader.component.js
```diff
@@ -15,30 +16,39 @@ type Props = {
...
+        mutate: dataEngine.mutate.bind(dataEngine),
+        absoluteApiPath: buildUrl(dataEngine.link.baseUrl, dataEngine.link.apiPath),
+    }), [dataEngine]);
+};
+
 const AppLoader = (props: Props) => {
     const { onRunApp, onCacheExpired } = props;
     const [loadError, setLoadError] = React.useState(null);
-    const dataEngine = useDataEngine();
+    const apiUtils = useApiUtils();
```
paschalidi, line 31: very nitty 
```suggestion
    const { querySingleResource, querySingleResource, apiUtils }  = useApiUtils();
```

src/core_modules/capture-core/metaDataStoreLoaders/IOUtils/IOUtils.types.js
```diff
@@ -1,42 +1,37 @@
...
     iterationSize?: number,
-};
+|};
 
-export type QueryRecursivelyOptions = {
+export type QueryRecursivelyOptions = {|
     convertQueryResponse?: ConvertQueryResponseFn,
     pageSize?: number,
-};
+|};
```
paschalidi, line 37: ⭐ super

paschalidi (approved):
Hey thanks for explaining what is going on here. 

Keen to see how we can use the `apiUtils` in the epics.

Great work 👍

src/components/AppLoader/AppLoader.component.js
JoakimSM: Makes sense 👍 Changing.

PR #1325, opened 2020-12-16
feat: DHIS2-9276 #6 - templates order

JoakimSM:

Enables custom order for the templates (default is alphabetically)


paschalidi (approved):
LGTM!

PR #1328, opened 2020-12-16
feat: DHIS2-9276 #9 - new text for filter buttons

JoakimSM:

New button text for the filter buttons.


paschalidi (approved):
On to the point!! 👍

PR #1330, opened 2020-12-16
feat: DHIS2-9276 #11 - minor UI changes

JoakimSM:
Fixing some minor UI discrepancies. Fully aware that some of the files I touched have potential for improvement, only fixed the  parts relevant to my changes here.

Moving a lot of code around to accommodate the few changes I did. 

Changes:

| Before  | After |
| ------------- | ------------- |
| image | image |
| image | image  |
| image  | image |


src/core_modules/capture-ui/internal/SelectionBoxes/SingleSelectionBoxes/SingleSelectBox/singleSelectBox.module.css
```diff
@@ -22,7 +22,7 @@
 .iconContainer {
     display: flex;
     align-items: center;
-    padding: 2px 12px 2px 12px;
+    padding: 10px 10px 10px 10px;
```
paschalidi, line 25: hey I know we are using 10px when we want to give paddings. There is a convention between designers and I think Joe is following it as well, where they prefer using derivatives of 4. Like 4px 8px 12px etc. The logic behind that is that it gives you more options in the long run to choose between paddings. Also is a good rule of thumb when we developers try to give padding on something that we dont have design for.

You can see if this applies for us here or if not please feel free to resolve this comment

src/core_modules/capture-core/components/FormFields/Options/SingleSelectBoxes/SingleSelectBoxes.component.js
```diff
@@ -7,29 +7,19 @@ import FormLabel from '@material-ui/core/FormLabel';
...
-        label: string,
+const styles = ({ typography, palette }) => ({
+    label: typography.formFieldTitle,
+    iconSelected: {
+        fill: palette.secondary.main,
     },
-    style?: ?Object,
-};
+    iconDeselected: {
+        fill: palette.grey[700],
```
paschalidi, line 20: Not really a blocker but would be cool if you can have a look. I think we could use some variation of grey which comes from the `@dhis2/ui` since we want slowly to converge into using that lib. 

You could use it as such `colors.grey700`

Here is the colors provided to us https://github.com/dhis2/ui/blob/master/packages/constants/src/colors.js#L69

paschalidi (approved):
hey looks good!

src/core_modules/capture-ui/internal/SelectionBoxes/SingleSelectionBoxes/SingleSelectBox/singleSelectBox.module.css
JoakimSM: Thanks, this makes sense. Changing to 12.

src/core_modules/capture-core/components/FormFields/Options/SingleSelectBoxes/SingleSelectBoxes.component.js
JoakimSM: Thanks, agree, changing.

PR #1331, opened 2020-12-16
fix: DHIS2-5490 rules engine numeric value parsing

JoakimSM:
[DHIS2-5490](https://jira.dhis2.org/browse/DHIS2-5490)

In rules engine: Convert empty source variables for numeric value types to 0.


src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
```diff
@@ -51,28 +52,36 @@ class RulesValueConverter implements IConvertInputRulesValue {
...
+    convertPercentage(value: string): number {
+        if (!value) {
+            return 0;
+        }
+        const numberValue = parseNumber(value);
+        if (isNaN(numberValue)) {
+            return 0;
+        }
+
+        return numberValue / 100;
```
paschalidi, line 84: Wondering if we wanna define a fixed amount of decimals here. to avoid cases like `3.213123%`

[toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed ) could be a quick way to do this.

paschalidi (commented):
Hey looks good to me, and thanks for the annotation!. 👍  One comment only.

src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
JoakimSM: I think, since this value will be used in calculations, we shouldn't manipulate it. We should rather supply a d2 helper function to be used in the rules if needed (in the same vein as d2:ceil, d2:floor, d2:round).

paschalidi (approved):
LGTM 👍

Bekkalizer (approved):
LGTM!

PR #1373, opened 2021-01-04
feat: [part 4] DHIS2-10179 displays no access in view message 

paschalidi:
ticket https://jira.dhis2.org/browse/DHIS2-10179

This is about restricting access to the user. For example it might be that a Tracker Program or a Tracked entity type is configured with data access false. In that case the user shouldnt be able to see the registration form but they see a message explaining the reason they are restricted. 

For example if a user doesnt has access on the tetype Person they see and they navigate to the `new/trackedEntityTypeId=nEenWmSyUEp&orgUnitId=DiszpKrYNg8`
image

Also I ve exported the access component so that we could reuse its styles.


JoakimSM (approved):
LGTM 👍

PR #1374, opened 2021-01-04
chore: [part 5] adds e2e tests

paschalidi:
And here finally I am adding the e2e tests


src/components/AppStart/unload/addBeforeUnloadEventListener.js
```diff
@@ -1,13 +1,11 @@
...
 import cleanUpCommon from 'capture-core/cleanUp/cleanUp';
-// import { cleanUp } from './cleanUp';
 
 export default function addBeforeUnloadEventListener(store: ReduxStore) {
     window.addEventListener('beforeunload', async (e) => {
         cleanUpCommon(store);
-        // cleanUp(store); // should move clean up from core to app!
 
-        if (store.value.offline.outbox.length > 0) {
+        if (store.getState().offline.outbox.length > 0) {
```
paschalidi, line 8: The app would throw an error because of this. I had refactored this some time ago as part of moving the app to rxjs v6, but I got this one mixed. Therefore basically this is reverting my change from back then. 

It is not really part of this PR but I fixed it in the process of understanding why some tests of mine were failing. As it seemed it was unrelated to this but I thought to add it in anyway since it is basically fixes the situation here.

JoakimSM (approved):
LGTM 👍

PR #1388, opened 2021-01-07
feat: DHIS2-10161 [part 02] Create enrollment page with routing | fetches enrollments

paschalidi:
### What is this PR about?

Basically we are fetching the enrollments and storing them on redux. This PR works out the case that the user types in the url the all the parameters needed. In other words this will work _only_ for urls that provide all the following information http://localhost:3000/#/enrollment?programId=uy2gU8kT1jF&orgUnitId=DiszpKrYNg8&teiId=EaOyKGOIGRp&enrollmentId=k8ZtDibWX5Y 

In the next PRs we will introduce the usecases where you can load the whole page using only the `enrollmentId` but for now this is out of the scope of _this_ PR.

Expect to see something between these lines when you run this code locally. 

### The next PR. 

The next pr will add styling in the locked selector and also include the persons name _and_ the enrollment id in the locked selector.


src/reducers/descriptions/trackerCapture.reducerDescriptions.js
```diff
@@ -74,16 +74,15 @@ import { currentSelectionsReducerDesc } from './currentSelections.reducerDescrip
 import { feedbackDesc } from './feedback.reducerDescription';
 import { activePageDesc } from '../../core_modules/capture-core/reducers/descriptions/activePage.reducerDescription';
 import { searchPageDesc } from '../../core_modules/capture-core/reducers/descriptions/searchPage.reducerDescription';
+import { enrollmentPageDesc } from '../../core_modules/capture-core/reducers/descriptions/enrollmentPage.reducerDescription';
 import { newPageDesc } from '../../core_modules/capture-core/reducers/descriptions/newPage.reducerDescription';
 
 export default [
```
paschalidi, line 80: I ordered alphabetically here, since it was already there I thought is good if we have it complete. 
Sorry this is out of scope for this PR.

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -0,0 +1,44 @@
...
+                                        moment.utc(a.enrollmentDate).diff(moment.utc(b.enrollmentDate)));
+
+                                    return successfulFetchingEnrollmentPageInformationFromUrl({
+                                        selectedName,
+                                        enrollmentsSortedByDate,
+                                    });
+                                }),
+                            )),
+                    startWith(showLoadingViewOnEnrollmentPage()),
+                    catchError(() => of(showErrorViewOnEnrollmentPage())),
```
paschalidi, line 41: This bit will be reworked a lot in the future PRs. Perhaps it doesnt look so amazing in this one, but I promise I will be making it look better when we also start fetching these information using the `teiId`. Now we only fetch using the `enrollmentId`

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -0,0 +1,44 @@
...
+import { getApi } from '../../../d2';
+import {
+    enrollmentPageActionTypes,
+    showErrorViewOnEnrollmentPage,
+    showLoadingViewOnEnrollmentPage,
+    successfulFetchingEnrollmentPageInformationFromUrl,
+} from './EnrollmentPage.actions';
+
+const fetchEnrollment = id => getApi().get(`enrollments/${id}`, { fields: 'trackedEntityInstance,program,orgUnit' });
+const fetchTrackedEntityInstance = id => getApi().get(`trackedEntityInstances/${id}`, { fields: 'attributes,enrollments' });
```
JoakimSM, line 15: Reminder to use the data engine here instead

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
```diff
@@ -0,0 +1,34 @@
...
+export const EnrollmentPage: ComponentType<{||}> = () => {
+    const dispatch = useDispatch();
+
+    const enrollmentId: string =
+      useSelector(({ router: { location: { query } } }) => query.enrollmentId);
+    const enrollments: Object =
+      useSelector(({ enrollmentPage }) => enrollmentPage.enrollments);
+
+    useEffect(() => {
+        const enrollmentIsNotPreset = !(enrollments && enrollments.some(({ enrollment }) => enrollmentId === enrollment));
```
JoakimSM, line 18: probably dealt with later, but I see a loop forming if you query an enrollmentId that doesn't exist.

JoakimSM (commented):
Looks good 👍 Would be perfect if you use the data engine instead of the old d2 library.

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -0,0 +1,44 @@
...
+        flatMap(() => {
+            const { query: { enrollmentId } } = store.value.router.location;
+
+            return from(fetchEnrollment(enrollmentId))
+                .pipe(
+                    flatMap(({ trackedEntityInstance }) =>
+                        from(fetchTrackedEntityInstance(trackedEntityInstance))
+                            .pipe(
+                                map(({ attributes, enrollments }) => {
+                                    const selectedName = attributes.reduce((acc, { value: dataElementValue }) =>
```
JoakimSM, line 29: We need to use the metadata when computing the name. We should display the attribute value only if the "display in list" flag is set. We should also use the attribute order from the metadata.

paschalidi: Okay I was wondering why I can see that display on list flag. Then I thought that it might have been that on the backend they return only the values that are good for display.. 

Okay then you mean that we should be using the data we have stored in memory right?

JoakimSM: Yes 👍

JoakimSM (approved):

PR #1393, opened 2021-01-07
feat: DHIS2-10161 [part 03] Create enrollment page with routing | styles & adds SingleLockedSelect components on locked selector

paschalidi:
### What is this PR about?

In this PR we bring the locked selector closer to the Joe's designs _and_ also including the enrollment id and persons name on the locked selector. Apologies that I could keep this shorter. 

Basically now we would be having a fully functional locked selector that looks somewhat like this 
image

Functionality: 
1. Removing the program will keep us in the same page. You can then select another program
image

2. You can select enrollments from a list of enrollments 
image

3. Removing the teiId will bring you back to main page.

Keep in mind please that you still need to use the full url to reach this page as shown in the images.

I just remembered that we wanted to remove the text `Selected Tracked Entity Type` from the locked selector and display the actual name of the type. I am creating a new ticker for this bit. https://jira.dhis2.org/browse/DHIS2-10370


src/core_modules/capture-core/components/LockedSelector/QuickSelector/Program/ProgramSelector.component.js
```diff
@@ -34,11 +37,10 @@ const styles = (theme: Theme) => ({
...
     },
     selectedText: {
-        marginTop: 5,
-        marginBottom: 5,
-        marginLeft: 5,
+        marginTop: 6,
+        marginBottom: 4,
         padding: 5,
-        borderLeft: '2px solid #71a4f8',
+        borderLeft: '2px solid #147cd7',
```
JoakimSM, line 43: `colors.blue600`?

JoakimSM (commented):
Code looks good 👍

On a more general note, I did observe some weird behaviour in the main page when using the selectors. For example, if I select a program and then select an org unit with no access to the program, the page crashes. The behaviour should be that the program is deselected.

One other thing a noticed is when I deselect the tei from the enrolment page, the question mark is missing in the main page link (and the list is not loaded).

Overall looks good, but great if you could look into those issues.

src/core_modules/capture-core/components/LockedSelector/QuickSelector/Program/ProgramSelector.component.js
paschalidi: makes sense yeah. 👍

PR #1394, opened 2021-01-07
feat: DHIS2-10161 [part 04] Create enrollment page with routing | completes chain with enrollment logic

paschalidi:
### What is this PR about? 

It derives from _only_ the enrollment id all the information needed to fetch the page. In other words you can start using urls that look like this  http://localhost:3000/#/enrollment?enrollmentId=k8ZtDibWX5Y


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
```diff
@@ -21,8 +23,11 @@ export const EnrollmentPage: ComponentType<{||}> = () => {
     [
         enrollments,
         enrollmentId,
+        teiId,
```
JoakimSM, line 26: Related: When I supply only the enrollmentId in the url; I can see that data from the enrollment/tei endpoints are requested twice. Is this necessary (something I don't grasp here)?

paschalidi: This shouldnt be the case. Let me double check please and I will come back to you

paschalidi: Things change quite a bit when I realised that we also have to fetch data just by using the `teiId`. So most likely this is been taken care of in one of the next PRs. However I will check and let you know for sure!

JoakimSM (approved):

PR #1409, opened 2021-01-12
chore: removes react-autosuggest

paschalidi:
Removes the `react-autosuggest` package since it is not being used in our project as far as I can see.


JoakimSM (approved):
LGTM

PR #1413, opened 2021-01-13
fix: TECH-522 redux offline retry handling

JoakimSM:
[TECH-522](https://jira.dhis2.org/browse/TECH-522)

After moving to the app runtime data engine, Redux Offline tries to resend requests that the response dictates should only be sent once. For example of you get a 409 error as a response, the request should not be resent. This was due to the response object returned from the data engine being different than the old one.

Note: I used optional chaining in this PR. Optional chaining should be supported since create-react-app (react-scripts) 3.3.0 [v3.3.0](https://github.com/facebook/create-react-app/releases/tag/v3.3.0) We use 3.4.1 at the moment.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -13,3 +13,4 @@ src/core_modules/capture-core/flow/
 module.name_mapper='^capture-core' ->'<PROJECT_ROOT>/src/core_modules/capture-core'
 module.name_mapper='^capture-ui' ->'<PROJECT_ROOT>/src/core_modules/capture-ui'
 module.name_mapper='^capture-core-utils' ->'<PROJECT_ROOT>/src/core_modules/capture-core-utils'
+esproposal.optional_chaining=enable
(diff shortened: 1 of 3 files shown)


.flowconfig
```diff
@@ -12,4 +12,5 @@ src/core_modules/capture-core/flow/
 [options]
 module.name_mapper='^capture-core' ->'<PROJECT_ROOT>/src/core_modules/capture-core'
 module.name_mapper='^capture-ui' ->'<PROJECT_ROOT>/src/core_modules/capture-ui'
-module.name_mapper='^capture-core-utils' ->'<PROJECT_ROOT>/src/core_modules/capture-core-utils'
\ No newline at end of file
+module.name_mapper='^capture-core-utils' ->'<PROJECT_ROOT>/src/core_modules/capture-core-utils'
+esproposal.optional_chaining=enable
```
paschalidi, line 16: yep thats cool feature! thanks for adding it

paschalidi (approved):
looks good! thanks for putting this together

PR #1424, opened 2021-01-14
feat: DHIS2-10161 [part 05]  Create enrollment page with routing | completes fetch from teiId AND enrollmentId

paschalidi:
### What is this PR about? 

Now you can load the page using only `teiId` as shown in the image. You could type in a url that is something between these lines. http://localhost:3000/#/enrollment?teiId=EaOyKGOIGRp
image

Then you can start by selecting program and you enrollment id. 

### Next PR will bring meaningful messages explaining the user what actions need to take in order to view the enrollment page. Things like
* You need to select program or 
* You need to select enrollment id 

etc


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.js
```diff
@@ -2,30 +2,39 @@
...
+    INFORMATION_FETCH: 'EnrollmentPage.Fetch',
+    INFORMATION_USING_TEI_ID_FETCH: 'EnrollmentPage.StartFetchingUsingTeiId',
+    INFORMATION_USING_ENROLLMENT_ID_FETCH: 'EnrollmentPage.StartFetchingUsingEnrollmentId',
+    INFORMATION_LOADING_FETCH: 'EnrollmentPage.LoadingOnFetching',
+    INFORMATION_ERROR_FETCH: 'EnrollmentPage.ErrorOnFetching',
+    INFORMATION_SUCCESS_FETCH: 'EnrollmentPage.SuccessOnFetching',
+
+    PAGE_OPEN: 'EnrollmentPage.Open',
+    PAGE_CLEAN: 'EnrollmentPage.CleanOnUnmount',
+    DEFAULT_VIEW: 'EnrollmentPage.DefaultView',
```
paschalidi, line 14: Renamed those a little. My logic is now that we are exporting as const the  action types aka `enrollmentPageActionTypes` we dont have to repeat in every action the `ENROLLMENT_PAGE`, because usually it will be that we use these actions in a way similar to this `enrollmentPageActionTypes.INFORMATION_FETCH`. This covers already where the action is coming from. 

Before we had the `enrollmentPageActionTypes.ENROLLMENT_PAGE_INFORMATION_FETCH`. Now that way we repeat ourselves. IMO

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -11,56 +12,97 @@ import {
...
 } from './EnrollmentPage.actions';
 import { urlArguments } from '../../../utils/url';
 
 const fetchEnrollment = id => getApi().get(`enrollments/${id}`, { fields: 'trackedEntityInstance,program,orgUnit' });
 const fetchTrackedEntityInstance = id => getApi().get(`trackedEntityInstances/${id}`, { fields: 'attributes,enrollments' });
 
+const sortByDate = (enrollments = []) => enrollments.sort((a, b) =>
+    moment.utc(b.enrollmentDate).diff(moment.utc(a.enrollmentDate)));
+const deriveSelectedName = (attributes = {}) => attributes.reduce((acc, { value: dataElementValue }) =>
+    (acc ? `${acc} ${dataElementValue}` : dataElementValue), '');
```
paschalidi, line 26: Still need to sort this out, and filter only the first two as we discussed. I will do together with the https://jira.dhis2.org/browse/DHIS2-10370

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -11,56 +12,97 @@ import {
...
+
+
+const fetchTeiStream = teiId => from(fetchTrackedEntityInstance(teiId))
+    .pipe(
+        map(({ attributes, enrollments }) => {
+            // todo this is not scaling when you have many attributes the name will be huge
```
paschalidi, line 32: Still need to filter here for the first two of the list as we discused so that we display some meaningful name. 
I will tackle this together with this subtask https://jira.dhis2.org/browse/DHIS2-10370

JoakimSM (approved):

PR #1438, opened 2021-01-18
fix: TECH-522 rework of redux offline retry handling

JoakimSM:
- reworked fix for redux offline retry handling after moving to app runtime dataengine


src/core_modules/capture-core/flow/app-runtime_v2.x.x.js
```diff
@@ -0,0 +1,121 @@
+declare module '@dhis2/app-runtime' {
```
JoakimSM, line 1: New code in this file is only line 97 - 120. See next comment.

src/core_modules/capture-core/trackerOffline/getEffectReconciler.js
```diff
@@ -0,0 +1,16 @@
+// @flow
```
JoakimSM, line 1: This file is old code. Moved to its own file.

src/core_modules/capture-core/trackerOffline/queueConfig.js
```diff
@@ -1,20 +1,6 @@
 // @flow
```
JoakimSM, line 1: This is old code, extracted `queueConfig` to its own file.

.eslintrc
```diff
@@ -43,7 +43,7 @@
                 "ignorePattern": "^\\s*var\\s.+=\\s*require\\s*\\("
             }
         ],
-        "complexity": [1, 5],
+        "complexity": [1, 8],
```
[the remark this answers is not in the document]
paschalidi: 8 looks good to me. I also have problems sometimes with this complexity warning

src/core_modules/capture-core/trackerOffline/shouldDiscard.js
```diff
@@ -0,0 +1,36 @@
...
+    const statusCodeSearchResult = message.match(/\(\d*\)$/);
+    if (statusCodeSearchResult) {
+        const statusCode = statusCodeSearchResult[0].replace(/[()]/g, '');
+        return Boolean(statusCode) && ![408, 429, 502, 503, 504].includes(statusCode);
+    }
+    return false;
+};
+
+export const shouldDiscard = (error?: ApiError) => {
+    if (!error || !error.type) {
```
paschalidi, line 22: just curious if you could do something like
```suggestion
    if (!error?.type) {
```

JoakimSM: Nice! That should work.

PR #1442, opened 2021-01-19
test: DHIS2-9276 cypress tests

JoakimSM:
e2e tests and integration tests for [DHIS2-9276](https://jira.dhis2.org/browse/DHIS2-9276) and the old event working lists.

### DIFF
--- a/cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
+++ b/cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
@@ -1,0 +1,307 @@
+beforeEach(() => {
+    cy.loginThroughForm();
+});
+
+Given('you open the main page with Ngelehun and malaria case context', () => {
+    cy.server();
+    cy.route('GET', '**/events**').as('getDefaultEvents');
+
+    cy.visit('#/programId=VBqh0ynB2wv&orgUnitId=DiszpKrYNg8');
+});
+
+Then('events should be retrieved from the api using the default query args', () => {
+    cy.wait('@getDefaultEvents', { timeout: 40000 }).as('result');
+
+    cy.get('@result')
+        .its('status')
+        .should('equal', 200);
+
+    cy.get('@result')
+        .its('url')
+        .should('include', 'pageSize=15');
+
+    cy.get('@result')
+        .its('url')
+        .should('include', 'page=1');
+
+    cy.get('@result').its('response.body.events').as('events');
+});
+
+Then('the list should display the events retrieved from the api', () => {
+    cy.get('@events')
+        .then((events) => {
+            cy.get('[data-test="event-working-lists"]')
+                .find('tr')
+                .should('have.length', events.length + 1);
+        });
+
+    cy.get('@events')
+        .then((teis) => {
+            cy.get('[data-test="event-working-lists"]')
+                .find('tr')
+                .each(($teiRow, index) => {
+                    const rowId = $teiRow.get(0).getAttribute('id');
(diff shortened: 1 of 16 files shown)


cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
```diff
@@ -0,0 +1,307 @@
...
+                .find('tr')
+                .should('have.length', events.length + 1);
+        });
+
+    cy.get('@events')
+        .then((teis) => {
+            cy.get('[data-test="event-working-lists"]')
+                .find('tr')
+                .each(($teiRow, index) => {
+                    const rowId = $teiRow.get(0).getAttribute('id');
```
paschalidi, line 43: curious if we could use `data-test` here. 
Perhaps something like
```suggestion
                    const rowId = $teiRow.get(0).getAttribute('data-test');
```

cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
```diff
@@ -0,0 +1,307 @@
...
+
+When('you click the first page button', () => {
+    cy.route('GET', '**/events**').as('getEvents');
+
+    cy.get('[data-test="dhis2-capture-search-pagination-first-page"]')
+        .click();
+});
+
+When('you change rows per page to 50', () => {
+    cy.get('div[id="rows-per-page-selector"]')
```
paschalidi, line 245: I see you chose to use an id  here. Isnt there a way to go with the `data-test` attribute?

src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
```diff
@@ -262,6 +262,7 @@ class Index extends React.Component<Props> {
                             return (
                                 <Row
                                     key={row[rowIdKey]}
+                                    id={row[rowIdKey]}
```
paschalidi, line 265: I guess this is related to my other comment

paschalidi (approved):
Good work on moving them a step toward a more user oriented testing. Looks very good. Made one comment that you might wanna look into.

src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
JoakimSM: Well, I think my preferred data-test for this would be something like 'online-list-row-[rownr]'. We discussed this some time ago and I think we agreed that this more generic approach would be the better.

In some of the integration tests I'm checking the actual id of the event/tei is the same as what is retrieved from the api. So I don't feel data-test would be the right place to put the eventId/teiId.

cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
JoakimSM: https://github.com/dhis2/capture-app/pull/1442#discussion_r560803928

src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
paschalidi: Cool thanks for explaining the logic behind it. :)

cypress/integration/EventWorkingLists/EventWorkingListsDev/index.js
JoakimSM: I was working with what I already had here. But I agree data-test would be better here. Adding a data-test attribute to this one 👍

PR #1446, opened 2021-01-21
chore: stores reducers alphabetically

paschalidi:
I was having the feeling there was some effort in sorting the reducers alphabetically, so I completed it. 
Since we have quite a few reducers is nice to know where to find which, I think.
_PS: No functional changes_

Let me know your thoughts please :)


JoakimSM (approved):
👍

PR #1447, opened 2021-01-21
refactor: TECH-523 [part 01] bypass url sync | prepares by renaming certain actions 

paschalidi:
👋 hey @JoakimSM this is the ticket https://jira.dhis2.org/browse/TECH-523

### What is this PR about?
Only renames the action creators related to the refactoring that is coming. While doing the refactoring work, I must admit I found it quite difficult to work with certain namings. Therefore I took the chance to rename only those actions creators that have anything to do with the future work of this chain. 

There should not be any functional changes on this one.


src/core_modules/capture-core/components/LockedSelector/LockedSelector.actions.js
```diff
@@ -2,30 +2,34 @@
...
-    SELECTIONS_FROM_URL_VALID: 'SelectionsFromUrlValid',
-    SELECTIONS_FROM_URL_INVALID: 'SelectionsFromUrlInvalid',
-
-    BASED_ON_URL_ORG_UNIT_SET: 'BasedOnUrlOrgUnitSet',
-    BASED_ON_URL_ORG_UNIT_ERROR_RETRIEVING: 'BasedOnUrlOrgUnitErrorRetrieving',
-    BASED_ON_URL_ORG_UNIT_EMPTY_SET: 'BasedOnUrlOrgUnitEmptySet',
-
-    NEW_REGISTRATION_PAGE_OPEN: 'NewRegistrationPageOpen',
-    SEARCH_PAGE_OPEN: 'SearchPageOpen',
+    ORG_UNIT_ID_SET: 'LockedSelector.OrgUnitSet',
```
paschalidi, line 5: Hope you dont disslike my change here. It actually helped me a lot while working with complex redux sequences to have an idea where the actions come from. 

Hope this wont be a no go for you

JoakimSM: No strong opinion here. I've never used the period to namespace the action type before, but lets try it 👍

src/core_modules/capture-core/components/LockedSelector/LockedSelector.actions.js
```diff
@@ -43,9 +47,10 @@ export const openSearchPageFromLockedSelector = () => actionCreator(lockedSelect
 
 // these actions are being triggered only when the user updates the url from the url bar.
 // this way we keep our stored data in sync with the page the user is.
-export const updateSelectionsFromUrl = (data: Object) => actionCreator(lockedSelectorActionTypes.SELECTIONS_FROM_URL_UPDATE)(data);
-export const validSelectionsFromUrl = () => actionCreator(lockedSelectorActionTypes.SELECTIONS_FROM_URL_VALID)();
-export const invalidSelectionsFromUrl = (error: string) => actionCreator(lockedSelectorActionTypes.SELECTIONS_FROM_URL_INVALID)({ error });
-export const setCurrentOrgUnitBasedOnUrl = (orgUnit: Object) => actionCreator(lockedSelectorActionTypes.BASED_ON_URL_ORG_UNIT_SET)(orgUnit);
-export const errorRetrievingOrgUnitBasedOnUrl = (error: string) => actionCreator(lockedSelectorActionTypes.BASED_ON_URL_ORG_UNIT_ERROR_RETRIEVING)({ error });
-export const setEmptyOrgUnitBasedOnUrl = () => actionCreator(lockedSelectorActionTypes.BASED_ON_URL_ORG_UNIT_EMPTY_SET)();
+export const updateSelectionsFromUrl = (data: Object) => actionCreator(lockedSelectorActionTypes.CURRENT_SELECTIONS_UPDATE)(data);
```
JoakimSM, line 50: I don't know what is coming up here later, but if we're only fixing the enrollment page at this point like we discussed, I think it makes sense to keep the url part of the action type (key and value). The point here is that these actions should only be triggered when the user manipulates the url directly for the url bar.

paschalidi: Right, that makes sense. Will take care of it.

PR #1448, opened 2021-01-21
refactor: TECH-523 [part 02] bypass url sync | refactors LockedSelector container into hook

paschalidi:
https://jira.dhis2.org/browse/TECH-523

### What is this PR about?

This only refactors the LockedSelector into a hook. We do that here, so that in the future we can dispatch the actions from this component. The actions that we will need to dispatch have to do with fetching org units. For example on the enrollment page when we will be bypassing the url sync we will have the need to fetch the org units.


src/core_modules/capture-core/components/LockedSelector/LockedSelector.container.js
```diff
@@ -17,86 +16,173 @@ import {
...
+
+      const dispatchOnOpenNewEventPage = useCallback(
+          () => {
+              dispatch(openNewRegistrationPageFromLockedSelector());
+          },
+          [dispatch]);
+
+      const dispatchOnOpenNewRegistrationPageWithoutProgramId = useCallback(
+          () => {
+              // todo this is problematic here
```
paschalidi, line 99: Ignore those todos for now please

JoakimSM (approved):
LGTM 👍

PR #1449, opened 2021-01-21
refactor: TECH-523 [part 03] bypass url sync | splits the set and reset actions into four

paschalidi:
https://jira.dhis2.org/browse/TECH-523

### What is this PR about? 

Splits the locked selector set and reset actions into separate actions. This is because in the ....


src/core_modules/capture-core/components/LockedSelector/LockedSelector.epics.js
```diff
@@ -23,22 +23,63 @@ const exactUrl = (page: string, url: string) => {
...
         }));
 
+export const resetProgramIdEpic = (action$: InputObservable, store: ReduxStore) =>
+    action$.pipe(
+        ofType(lockedSelectorBatchActionTypes.PROGRAM_ID_RESET_BATCH),
+        map(() => {
+            const { app: { page } } = store.value;
+            const { programId, ...restOfQueries } = deriveUrlQueries(store.value);
+
+            // todo this is problematic here because when you are on the enrollment page you want to reset the page using the `pathname`
```
paschalidi, line 76: this will be fixed when I am working on the enrollment page

src/core_modules/capture-core/components/LockedSelector/LockedSelector.actions.js
```diff
@@ -32,8 +32,8 @@ export const lockedSelectorBatchActionTypes = {
     ORG_UNIT_ID_RESET_BATCH: 'LockedSelector.BatchOrgUnitIdReset',
 };
 
-export const setOrgUnitFromLockedSelector = (id: string, orgUnit: Object) => actionCreator(lockedSelectorActionTypes.ORG_UNIT_ID_SET)({ id, orgUnit });
-export const setProgramIdFromLockedSelector = (id: string) => actionCreator(lockedSelectorActionTypes.PROGRAM_ID_SET)(id);
+export const setOrgUnitFromLockedSelector = (id: string, orgUnit: Object) => actionCreator(lockedSelectorActionTypes.ORG_UNIT_ID_SET)({ orgUnitId: id, orgUnit });
```
JoakimSM, line 35: NIT: maybe rename `setOrgUnitFromLockedSelector` input param instead?

JoakimSM (approved):

src/core_modules/capture-core/components/LockedSelector/LockedSelector.actions.js
paschalidi: you mean to change the `id: string` on line 35 to be `orgUnitId` already?

JoakimSM: yep, doesn't make sense to you? (very nitty I know)

paschalidi: Yep makes sense surely. Was only making sure I understood well

PR #1450, opened 2021-01-21
refactor: TECH-523 [part 04] bypass url sync | the urls are following the react-router convention

paschalidi:
Hey 👋 

### What is this PR about?

Basically what we are doing here is refactoring the urls so that they are using the react router convention.

I do that because it actually is simpler to go down this way. Simpler as the code is simpler and easier to reason about. 
Therefore urls which used to be `search/programId=123&orgUnitId=123` are now according to react-router "standand"/convention `search?programId=123&orgUnitId=123`.

There isnt any functional changes apart from that. I went down this road because otherwise it was much more complex to manipulate and handle the urls..


JoakimSM (approved):
Looks solid to me. Good work 👍

PR #1451, opened 2021-01-21
refactor: TECH-523 [part 05] bypass url sync | bypasses the urls that are not included in the list

paschalidi:
Lastly this adds the logic for bypassing the pages that are not included in the `pagesKeys` object. Therefore new pages will be responsible for fetching information they might need independently.

Also this ticket takes care of the ticket https://jira.dhis2.org/browse/TECH-526, this way the enrollment page will be ready to be worked after this PR, is merged on `master`

### DIFF
--- a/src/core_modules/capture-core/components/LockedSelector/LockedSelector.epics.js
+++ b/src/core_modules/capture-core/components/LockedSelector/LockedSelector.epics.js
@@ -15,7 +15,8 @@ import {
 } from './LockedSelector.actions';
 import { programCollection } from '../../metaDataMemoryStores';
 import { getApi } from '../../d2';
-import { pageKeys, pagesWithRouter, urlArguments } from '../../utils/url';
+import { deriveUrlQueries, pagesWithRouter, urlArguments } from '../../utils/url';
+import { pageKeys } from '../App/withAppUrlSync';
 
 const exactUrl = (page: string, url: string) => {
     if (page && page !== 'viewEvent') {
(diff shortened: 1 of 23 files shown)


src/core_modules/capture-core/components/LockedSelector/LockedSelector.epics.js
```diff
@@ -113,3 +113,15 @@ export const validateSelectionsBasedOnUrlUpdateEpic = (action$: InputObservable,
 
             return validSelectionsFromUrl();
         }));
+
+export const fetchOrgUnitEpic = (action$: InputObservable) =>
+    action$.pipe(
+        ofType(lockedSelectorActionTypes.FETCH_ORG_UNIT),
+        switchMap(({ payload: { orgUnitId } }) =>
+            from(fetchOrgUnits(orgUnitId))
```
JoakimSM, line 121: Should have been a comment on a previous PR, but I suggest we try using the app runtime data engine when we're quyering the api (replacing getApi). The epic signature would look something like [this](https://github.com/dhis2/capture-app/blob/0e6d99f8b3112e6a21699e50be4b82127d20b50a/src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/epics/teiViewEpics/teiView.epics.js#L9 ). From there you can use `querySingleResource`. At some point we can remove getApi altogether (and the entire old d2).

JoakimSM (approved):
Looks good @paschalidi. One suggestion you might have time to look into.

src/core_modules/capture-core/components/LockedSelector/LockedSelector.epics.js
paschalidi: Okay didnt realise that we dont wanna be using that anymore. Thanks!

PR #1471, opened 2021-01-29
fix: DHIS2-10161 [part 07] Create enrollment page with routing | adds the tet name on the locked selector

paschalidi:
https://jira.dhis2.org/browse/DHIS2-10370

### What is this PR about?
We are having better wording on the locked selector. Replacing the "Tracked entrity type" with the actual tet name. In the case bellow is Person. Could be anything tho :) 
image

### DIFF
--- a/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
+++ b/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
@@ -30,7 +30,6 @@ const deriveSelectedName = (attributes = {}) => attributes.reduce((acc, { value:
 const fetchTeiStream = teiId => from(fetchTrackedEntityInstance(teiId))
     .pipe(
         map(({ attributes, enrollments, trackedEntityType }) => {
-            // todo this is not scaling when you have many attributes the name will be huge
             const teiDisplayName = deriveSelectedName(attributes);
             const enrollmentsSortedByDate = sortByDate(enrollments);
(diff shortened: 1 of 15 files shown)


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -28,13 +29,14 @@ const deriveSelectedName = (attributes = {}) => attributes.reduce((acc, { value:
 
 const fetchTeiStream = teiId => from(fetchTrackedEntityInstance(teiId))
     .pipe(
-        map(({ attributes, enrollments }) => {
-            // todo this is not scaling when you have many attributes the name will be huge
-            const selectedName = deriveSelectedName(attributes);
+        map(({ attributes, enrollments, trackedEntityType }) => {
+            const teiDisplayName = deriveSelectedName(attributes);
             const enrollmentsSortedByDate = sortByDate(enrollments);
+            const { name: tetDisplayName } = getTrackedEntityTypeThrowIfNotFound(trackedEntityType);
```
JoakimSM, line 35: I would prefer if we derived this from the program metadata when rendering the page instead of saving it to the redux store. What do you think?

paschalidi: Yep lets tackle this here. You ve made another comment about it. But I think it is best if I deal with it here since its much higher up the chain. ⛓️

PR #1483, opened 2021-02-01
refactor: exports shared steps & fixes all tests

paschalidi:
exports common steps onto the `integration/sharedSteps.js` file. 

Thought is time to clean this up a little and keep the steps DRY


JoakimSM (commented):
I'm concerned we're about to make a sharedSteps monster file here. For example: 
```
When('you change rows per page to 10', () => {
     cy.get('div[data-test="rows-per-page-selector"]')
         .click()
         .contains('10')
         .click();
 });
```

is only applicable to the working lists tests and could be in shared working lists file.

I would prefer if we differentiate a bit here.

paschalidi:
Good point good point. Will keep those steps close to their tests definitions then!

JoakimSM (approved):
LGTM

PR #1490, opened 2021-02-05
refactor : TECH-524 cleanup data entry widget output

paschalidi:
@JoakimSM here is the one about the `DataEntryWidgetOutput` previously called `GeneralOutput`.

### What is happening int this PR?

I am trying to put together a component called `DataEntryWidgetOutput` which will be responsible for displaying the widgets we have now. 

At the moment the ViewEvent page has its own and then the other pages have their own "GeneralOutput". Mostly I move things around and twick the props so that everything works as before. 

I would appreciate if you can think of any edge cases I couldnt think of!


JoakimSM:
We will most likely use new widgets here and structure this differently. Converting this one to draft for now.

simonadomnisoru:
Clean old PRs.

PR #1496, opened 2021-02-09
feat: DHIS2-10321 adds View Dashboard button on duplicates modal

paschalidi:
https://jira.dhis2.org/browse/DHIS2-10321

I also did some refactoring on the code that I touched. I tried to give a generic option to the user in case they wanted to pass any action button on the `CardListItem` component. Before it felt it was mostly done with `onLink` in mind. 

I think you might agree that we now have more usecases and therefore adding "generalising" the property that passes this button on the bottom of the list makes some sense.


src/core_modules/capture-core/components/CardList/CardList.types.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
paschalidi, line 1: This whole file used to be in the search page. I moved it here since it feels like is a better place to have it. Open to suggestions tho

src/core_modules/capture-core/components/CardList/CardListItem.component.js
```diff
@@ -227,15 +221,23 @@ const CardListItemIndex = ({
...
+                &&
+                <div className={classes.buttonMargin}>
+                    {
+                        renderCustomCardActions({
+                            item,
+                            // we pass the programName because we have the case that the scope of the search
+                            // can be different that the scopeId from the url
+                            // this can happen for example when you are registering through the relationships
+                            programName: currentSearchScopeName,
+                            enrollmentType,
```
paschalidi, line 234: In my try to make this more generic I moved away from the `navigationButtonState` which was very search page oriented. Now I only pass the enrollment type and then the buttons on the search page will decide what is to be rendered.

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
```diff
@@ -54,6 +56,22 @@ const DialogButtons = ({ onCancel, onSave }) => (<>
     }
 </>);
 
+const CardListButton = (({ teiId, orgUnitId }) => {
+    const scopeHierarchy = useSelector(({ router: { location: { query } } }) => (query.programId ? 'PROGRAM' : 'TRACKED_ENTITY_TYPE'));
```
paschalidi, line 60: Not huge fun of the name here. feel free to point out a better one :)

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
```diff
@@ -207,6 +235,9 @@ const RegistrationDataEntryPlain = (
                                         selectedScopeId={selectedScopeId}
                                         open={modalIsOpen}
                                         onCancel={closeModal}
+                                        renderCardActions={({ item }) =>
+                                            <CardListButton teiId={item.id} orgUnitId={item.tei.orgUnit} item={item} />
+                                        }
```
paschalidi, line 240: It might look odd at the beggining that we are passing the _exact_ same property in the `PossibleDuplicatesDialog` and the `DataEntryWidgetOutput`. 

I wanted to remind you though that the `DataEntryWidgetOutput` has its own modal, which pops up when you click the warning indicator from the right side.

src/core_modules/capture-core/components/Pages/Search/SearchResults/SearchResults.component.js
```diff
@@ -48,13 +45,28 @@ const CardListButtons = withStyles(buttonStyles)(
...
+              case enrollmentTypes.CANCELLED:
+              case enrollmentTypes.COMPLETED:
+                  return availableCardListButtonState.SHOW_RE_ENROLLMENT_BUTTON;
+              default:
+                  return availableCardListButtonState.DONT_SHOW_BUTTON;
+              }
+          };
+
+        const navigationButtonsState = deriveNavigationButtonState(enrollmentType);
+
```
paschalidi, line 66: As I was explaining before now the buttons component are responsible of what buttons is to be shown

src/core_modules/capture-core/components/CardList/CardList.component.js
```diff
@@ -1,18 +1,19 @@
 // @flow
 import React from 'react';
-import type { ComponentType, Element } from 'react';
+import type { ComponentType } from 'react';
 import { withStyles } from '@material-ui/core';
 import { CardListItem } from './CardListItem.component';
 import { makeElementsContainerSelector } from './CardList.selectors';
-import type { CardDataElementsInformation, SearchResultItem } from '../Pages/Search/SearchResults/SearchResults.types';
+import type { CardDataElementsInformation } from '../Pages/Search/SearchResults/SearchResults.types';
```
JoakimSM, line 7: I think it would be better to have this type defined in the CardList "module" itself. Assuming my context here is correct and the CardList is a more generic component (currently used by both the SearchPage, RegistrationPage, Relationship part of the app)

EDIT: Probably out of scope since I can see you only removed `SearchResultItem` part. And also this is a reoccurring problem throughout this "module".

JoakimSM (approved):
LGTM

src/core_modules/capture-core/components/CardList/CardList.component.js
paschalidi: Perhaps we create a TECH ticket to clean this up as we did with the GeneralOutput Component?

PR #1498, opened 2021-02-10
feat: DHIS2-10161 [part 05] Create enrollment page with routing | demo improvements

paschalidi:
Hey 👋 

This adds the suggestions from Joe on wording and also I am adding the case where the selected program accepts "building" tracked entity type, not "person".

### DIFF
--- a/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.js
+++ b/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.js
@@ -38,15 +38,12 @@ export const showMissingMessageViewOnEnrollmentPage = () =>
 export const showErrorViewOnEnrollmentPage = ({ error }: { error: string }) =>
     actionCreator(enrollmentPageActionTypes.INFORMATION_ERROR_FETCH)({ error });
 
-export const successfulFetchingEnrollmentPageInformationFromUrl = ({ teiDisplayName, enrollmentsSortedByDate }: Object) =>
+export const successfulFetchingEnrollmentPageInformationFromUrl = ({ teiDisplayName, tetId, enrollmentsSortedByDate }: Object) =>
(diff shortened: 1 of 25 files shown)


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.js
```diff
@@ -38,15 +38,12 @@ export const showMissingMessageViewOnEnrollmentPage = () =>
...
+        { teiDisplayName, tetId, enrollmentsSortedByDate });
 
 export const openEnrollmentPage = ({ programId, orgUnitId, teiId, enrollmentId }: Object) =>
     actionCreator(enrollmentPageActionTypes.PAGE_OPEN)({ programId, orgUnitId, teiId, enrollmentId });
 
 export const cleanEnrollmentPage = () =>
     actionCreator(enrollmentPageActionTypes.PAGE_CLEAN)();
-
-export const resetProgramOnEnrollmentPage = () =>
-    actionCreator(enrollmentPageActionTypes.CUSTOM_PROGRAM_RESET)();
```
paschalidi, line 52: realised I wasnt using this.

src/core_modules/capture-core/components/Pages/New/NewPage.container.js
```diff
@@ -68,7 +68,7 @@ export const NewPage: ComponentType<{||}> = () => {
     // This is combo category selection. When you have selected a program but
     // the selection is incomplete we want the user to see a specific message
     const programCategorySelectionIncomplete: boolean =
-      useSelector(({ currentSelections: { programId, complete } }) => programId && !complete);
```
paschalidi, line 71: Also figured out there was a bug here because I was looking into the `currentSelections` to verify wether or not programs and org units where present. However looking at the url is for this case a better solution

JoakimSM (approved):
LGTM 👍

PR #1505, opened 2021-02-12
feat: DHIS2-10312 adds default value in enrolment date

paschalidi:
I like it when prs are close to one line! 

https://jira.dhis2.org/browse/DHIS2-10312

Basically adds default value in the enrollment date 
image


JoakimSM (approved):
👍

PR #1514, opened 2021-02-16
fix: DHIS2-10493 Preserve sharing settings

JoakimSM:
JIRA issue: [DHIS2-10493](https://jira.dhis2.org/browse/DHIS2-10493)

Fixed an issue that caused the sharing settings for a working list (eventFilters in the api) to reset when the working list is updated.

Workflow that earlier caused the issue: 1) Change the Sharing settings by clicking "Share view..." 2) Update the view 3) Refresh.

Note: I changed some api requests to use the app runtime data engine.


src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/TemplateMaintenance/SharingDialog.component.js
```diff
@@ -1,35 +1,42 @@
 // @flow
-import * as React from 'react';
+import React, { useCallback, type ComponentType } from 'react';
 import { withStyles } from '@material-ui/core/styles';
 import D2UISharingDialog from '@dhis2/d2-ui-sharing-dialog';
 import { getD2 } from '../../../../../d2';
+import type { Props } from './sharingDialog.types';
 
-const getStyles = () => ({
+const styles = {
```
paschalidi, line 8: I always thought this had to be a function. Good to know it isnt the case

src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
```diff
@@ -276,6 +276,33 @@ export const workingListsTemplatesDesc = createReducerDescription({
...
+        const initialTemplates = state[storeId].templates;
+        const templateIndex = initialTemplates.findIndex(template => template.id === templateId);
+
+        const template = {
+            ...initialTemplates[templateIndex],
+            ...sharingSettings,
+        };
+
+        const updatedTemplates = [...initialTemplates];
+        updatedTemplates[templateIndex] = template;
```
paschalidi, line 296: Looks like to me there is some business logic in this bit. Even though we dont have a hard rule on this, I think keeping the reducers slim and only for updating values  would make your solution more aligned with the rest of our reducers. 

This isnt a blocker ofc

paschalidi (approved):
Cool stuff thanks for the PR Joakim. Feel free to merge it

src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/TemplateMaintenance/SharingDialog.component.js
JoakimSM: I actually thought I picked this up from you 🙂

paschalidi: Maybe, I think I always use it as a function though. Perhaps I am forgetful

src/core_modules/capture-core/reducers/descriptions/workingLists/workingLists.reducerDescription.js
JoakimSM: Totally agree. I need to refactor how the templates are stored, should be stored as an object literal instead of an array, this would fix the problem I've run into here.

This would require some effort to fix so I would rather do this in a separate PR though.

PR #1520, opened 2021-02-22
feat: DHIS2-10525 adds url

paschalidi:
Hey both 
ticket is https://jira.dhis2.org/browse/DHIS2-10525

As we have discussed I encoded the url, let me know if you find that I missed anything.

### DIFF
--- a/src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
+++ b/src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
@@ -3,17 +3,7 @@ import { config } from 'd2';
 export const navigateToTrackedEntityDashboard = (teiId, orgUnitId, scopeSearchParam, currentUrl) => {
     const { baseUrl } = config;
     const instanceBaseUrl = baseUrl.split('/api')[0];
-
-    const returnUrl =
-      // when developing locally the following will always be false
-      window.location.href.includes(instanceBaseUrl)
-          ?
-          window.location.href.split(instanceBaseUrl)[1]
-          :
-          `/dhis-web-capture/#${currentUrl}`;
-
-    const base64Url = btoa(returnUrl);
-
+    const base64Url = btoa(`/dhis-web-capture/#${currentUrl}`);
 
     // `returnUrl` serves the purpose that the tracker capture app can sent the user back to the capture app.
     // In the future when there is not gonna be linking between the two apps this wont be needed.
--- a/src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
+++ b/src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
@@ -1,7 +1,11 @@
 import { config } from 'd2';
 
-export const navigateToTrackedEntityDashboard = (teiId, orgUnitId, scopeSearchParam) => {
+export const navigateToTrackedEntityDashboard = (teiId, orgUnitId, scopeSearchParam, currentUrl) => {
     const { baseUrl } = config;
     const instanceBaseUrl = baseUrl.split('/api')[0];
-    window.location.href = `${instanceBaseUrl}/dhis-web-tracker-capture/#/dashboard?tei=${teiId}&ou=${orgUnitId}&${scopeSearchParam}`;
+    const base64Url = btoa(`capture-app/index.html#${currentUrl}`);
+
+    // `returnUrl` serves the purpose that the tracker capture app can sent the user back to the capture app.
+    // In the future when there is not gonna be linking between the two apps this wont be needed.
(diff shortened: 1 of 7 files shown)


src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
```diff
@@ -1,7 +1,11 @@
 import { config } from 'd2';
 
-export const navigateToTrackedEntityDashboard = (teiId, orgUnitId, scopeSearchParam) => {
+export const navigateToTrackedEntityDashboard = (teiId, orgUnitId, scopeSearchParam, currentUrl) => {
     const { baseUrl } = config;
     const instanceBaseUrl = baseUrl.split('/api')[0];
-    window.location.href = `${instanceBaseUrl}/dhis-web-tracker-capture/#/dashboard?tei=${teiId}&ou=${orgUnitId}&${scopeSearchParam}`;
+    const base64Url = btoa(`/capture-app/index.html#${currentUrl}`);
```
paschalidi, line 6: The urls will be of the following format `/capture-app/index.html#${currentUrl}` which basically will be something like `/capture-app/index.html#/new?programId=ID&orgUnitId=ID` or `/capture-app/index.html#/search?trackedEntityType=IDorgUnitId=ID` and so on

JoakimSM: I have a suggestion :) Can the `returnUrl` be the whole url except the `baseUrl`? That way we would avoid any potential problems if the app name changes (or if we're uploading a custom app).

JoakimSM (commented):
One suggestion, apart from that looks good 👍

src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
paschalidi: uhuh surely can do that

JoakimSM: Alternatively we need to change `capture-app` in the url to `dhis-web-capture`. But I would prefer the above.

paschalidi: Help me understand this right, we want to omit the first bit of this `const base64Url = btoa(`/capture-app/index.html#${currentUrl}`);` and have as a result this ->`const base64Url = btoa(`/index.html#${currentUrl}`);`?

paschalidi: Problem is I dont know bu heart what value the `baseUrl` holds usually..

JoakimSM: On line 4 in the current file you have: `const { baseUrl } = config;`

The return url I would like to be the entire url in the address bar, except the baseUrl (from line 4). 

Does that make sense?

paschalidi: Almost. Can you give an example please?

paschalidi: Okay now got it. Makes sense

src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
```diff
@@ -1,7 +1,21 @@
...
     const instanceBaseUrl = baseUrl.split('/api')[0];
-    window.location.href = `${instanceBaseUrl}/dhis-web-tracker-capture/#/dashboard?tei=${teiId}&ou=${orgUnitId}&${scopeSearchParam}`;
+
+    const returnUrl =
+      // when developing locally the following will always be false
+      window.location.href.includes(instanceBaseUrl)
+          ?
+          window.location.href.split(instanceBaseUrl)[1]
+          :
+          `/dhis-web-capture/#${currentUrl}`;
```
paschalidi, line 13: I dont think this is self explanatory but when you develop locally you can match the `instanceBaseUrl` (which will be somthing between these lines`https://debug.dhis2.org/ca-2.36`) with the current `window.location.href`. 

For this cases I hardcoded the value ``/dhis-web-capture/#${currentUrl}`;`

I think there is much room for improvement here but cant find a better solution. Very open to suggestions

Bekkalizer (approved):

src/core_modules/capture-core/utils/navigateToTrackedEntityDashboard.js
JoakimSM: On second thought, it won't work as I imagined. Let's go with the easy solution.

JoakimSM (approved):

PR #1524, opened 2021-02-23
fix: DHIS2-10461 Date filters

JoakimSM:
JIRA-issue: [DHIS2-10461](https://jira.dhis2.org/browse/DHIS2-10461)

Fixed an issue that caused the date filters to not save properly when saving a working list (eventFilters in the api)


paschalidi (approved):
LGTM

PR #1526, opened 2021-02-24
feat: DHIS2-10326 open enrollment from working list

JoakimSM:
ticket: [DHIS2-10326](https://jira.dhis2.org/browse/DHIS2-10326)

Open enrollment in the old tracker capture app from the working lists

### DIFF
--- a/src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
+++ b/src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
@@ -1,6 +1,8 @@
 // @flow
 import React, { useCallback } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
+import { useLocation } from 'react-router';
+import { navigateToTrackedEntityDashboard } from '../../../../../utils/navigateToTrackedEntityDashboard';
 import { TeiWorkingListsSetup } from '../Setup';
 import { useWorkingListsCommonStateManagement, fetchTemplatesSuccess, fetchTemplates } from '../../WorkingListsCommon';
 import { useTrackerProgram } from '../../../../../hooks/useTrackerProgram';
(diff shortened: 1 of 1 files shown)


src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
```diff
@@ -12,31 +14,43 @@ export const TeiWorkingListsReduxProvider = ({ storeId }: Props) => {
     const program = useTrackerProgram(programId);
 
     // Being pragmatic here, disabling behavior we will implement later
```
paschalidi, line 16: we want to keep this comment?

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
```diff
@@ -12,31 +14,43 @@ export const TeiWorkingListsReduxProvider = ({ storeId }: Props) => {
...
-    const commonStateManagementProps: Object = useWorkingListsCommonStateManagement(storeId, TEI_WORKING_LISTS_TYPE, program);
+    const {
+        orgUnitId,
+        lastTransaction,
+        lastTransactionOnListDataRefresh,
+        listDataRefreshTimestamp,
+        onAddTemplate,
+        onUpdateTemplate,
+        onDeleteTemplate,
+        onSetTemplateSharingSettings,
```
paschalidi, line 25: One question, then plan is to have these variables here because we want to omit them from the `commonStateManagementProps`?

paschalidi (approved):
👍

src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/ReduxProvider/TeiWorkingListsReduxProvider.container.js
JoakimSM: Yeah, I don't particularly like this solution. At the moment, the variables shouldn't have been exposed from `useWorkingListsCommonStateManagement` at all (but rather moved to the event working list part). But all that aside, given we want to omit the variables, what would be your preferred solution? Should we make a custom utility function? Curious what you think.

paschalidi: Will need to dive into it a little bit more. Give me sometime to think about it :)

paschalidi: Usually my first thought would be to pass the props explicitly. Aka not having the `...passOnProps` but I understand that if there are 20 props in the `...passOnProps` this might be a little bit difficult.

JoakimSM: I agree, explicit should be the preferred option overall. But like you said, sometimes it is useful to get all except some of the props. I can think of a few solutions, but they're not as elegant.

paschalidi: Yeah I cant think of a better one that what you did here 👍

JoakimSM: Ok, let's keep it for now then

PR #1532, opened 2021-03-01
fix: program error bug

paschalidi:
Hey this fixes the https://jira.dhis2.org/browse/DHIS2-10584 and adds tests to cover it. 

Check it out and please let me know what you thinking :)


JoakimSM (approved):
👍

PR #1536, opened 2021-03-01
fix: url back and forth bug

paschalidi:
There was a bug I found related to the url and is pretty easy to explain. 

Basically when you will have the url like so 
image

And you press to go back to the previous page by the browser button sometimes the page goes into an infinite loop. This is because the url queries aka `orgUnitId` and `programId` arent in sync with the `currentSelections`.

This PR fixes this bug by updating the store with the newest `currentSelections` and omiting the ones that were there beforehand


JoakimSM (approved):
Thanks for staying on top of this @paschalidi 👍

PR #1550, opened 2021-03-10
fix: TECH-542 Switch to new endpoint for org units

JoakimSM:
[TECH-542](https://jira.dhis2.org/browse/TECH-542)

The straight up switch to the new endpoint for downloading organisation units.

Will create a new issue for fixing the memory consumption part. That part will probably be merged into master only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "capture-app",
   "homepage": ".",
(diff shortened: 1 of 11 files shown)


Bekkalizer (approved):
Looks good!

PR #1554, opened 2021-03-11
chore: uses minDayFromStart

paschalidi:
fixes bug related to the min days from start


Bekkalizer (approved):
Looks good!
Double checked in the old tracker, we are using minDaysFromStart.

PR #1562, opened 2021-03-16
DHIS2-10315 - part 02 enrollment event edit page - refactors the locked selector

paschalidi:
Hello 👋 

### What is this PR about?
@JoakimSM basically refactoring the locked selector. Now we can use the property `renderExtraSelectors` and pass extras selectors we might wanna be rendered additionally.

Not really any functional changes regarding the ticket yet. Just preparation for it.


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.component.js
```diff
@@ -22,8 +44,65 @@ const getStyles = ({ typography }) => ({
...
+                />
+            </Grid>
+        }
+    </>);
+};
+
+
 const EnrollmentPagePlain = ({ classes, enrollmentPageStatus }) => (<>
-    <LockedSelector pageToPush="enrollment" />
+    <LockedSelector pageToPush="enrollment" renderExtraSelectors={(width, extraClasses) => <Extras width={width} classes={extraClasses} />} />
```
JoakimSM, line 105: I would prefer wrapping `renderExtraSelectors` in `useCallback`. I still think it is good practice to use the same reference whenever possible.

JoakimSM (approved):
Looks good @paschalidi. Would be great if you looked into my minor comment.

simonadomnisoru:
Clean old PRs.

PR #1564, opened 2021-03-16
DHIS2-10315 - part 04 enrollment event edit page - fetches the information

paschalidi:
Hey

### What is this PR about?
Here is where all the relevant to the page information are being fetched and displayed in the locked selector.
Also logic is being added for what happens when you deselect program and orgunit. For now I cant find a better solution than this, but as we said we wanted to refactor the locked selector, so a better solution would be sort of a waste at this point.


simonadomnisoru:
Clean old PRs.

PR #1565, opened 2021-03-16
DHIS2-10315 - part 05 enrollment event edit page - cleans up and redirects

paschalidi:
For the time being this is workaround on the fact that we want to have specific actions running while we pass them on the `LockedSelector` component.

In simple words what I am doing is checking if there is a custom action passed in for reseting the org unit `customActionsOnOrgUnitIdReset`. When there is an action passed then I dont trigger the `push` from the `LockedSelector` epics but from that custom action. 

Now after our refactoring this will be removed this is why I said its a workaround.


simonadomnisoru:
Clean old PRs.

PR #1566, opened 2021-03-17
fix: the breaking api endpoint 

paschalidi:


JoakimSM (commented):
I have a request for this one: Can we use `onGetValidationContext` instead? In both `TrackedEntityInstanceDataEntry` and `EnrollmentDataEntry` you have the function `getValidationContext` where you can add the programId/tetId (already part of the props). The value you add to the return object in these functions will be sent through to for example the `DataElementFactory` and be part of the contextProps.

This way, the more generic components, like D2Form, won't have a scope passed in directly, but this would rather be part of the validationContext callback function that is needed for unique validations.

Please let me know if have any questions about this or you would like a better explanation.

paschalidi (commented):
@JoakimSM I applied your suggestion, thanks for staying on top of this! Ready for another review

src/core_modules/capture-core/components/DataEntries/Enrollment/EnrollmentDataEntry.component.js
```diff
@@ -312,8 +312,9 @@ class PreEnrollmentDataEntryPure extends React.PureComponent<Object> {
 
 export class EnrollmentDataEntryComponent extends React.Component<PreEnrollmentDataEntryProps> {
     getValidationContext = () => {
-        const { orgUnit, onGetUnsavedAttributeValues } = this.props;
+        const { orgUnit, onGetUnsavedAttributeValues, programId } = this.props;
         return {
+            scope: programId,
```
JoakimSM, line 317: same as other comment:

(I think I would prefer to use the prop name programId here. This is always a program id and I don't think there is a reason the generalise here. Let me know what you think.)

JoakimSM (commented):
Logic looks good, some minor comments that it would be great if you had a look at.

JoakimSM (approved):

PR #1572, opened 2021-03-19
fix: DHIS2-10711 duplicated calls on search page

paschalidi:
https://jira.dhis2.org/browse/DHIS2-10711

### What is this PR doing?
How this workes is that the page will be immediately rendered _and_ fetch the relevant data. But then the page would go loading again since we based the readiness of the page on the `lockedSelectorLoads`.

When I was developing this I didnt have a clear separation in my mind, for the locked selector variable called `lockedSelectorLoads`. However, it might make sense that this value is only responsible for the loading state of the locked selector and not of whole pages as I did here.

### DIFF
--- a/src/core_modules/capture-core/components/DataEntries/index.js
+++ b/src/core_modules/capture-core/components/DataEntries/index.js
@@ -1,7 +1,6 @@
 // @flow
 export {
     openDataEntryForNewEnrollmentBatchAsync,
-    sectionKeys as sectionKeysForEnrollmentDataEntry,
     runRulesOnUpdateFieldBatch,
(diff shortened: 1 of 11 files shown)


JoakimSM (commented):
good work, I can verify it's working :)

It would be great if you could also have a look at the relationship-part, same problem there it seems.

paschalidi:
This must fix the error on the relationship part. The problem there was a bit different. Some time ago on a refactoring, I made it that the data entry itself is responsible for its initialisation. Apparently I forgot to change that bit in the relationship part which resulted in initialising the form twice.

JoakimSM (approved):
Perfect!

PR #1579, opened 2021-03-23
fix: DHIS2-10130 Truncate label and added tooltip

EirikHaugstulen:
Truncated overflowing labels and adding tooltips.

https://jira.dhis2.org/browse/DHIS2-10130

### DIFF
--- a/src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
+++ b/src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
@@ -17,7 +17,6 @@ const getStyles = (theme: Theme) => ({
 
 const getStylesLabel = (theme: Theme) => ({
     container: {
-        display: 'block',
         alignItems: 'center',
         overflow: 'hidden',
         whiteSpace: 'nowrap',
--- a/src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
+++ b/src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
@@ -17,10 +17,12 @@ const getStyles = (theme: Theme) => ({
 
 const getStylesLabel = (theme: Theme) => ({
     container: {
-        alignItems: 'center',
         overflow: 'hidden',
         whiteSpace: 'nowrap',
         textOverflow: 'ellipsis',
+    },
+    truncateContainer: {
+        width: '-webkit-fill-available',
     },
     required: {
         color: theme.palette.required,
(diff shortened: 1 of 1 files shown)


src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
```diff
@@ -16,8 +17,11 @@ const getStyles = (theme: Theme) => ({
 
 const getStylesLabel = (theme: Theme) => ({
     container: {
-        display: 'flex',
+        display: 'block',
```
paschalidi, line 20: `div` elements have default display value `block`. You might wanna remove this line completely if it doesnt serve any other reason. Does it?

paschalidi (commented):
Hey very good small and focused PR! 

Easy to review :) 
You might wanna look to that one comment of mine but apart from that the rest looks great!

src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
EirikHaugstulen: Thanks! It does not serve any other reason, so i have updated with another commit.

paschalidi (commented):
Doing some manual testing 
I see the labels are not centred in my screen, or to put it in other words, they are centred related to the whole width of the element. I think your solution would be better and would accommodate more use cases if we can align the tooltip in the centre of the text.  
What you think @EirikHaugstulen?
image
image

EirikHaugstulen:
Absolutely, and that was what I originally planned. The problem is that this would look good on every label that isn't truncated. Altough, when the label is truncated the tooltip doesn't render at the middle of the truncated text, it renders in the middle of the original text as if it was the whole text. I'll try to post some photos to explain what i mean. 

image

image

src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
```diff
@@ -16,8 +17,12 @@ const getStyles = (theme: Theme) => ({
 
 const getStylesLabel = (theme: Theme) => ({
     container: {
-        display: 'flex',
-        alignItems: 'center',
```
paschalidi, line 20: I think the bit that will make this work is to have this container back, and wrap the whole `Tooltip`.

Can you please try that if you got the time Eirik?

JoakimSM:
We need a design decision to continue working on this issue (see https://jira.dhis2.org/browse/DHIS2-10130). Converting this to draft for now.

PR #1580, opened 2021-03-23
fix: DHIS2-10739 handles tei message

paschalidi:
Fixes: https://jira.dhis2.org/browse/DHIS2-10739

Now the app handles the error message for the max TEI count when you are coming from a fallback.

Basically the bug was reporting that when you where doing a fallback search and there was a max number on the TEType then you would not get a warning message but a generic error message


JoakimSM (approved):
Looks good @paschalidi 👍

PR #1581, opened 2021-03-24
fix: DHIS2-10738 pressing the back button after a fallback crashes app

paschalidi:
https://jira.dhis2.org/browse/DHIS2-10738

Fixes the bug as it is described in the ticket and covers the case with tests


JoakimSM (approved):
LGTM

PR #1582, opened 2021-03-24
fix: DHIS2-10394 ErrorOnComplete showing regardless

EirikHaugstulen:
Added logic for checking if event is marked as complete, and only showing error if selected. 

https://jira.dhis2.org/browse/DHIS2-10394


paschalidi:
@EirikHaugstulen could you take a moment and explain how I could test this in the UI?

What I have tried already is:

1. go in the maintenance app
2. create a rule for the Antenatal care visit program that will be triggered on complete, like so 
image
3. In the capture app I am registering a new event for the Antenatal care visit program, like so
image

Is that you think the right way to test this ticket?

EirikHaugstulen:
Yes, I can go through how I triggered this error on my local project:

1.  Go to maintenance -> Program
2. Create a program rule variable in the program you want (I created in the Inpatient morbidity and mortality)
image
3. Create a program rule in the same program
image
image
4. Save and clear cache on local browser
5. Go to capture and insert program and registering unit and fill in the form
6. You should now get the error on complete even if the event is not completed and will therefore not be able to save the event.
image
image
7. Update your project with the PR-branch and try to do the same thing. You should now get the error if you try to complete the event, but will go through if you uncheck complete
image

Let me know is something is unclear or if it doesn't work! ☺️

paschalidi:
@EirikHaugstulen it looks for my test case that the error is there even when I dont have the status to completed.

Strange. Any ideas why that might be?

EirikHaugstulen:
@paschalidi yes, that is correct. You should be able to save the event even if the value is wrong if you are not completing. 

I had some trouble with program rules in general since you have to completely clear the cache each time something is changed. (For me, this includes google chrome, not just clearing chaching in DHIS2)

paschalidi:
After chatting over slack we figured the above out. 

One other thing that I dont fully understand from the ticket description is what happens when the program is a tracker program and doesnt have a complete status? One example of that would be the Childprogramm. 

How do we want the on complete rule behaves in that case?

EirikHaugstulen:
I think perhaps @JoakimSM has to answer that one. 
I'm not sure.

JoakimSM:
> One other thing that I dont fully understand from the ticket description is what happens when the program is a tracker program and doesnt have a complete status? One example of that would be the Childprogramm.

@paschalidi excellent question! I think we should change the behaviour of the enrollment / tei registration form so that it is always completing (and hence displaying the applicable "ERROR_ON_COMPLETE" messages). That's something fun for your to look into @EirikHaugstulen? Let me know when you have the time and I will get you started.

JoakimSM (approved):
Looks good 👍

PR #1589, opened 2021-03-25
fix: tests actually rerun

paschalidi:
@JoakimSM this fixes the issue we had with cypress not re-running the tests. See slack [discussion](https://dhis2.slack.com/archives/G014JTQQ5UN/p1616680292001300)


JoakimSM (approved):
nice!!

PR #1592, opened 2021-03-29
fix: DHIS2-10206 displays dialog when data has entered and user is exiting the page

paschalidi:
This is a chain of https://github.com/dhis2/capture-app/pull/1598 so please review that one first Joakim.

https://jira.dhis2.org/browse/DHIS2-10206

Basically this now will work for the old pages _and_ also for the enrollment registration


karolinelien:
I tested this on my local instance @paschalidi - if you enter some data and then for example click X on the org unit, the dialogue shows up which is great, but if I then click "Yes, discard changes", and then click X on program after that, the dialogue shows up again, which I guess is not really needed, you have already exited the event and haven't added any new data to save?

paschalidi:
Very nice, thanks for checking this @karolinelien :) Will fix it and poke you again.

paschalidi:
PS: @karolinelien this is now fixed, I believe. Would be cool if you can have another look into it

paschalidi:
@karolinelien this must be fixed now :) Would be cool if you want to have another look and please let me know if you see anything odd again. Hopefully everything works as expected though!

src/core_modules/capture-core/components/D2Form/asyncHandlerHOC/actions.js
```diff
@@ -39,10 +38,6 @@ export const fieldsValidated = (
     actionCreator(actionTypes.FIELDS_VALIDATED)(
         { fieldsUI, formBuilderId, formId, validatingUids });
 
-export const cleanUpFormBuilder = (remainingUids: Array<string>, formId: string) =>
-    actionCreator(actionTypes.CLEAN_UP_FORM_BUILDER)(
-        { remainingUids, formId });
-
```
paschalidi, line 45: I removed this since I didnt find anywhere that is being used. Let me know if you think I missed anything Joakim.

src/core_modules/capture-core/reducers/descriptions/dataEntry.reducerDescription.js
```diff
@@ -266,6 +219,7 @@ export const dataEntriesFieldsUIDesc = createReducerDescription({
         dataEntryValuesUI[payload.fieldId] = { ...dataEntryValuesUI[payload.fieldId], ...payload.valueMeta, modified: true };
         return newState;
     },
+    [registrationFormActionTypes.CLEAN_UP_DATA_ENTRY]: cleanUp,
```
JoakimSM, line 222: This action type is not defined. Hence, could you reflect on whether this cleanup is really needed?

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
```diff
@@ -14,12 +14,21 @@ import getOrgUnitRootsKey from '../../components/D2Form/field/Components/OrgUnit
...
     const remainingKeys = Object.keys(state).filter(key => !key.includes(formId));
     return remainingKeys.reduce((acc, key) => ({ ...acc, [key]: state[key] }), {});
 };
 
+// cleans up data entries that _start with_ dataEntryId
+const cleanUp = (state, { payload: { dataEntryId } }) => {
+    const newState = Object.keys(state).reduce((acc, curr) =>
+        (curr.startsWith(dataEntryId) ? { ...acc, [curr]: {} } : { ...acc }), {});
+
+    return newState;
```
JoakimSM, line 29: I don't think this one is ever triggered? But if it was triggered, I think it would wipe all properties of the state object.

src/core_modules/capture-core/components/Pages/New/NewPage.component.js
```diff
@@ -60,7 +62,12 @@ const NewPagePlain = ({
     ]);
 
     return (<>
-        <LockedSelector pageToPush="new" />
+        <LockedSelector
+            pageToPush="new"
+            isUserInteractionInProgress={isUserInteractionInProgress}
+            customActionsOnProgramIdReset={[cleanUpDataEntry(NEW_TEI_DATA_ENTRY_ID)]}
+            customActionsOnOrgUnitIdReset={[cleanUpDataEntry(NEW_TEI_DATA_ENTRY_ID)]}
```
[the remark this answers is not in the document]
JoakimSM: Is this really needed? See my other comment.

src/core_modules/capture-core/components/Pages/New/NewPage.container.js
```diff
@@ -80,6 +81,10 @@ export const NewPage: ComponentType<{||}> = () => {
     };
 
     const writeAccess = useUserWriteAccess(currentScopeId);
+
+    const isUserInteractionInProgress: boolean = useSelector(
+        state => dataEntryHasChanges(state, 'singleEvent-newEvent') || dataEntryHasChanges(state, 'newPageDataEntryId-newEnrollment'));
```
JoakimSM, line 86: Couldn't this lead to a potential problem if we move between the different forms (for example: singleEvent has changes, but we're in the newEnrollment form without changes?)  Can we be more specific here on what state we're checking?

JoakimSM (commented):
Made some comments on this one. Please have a look.

I also noticed that the dialog doesn't work for tracked entity registration. We might need a new issue for that, whatever you prefer.

src/core_modules/capture-core/components/Pages/New/NewPage.component.js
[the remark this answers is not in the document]
paschalidi: Yes there is an edge case that this was needed, I can try to remember and explain it

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
paschalidi: I thought so too, but when you do that the app crashes. In some cases we arent really check if the object is `undefined`. Therefore if I erase the object completely aka make it `undefined` the app will crash..

src/core_modules/capture-core/reducers/descriptions/dataEntry.reducerDescription.js
paschalidi: Hey that was a mistake of mine. I wanted to clean the data entry when the user is leaving the `newEnrollment` page. Is should be fixed now.

src/core_modules/capture-core/components/Pages/New/NewPage.container.js
paschalidi: Hm about this, this is the reason I am trying to clean up the store from all the relevant information of a data entry when the user is leaving the new registration screen. I thought this is anyway a good practise (cleaning up the form when you leave the page).

So if the user is registering a single event and then they leave the page all the relevant data will be cleaned. Then for instance when they try to register a new person they will have a "clean" redux store in regards to forms and data entries. 

You see any wholes in my thoughts in this regard?

JoakimSM: I agree cleaning up on exit is good practice 👍 Did some testing now after your update, and the clean up happens when you deselect the org.unit or the program, but if you click `Cancel` or `Submit` in the form it doesn't. Could we do the the clean up on unmount of the most applicable component? (note: make sure the clean up doesn't trigger when we navigate to the relationship part through add relationship to an event). Might be edge cases that we need to handle.

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
JoakimSM: I mean it should be possible to use without cleanup:

Object.keys(state).reduce((acc, curr) =>
        (curr.startsWith(dataEntryId) ? { ...acc, [curr]: {} } : { ...acc, **[curr]: state[curr]** }), {});

src/core_modules/capture-core/components/Pages/New/NewPage.container.js
paschalidi: We chatted about this on a meeting. Now it should be fixed

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
paschalidi: Not sure I am following, what would then be the difference?

paschalidi: Applied the changes you are suggesting, thanks

JoakimSM (approved):
👍

PR #1598, opened 2021-03-31
chore: merges the change of the data-test naming into v36

paschalidi:
This merges the change in naming of the data-test in v36. I found it super confusing these are different in those two versions.

No functional changes are added.


JoakimSM (approved):
Looks good to me 👍 Great if you squash merge this to keep the history a bit cleaner.

PR #1599, opened 2021-04-01
fix: correct app name and title specified in d2.config.js

amcgee:
This fixes two issues in the Capture app configuration:

1. In order to successfully override the bundled Capture app, the name needs to match `capture` exactly 
2. The title displayed in the Headerbar and in the browser tab bar is `capture-app` which is incorrect, this corrects it to `Capture`.

### DIFF
--- a/d2.config.js
+++ b/d2.config.js
@@ -1,12 +1,12 @@
 const config = {
-  name: "capture",
-  title: "Capture",
-  type: "app",
-  coreApp: true,
+    name: 'capture',
+    title: 'Capture',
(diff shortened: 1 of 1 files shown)


paschalidi (approved):
Thanks @amcgee

paschalidi:
FYI @amcgee all the tests but one will fail. There is a backend bug we are detecting on our tests that needs to be fixed in order for the test case to get green.

amcgee:
@paschalidi OK, feel free to merge it whenever it works for you

PR #1600, opened 2021-04-01
fix: DHIS2-10830 capture app will crash when program is missing some orgUnits

paschalidi:
This is based on https://github.com/dhis2/capture-app/pull/1598 would be amazing if you could check that PR first @JoakimSM  🙏 

https://jira.dhis2.org/browse/DHIS2-10830

This bug will appear only when we login with the user `englishB`. @vilkg I thought this is a good opportunity to write a smoke test for that user. Let me know your thoughts please. As I said on slack I am concerned QA might be doing the same sort of testing. If this is the case we might want to find a way not to duplicate that effort :slightly_smiling_face:


vilkg (approved):
Nice!

paschalidi:
@vilkg this is a backend issue, I believe @JoakimSM knows more about it.

JoakimSM (approved):
Thanks @paschalidi!

I pushed an additional commit to ensure organisationUnits in Program are never undefined / null.

PR #1605, opened 2021-04-07
feat: added section description boxes

superskip:
TODO: change the style


src/core_modules/capture-core/components/Section/Section.component.js
```diff
@@ -60,6 +61,7 @@ class Section extends Component<Props> {
         return (
             <div>
                 {this.getHeader()}
```
superskip, line 63: this.props.header is processed in the function getHeader(). I haven't understood why yet, and thus don't know if such a function is needed for this.props.description.

src/core_modules/capture-core/components/D2Form/SectionDescriptionBox.component.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import * as React from 'react';
```
JoakimSM, line 2: NIT: I would prefer `import React from 'react'` and importing potential react type definitions individually.

src/core_modules/capture-core/components/D2Form/SectionDescriptionBox.component.js
```diff
@@ -0,0 +1,50 @@
...
+        marginRight: 5,
+        marginTop: 1,
+    },
+    description: {
+        fontSize: 13,
+        lineHeight: '17px',
+    },
+};
+
+type Props = {
```
JoakimSM, line 26: NIT: Lately we've added the type definitions in its own file. Great if this can also be done here.

src/core_modules/capture-core/components/D2Form/SectionDescriptionBox.component.js
```diff
@@ -0,0 +1,50 @@
...
+                <IconInfo16 />
+            </div>
+            <div className={classes.description}>
+                {description}
+            </div>
+        </div>
+    );
+};
+
+export const SectionDescriptionBox = withStyles(styles)(SectionDescriptionBoxPlain);
```
JoakimSM, line 50: NIT: Maybe annotate this with ComponentType. If we don't annotate this one, flow will not produce an error if we send in invalid props.

src/core_modules/capture-core/components/Section/Section.component.js
JoakimSM: This component should totally be refactored, but let's leave that for another time. What you did here makes perfect sense 👍

src/core_modules/capture-core/metaData/Program/ProgramStage.js
```diff
@@ -10,6 +10,7 @@ import type { RelationshipType } from '../RelationshipType';
 export default class ProgramStage {
     _id: string;
     _name: string;
+    _displayDescription: string;
```
JoakimSM, line 13: I don't think this is needed?

JoakimSM (commented):
Thanks @superskip, looks good 👍 Couple of nit-picking comments that we can talk about in our sync meeting.

In addition, the demo database is now updated so you can configure your tests to reflect the changes made there.

cypress/integration/NewPage.feature
```diff
@@ -36,6 +36,12 @@ Feature: User creates a new entries from the registration page
     And you select the Malaria case registration program
     Then you see the registration form for the Malaria case registration
 
+  Scenario: Viewing section descriptions with organisation unit and event program Inpatient morbidity and mortality selected
+    Given you are on the default registration page
+    And you select org unit
+    And you select the Inpatient morbidity and mortality program
```
JoakimSM, line 42: I would also suggest using `When` for these two lines 🙂

src/core_modules/capture-core/components/D2Form/SectionDescriptionBox.types.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+export type PlainProps = {|
```
JoakimSM, line 2: Actually, I would have switched the name of PlainProps and Props. As in, PlainProps are props for SectionDescriptionBoxPlain and Props are props for the exported component.

JoakimSM (commented):
One final thing to look at, but you're almost there now 🙂

JoakimSM (approved):
Good work @superskip!

PR #1608, opened 2021-04-08
feat: DHIS2-10759 Implement sections in enrollment form

EirikHaugstulen:
Added feature for showing custom sections in enrollment form. 
https://jira.dhis2.org/browse/DHIS2-10759

How to test:
1. Go to Maintenance app - Program - choose eg Child Program
2. Go to Attributes - Create Registration Form - add Sections
3. Go to Capture app, register new TEI in Child Program, you should see Sections in the Enrollment form
4. Clear cache
5. Open new event in capture


EirikHaugstulen:
Had a look at the two cypress-tests that are failing, and I think neither of them are caused by something I have written in my code. The `WorkingLists/EventWorkingLists/EventWorkingListsUser.feature` is failing because the server is responding with a 404, so perhaps a backend issue.
The `SearchPage.feature` is reproducible in the current master-branch aswell. As far as I can see, it is caused by the test is looking for `'[data-test="dhis2-capture-back-button"]'` while the button is just called `'[data-test="back-button"]'`
Let me know if this is something I should look more into. @JoakimSM

JoakimSM (commented):
Looks good 👍 One minor comment for you.

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/factory/enrollment/EnrollmentFactory.js
```diff
@@ -121,15 +133,34 @@ class EnrollmentFactory {
...
+        if (cachedProgramSections && cachedProgramSections.length > 0) {
+            // $FlowFixMe
+            cachedProgramSections.asyncForEach(async (ps) => {
+                let arr;
+                if (cachedProgramTrackedEntityAttributes &&
+                    cachedProgramTrackedEntityAttributes.length > 0) {
+                    arr = cachedProgramTrackedEntityAttributes
+                        .filter(e => ps.trackedEntityAttributes
+                            .includes(e.trackedEntityAttributeId));
+                }
```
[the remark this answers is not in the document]
EirikHaugstulen: Thanks 👍

src/core_modules/capture-core/storageControllers/cache.types.js
```diff
@@ -153,6 +153,12 @@ export type CachedTrackedEntityTypeTranslation = {
     locale: string,
     value: string,
 };
+export type CachedProgramSection = {
+    id: string,
+    displayFormName: string,
+    sortOrder: number,
+    trackedEntityAttributes: Array<string>
```
huocha, line 160: Is there any reason why this `trackedEntityAttributes` is array of string while `trackedEntityAttributes` in  `apiProgramSections ` is array of object. I see these 2 type are quite identical ?

huocha (commented):
LGTM 👍  I just have a small question

simonadomnisoru (approved):
LGTM! Great job

src/core_modules/capture-core/storageControllers/cache.types.js
EirikHaugstulen: Good question! This is because the api returns objects of each trackedEntityAttribute, but this is already stored in the cache when the program itself is fetched. So to prevent double storage, I filter just the ids of each attribute and then match them with the ids of the attributes stored in the cache when the enrollment-form is built. This filtering is done in the `convertProgramSections` in `storePrograms.js` 👍 Let me know if anything is unclear!

EirikHaugstulen: And this is the matching of the ids ☺️
image

huocha: yep! thanks for the explanation

huocha (approved):

JoakimSM (approved):
Looks good 👍

PR #1638, opened 2021-04-21
feat: DHIS2-10218 part 2 - Icon refactor

JoakimSM:
Second part of [DHIS2-10218](https://jira.dhis2.org/browse/DHIS2-10218)

Adds an icon to the stage overview. Refactors the icon logic quite a bit for the dhis2 icons not bundled in the app.


src/core_modules/capture-core/components/D2Form/field/Components/OptionSetField/withOptionsIconElement.js
```diff
@@ -50,11 +51,13 @@ export default () =>
...
-                                className={classes.icon}
-                                src={icon.data}
-                                alt={label}
+                            <NonBundledDhis2Icon
+                                name={icon.name}
+                                color={icon.color}
+                                width={22}
+                                height={22}
+                                cornerRadius={2}
+                                alternativeText={label || undefined}
```
simonadomnisoru, line 60: Hey, I am wondering if `undefined` is the best value to pass to this prop. How about passing `''`?

paschalidi (approved):

src/core_modules/capture-core/components/D2Form/field/Components/OptionSetField/withOptionsIconElement.js
JoakimSM: I believe what I was thinking was that if we pass undefined we could potentially utilize any default values in the inner components. I'm not sure what that default value would be though. What do you think?

simonadomnisoru: Great idea! I think that the icon name is a good candidate to be used as the default value for the `alternativeText`. In `NonBundledDhis2Icon.component.js` I have written a deconstructor with a default value for alternativeText. 
WDYT?

src/core_modules/capture-core/components/NonBundledDhis2Icon/NonBundledDhis2Icon.component.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import React from 'react';
+import { useConfig } from '@dhis2/app-runtime';
+import { buildUrl } from 'capture-core-utils';
+import { NonBundledIcon } from 'capture-ui';
+import type { Props } from './nonBundledDhis2Icon.types';
+
+export const NonBundledDhis2Icon = ({ name, ...passOnProps }: Props) => {
```
simonadomnisoru, line 8:
```
export const NonBundledDhis2Icon = ({ name, alternativeText = name, ...passOnProps }: Props) => {
    const { baseUrl, apiVersion } = useConfig();
    const source = name && buildUrl(baseUrl, `api/${apiVersion}/icons/${name}/icon.svg`);

    return (
        <NonBundledIcon
            {...passOnProps}
            source={source}
            alternativeText={alternativeText}
        />
    );
};
```

src/core_modules/capture-core/components/D2Form/field/Components/OptionSetField/withOptionsIconElement.js
JoakimSM: Perfect! Makes sense. Suggestion applied.

src/core_modules/capture-core/components/NonBundledDhis2Icon/NonBundledDhis2Icon.component.js
JoakimSM: Perfect! Suggestion applied.

simonadomnisoru (approved):

src/core_modules/capture-core/components/FormFields/New/HOC/withLabel.js
```diff
@@ -103,16 +97,18 @@ export default (hocParams?: ?HOCParams) => (InnerComponent: React.ComponentType<
...
         return (
-            <img
-                className={iconClass}
-                src={icon.data}
-                style={{ backgroundColor: icon.color }}
-                alt={`Icon for ${label || ''}`}
+            <NonBundledDhis2Icon
+                name={icon.name}
+                color={icon.color}
+                alternativeText={`Icon for ${label || ''}`}
```
huocha, line 108: Will it need a i18n text here?

JoakimSM: Good catch @huocha! Fixing.

PR #1641, opened 2021-04-23
feat: [DHIS2-10941] Enrollment widget

simonadomnisoru:
[DHIS2-10941]( https://jira.dhis2.org/browse/DHIS2-10941) Added an empty version of the enrollment widget.


JoakimSM (approved):
nice, looks good 👍

PR #1644, opened 2021-04-23
feat: [DHIS2-10947] Profile widget

huocha:


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
```diff
@@ -40,7 +41,7 @@ export const EnrollmentPageDefaultPlain = ({ program, classes }: PlainProps) =>
                 />
             </div>
             <div className={classes.rightColumn}>
-                [placeholder profile widget]
+                <ProfileWidget attributes={attributes} />
```
simonadomnisoru, line 44: Hey, 
We discussed that we want to make the profile widget component self-contained to be able to use it in other apps. Does this mean it must pull its own content and not receive it from the pattent? 
Thanks!

huocha: that makes sense! I will change that

src/core_modules/capture-core/components/Widgets/ProfileWidget/ProfileWidget.component.js
```diff
@@ -0,0 +1,61 @@
...
+            <div>{item.value}</div>
+        </div>
+    );
+    return (
+        <div
+            data-test="profile-widget"
+        >
+            <Widget
+                header={i18n.t('Person Profile')}
+                onOpen={useCallback(() => setOpenStatus(true), [setOpenStatus])}
```
paschalidi, line 46: 🤩 good thinking!

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
```diff
@@ -40,7 +41,7 @@ export const EnrollmentPageDefaultPlain = ({ program, classes }: PlainProps) =>
                 />
             </div>
             <div className={classes.rightColumn}>
-                [placeholder profile widget]
+                <ProfileWidget />
```
paschalidi, line 44: There was this thread in the PR where we introduced the Widgets folder where we discussed that we wanted to have each widget to live in its own inside the `/components` folder directly. Simona, suggested we go with the naming `Widget.....`. 

I am wondering if we want to start applying this from this widget onwards, what you think?

src/core_modules/capture-core/components/Widgets/ProfileWidget/profileWidget.types.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+export type Props = {|
+    classes: {
+        flatListWrapper: string;
+        itemRow: string;
+        itemKey: string;
+    }
```
paschalidi, line 8: Thats not by any means bad annotation here, but you might wanna consider using `CssClasses` type we have "globally" defined. This way you dont have to all _every single_ class in this annotation here. I used to find it a bit annoying that flow will complain because a single class was missing from the annotation.

You can find that global definition in the file `src/core_modules/capture-core/flow/typeDeclarations.js` if you wanna check it out

src/core_modules/capture-ui/FlatList/FlatList.component.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import React from 'react';
+import cx from 'classnames';
+import uuid from 'uuid/v4';
+import classes from './FlatList.module.css';
```
paschalidi, line 5: I am wondering if we could also use the `withStyles` which come from the `import { withStyles } from '@material-ui/core';`. 

Or better you might wanna help me understand why you went with this alternative in this case? :)

src/core_modules/capture-ui/FlatList/flatList.types.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import * as React from 'react';
```
paschalidi, line 3: I would only import the `type Element` in this case, but that's only a suggestion :) 
```suggestion
import { type Element } from 'react';
```

src/core_modules/capture-ui/FlatList/flatList.types.js
```diff
@@ -0,0 +1,10 @@
...
+
+export type Props = {
+    list: any[],
+    renderItem: (item: any) => React.Element<any>,
+    className?: string,
+    dataTest?: string,
+};
```
paschalidi, line 10: Can we try to be exact in this annotation over here or you reckon it is impossible?

paschalidi (commented):
Hey super nice small and cohesive PR. Really easy to review! Thanks @huocha! 

Made some comments but overall great piece of work :blush:

src/core_modules/capture-ui/FlatList/FlatList.component.js
huocha: I agree, this is removed on another branch 10948

src/core_modules/capture-ui/FlatList/FlatList.component.js
```diff
@@ -0,0 +1,16 @@
...
+import type { Props } from './flatList.types';
+
+export const FlatList = ({ list, className, dataTest, renderItem, ...passOnProps }: Props) => (
+    <div
+        {...passOnProps}
```
paschalidi, line 10: can you help me understand what those might be in our case?

src/core_modules/capture-ui/FlatList/flatList.types.js
huocha: good point! I made it `any` so that it can fit all kinds of object

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
huocha: Absolutely. I will change it

src/core_modules/capture-ui/FlatList/flatList.types.js
paschalidi: Yes that's cool I think, but I was referring more to the whole `Props` annotation. 

Maybe would be easier if I post this https://flow.org/en/docs/types/objects/#toc-exact-object-types for reference.

Basically when you annotate exactly you will get an flow error if someone is trying to pass another property in a component. As we have it now I could pass any property I might want in the `FlatList` and there wont be an error warning me I am doing something wrong. In my eyes it is quite useful to have these errors in place so that we developers can communicate with each other of the appropriate way a component is meant to be used.

simonadomnisoru:
Hey @huocha, 

Great job on this PR! 👏 
I just have one question regarding the Cypress tests. Do you plan to add them in another PR? 
Thank you!

src/core_modules/capture-core/components/Widgets/ProfileWidget/profileWidget.types.js
JoakimSM: Nothing wrong with the approach here @huocha, but as @paschalidi says we've been using the more pragmatic `CssClasses` lately because it gets cumbersome to annotate the classes. I would say this one is up to you, nothing wrong with keeping this as is.

JoakimSM (approved):
Good work @huocha! It looks like everything I wanted to comment on is fixed In the next PR or is better discussed in the next PR. Lets get this in 👍

JoakimSM:
Let's squash and merge this to keep the master branch cleaner!

PR #1649, opened 2021-04-26
feat: adds PR previews

paschalidi:


JoakimSM:
Great if you could have a look at this one @varl, wasn't really in the loop. After adding the non secure `http://*.netlify.app` to the CORS whitelist this seems to work (but the idea might have been to run this securely).

varl (approved):
LGTM

PR #1652, opened 2021-04-27
chore: update flow type definitions for dependencies

JoakimSM:
Downloaded updated definitions from flow-typed (through running the command `yarn run flow:addTypes`). Many of the dependencies are missing flow definitions so we will just get something like `declare module '@dhis2/cypress-commands' {
  declare module.exports: any;
}` in the file, meaning it accepts importing anything basically. 

We have custom definitions for the app-runtime; I've updated this one with a definition for `useDataQuery`. 

Also fixed a couple of minor annotation bugs that surfaced after the update.

EDIT: I presume we need [1649](https://github.com/dhis2/capture-app/pull/1649) merged before we get a green checkmark here, but I think this PR should be reviewable anyway.


huocha (approved):

simonadomnisoru (approved):
LGTM

PR #1642, opened 2021-04-23
fix(translations): [DHIS2-8998] missing translation

simonadomnisoru:
Backporting to 36 from #1617, #1640 and https://github.com/dhis2/capture-app/pull/1655

cherry picked from commit 70b103e
cherry picked from commit b92350c
cherry picked from commit 475b27d142c52672efe9a7b399fb92342bcc1837

PR #1655, opened 2021-04-29
fix: [DHIS-8998] unsplitting translatable strings

simonadomnisoru:
Address the comments from the PR to backport to `v36` https://github.com/dhis2/capture-app/pull/1642


src/core_modules/capture-core/hooks/useScopeTitleText.js
```diff
@@ -7,10 +7,15 @@ export const useScopeTitleText = (scopeId: ?string) => {
...
-        [scopeTypes.EVENT_PROGRAM]: `${programName}`,
+        [scopeTypes.EVENT_PROGRAM]: programName,
         [scopeTypes.TRACKER_PROGRAM]:
-            `${i18n.t('{{trackedEntityName}} in program', { trackedEntityName })}: ${programName}`,
-        [scopeTypes.TRACKED_ENTITY_TYPE]: `${trackedEntityName}`,
+            i18n.t('{{trackedEntityName}} in program{{escape}} {{programName}}', {
+                trackedEntityName,
+                escape: ':',
+                programName,
+                interpolation: { escapeValue: false },
```
JoakimSM, line 16: I'm curious: why did you find it necessary to add the escapeValue here? Tested a bit and to me it seems to work without it?

JoakimSM (commented):
This would work I think. Can we also update the ones that is typed like this in the other PR: `{i18n.t('New')} {titleText}`

Sorry for stressing this, not that important of course 😊

src/core_modules/capture-core/hooks/useScopeTitleText.js
simonadomnisoru: Hey, 

I skiped the `: ` character to get rid of an error generated by the pre-commit hook.
 
Screenshot 2021-04-29 at 14 31 51

Do you get the same error on your computer?

simonadomnisoru: Another problem I found was in the Cypress tests. The `Scenario: Navigating to the enrollment page for the same tet by clicking the link button` from the `EnrollmentPageNavigation.feature ` fails because the `programName` is not correctly translated. If I use` { escapeValue: false } `the error is corrected and the text is properly displayed. 

Screenshot 2021-04-30 at 14 20 58

Let me know if this explains the reasons behind using it. 
Thank you!

JoakimSM: Interesting, you're right in the sense that I do get an error when adding the colon directly to the string. But I don't think I need the `interpolation: { escapeValue: false }` part?

simonadomnisoru: Hey and how about the Cypress failure from `Scenario: Navigating to the enrollment page for the same tet by clicking the link button` does it fail for you if you don't add `interpolation: { escapeValue: false }`? 
Thank you!

JoakimSM: Sorry, didn't refresh to see your last comment before replying. I give in, you're right, we do actually need this 🙂 Never encountered this problem before, good to know.

Thanks for explaining @simonadomnisoru 👍

JoakimSM (approved):

PR #1657, opened 2021-04-30
feat: [DHIS2-10942] Enrollment widget content

simonadomnisoru:
[DHIS2-10942]( https://jira.dhis2.org/browse/DHIS2-10942) Query and display the information in the Enrollment Widget inside the Enrollment page


src/core_modules/capture-core/components/WidgetEnrollment/Status/Status.component.js
```diff
@@ -0,0 +1,39 @@
...
+
+const styles = {
+    status: {
+        margin: spacersNum.dp4,
+    },
+};
+
+export const translateStatus = (value: string) => {
+    switch (value) {
+    case 'ACTIVE': return i18n.t('Active');
```
huocha, line 16: I think it may be better to use enum for statuses, will reduce the mis-typo. What do you think?

src/core_modules/capture-core/components/WidgetEnrollment/HOC/withTrackedEntityInstances.js
```diff
@@ -0,0 +1,47 @@
...
+            }),
+            [props.teiId, props.programId],
+        ),
+    );
+
+    if (error) {
+        throw error;
+    }
+
+    return !loading &&
```
huocha, line 32: this condition is quite complex, I would suggest separate them and using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

huocha (commented):
@simonadomnisoru  good work :tada:
I just add a few comments!
I realize we have many things in common in the enrollment page so I hope we won't get a lot of conflicts

PR #1659, opened 2021-05-04
feat: [DHIS2-10942] Enrollment widget content

simonadomnisoru:
Query and display the information in the Enrollment Widget inside the Enrollment page


src/core_modules/capture-core/components/WidgetEnrollment/HOC/withOrganizationUnit.js
```diff
@@ -0,0 +1,48 @@
...
+                    },
+                },
+            }),
+            [props.ownerOrgUnit],
+        ),
+    );
+
+    if (error) {
+        log.error(errorCreator(ENROLLMENT_RETRIEVE_ERROR)({ error }));
+        return <span>{i18n.t(ENROLLMENT_RETRIEVE_ERROR)} </span>;
```
huocha, line 34: I wonder with this approach when `withOrganizationUnit` , `withProgram` and `withEnrollment` all have error at the same time it will return 3 times the same message error? 
Should we just display it once

simonadomnisoru: Hey @huocha,

The error message will only be displayed once. When an error happens in any of the HOCs then its child `Component` is not rendered anymore. The "second error" will never appear because the next calls to the API don't even happen.

Let me know if this explanation clears your doubts. Thank you!

huocha: @simonadomnisoru thank you for your clear explanation :) that's really helpful. Great work!

huocha (approved):

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.component.js
```diff
@@ -1,25 +1,148 @@
...
+                                {convertValueClientToView(
+                                    convertValueServerToClient(
+                                        enrollment.geometry.coordinates,
+                                        dataElementTypes.COORDINATE,
+                                    ),
+                                    dataElementTypes.COORDINATE,
+                                )}
+                            </>
+                        </div>
+                    )}
```
JoakimSM, line 146: My suggestion until we add the minimap in another ticket:

- Use the location icon to the left.
- If point/coordinate: show it like you do now (coordinates the right of the icon)
- If polygon: just show the text `Polygon` (you can change the `clientToView.js` file)

Very pragmatic approach, I know, but It's only there until we get the minimap in there. Questions/comments welcome 🙂 (we can discuss details on slack)

JoakimSM (commented):
Overall this looks great @simonadomnisoru 👍

Some comments in the code that I hope you can have a look at.

Also, do you think it would be sensible to use custom hooks instead of the HOCs? I really like the separation you've done between the fetching and the actual render (container and component), but lately I've grown fond of using custom hooks instead of HOCs where this makes sense. Maybe you could keep the container-file but add custom hooks instead of the HOCs? I know the project is flooded with HOCs, but I would prefer if we moved in direction of using custom hooks where they make sense. Up for discussion of course 🙂

JoakimSM (approved):
Good work @simonadomnisoru!

Great if you can have a look into my comment before you merge though.

JoakimSM (approved):

PR #1695, opened 2021-05-17
fix: Program rules for org unit data elements [v37]

SferaDev:
Program rules using an organisation unit data element do not currently work because there's no ``converter`` method defined for them. The issue affects both to the program rule expression and the program rule action.

JIRA Issue:

- [DHIS2-11154](https://jira.dhis2.org/browse/DHIS2-11154)

Backport PRs:

- https://github.com/dhis2/capture-app/pull/1692
- https://github.com/dhis2/capture-app/pull/1693
- https://github.com/dhis2/capture-app/pull/1694

Since the related code of ``RuleEngine`` is not typed I intentionally did not type the additions. The intention of the PR is to fix the issue without rewriting the typings, that should be done in future PRs.

### DIFF
--- a/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
+++ b/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
@@ -93,7 +93,7 @@ class RulesValueConverter implements IConvertInputRulesValue {
     }
 
     convertOrgUnit(value: any): string {
-        return value?.id;
+        return value?.id || '';
     }
 }
 
--- a/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
+++ b/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
@@ -91,6 +91,10 @@ class RulesValueConverter implements IConvertInputRulesValue {
     convertAge(value: any): string {
         return this.convertDate(value);
     }
+
+    convertOrgUnit(value: any): string {
+        return value.id;
+    }
 }
 
 export default new RulesValueConverter();
--- a/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
+++ b/src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
@@ -93,7 +93,7 @@ class RulesValueConverter implements IConvertInputRulesValue {
     }
 
     convertOrgUnit(value: any): string {
-        return value.id;
+        return value?.id;
     }
 }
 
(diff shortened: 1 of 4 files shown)


SferaDev:
CC: @paschalidi @JoakimSM

SferaDev:
@simonadomnisoru

src/core_modules/capture-core/rules/engine/converters/inputValueConverter.js
```diff
@@ -91,6 +91,10 @@ class RulesValueConverter implements IConvertInputRulesValue {
     convertAge(value: any): string {
         return this.convertDate(value);
     }
+
+    convertOrgUnit(value: any): string {
```
simonadomnisoru, line 95: Hey @SferaDev,

In order to avoid en `Uncaught TypeError` I suggest you either give a default value to the `value` paramether or you use a [Optional chaning ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). WHYD?
Thanks!

SferaDev: Sure thing!

SferaDev: Actually since none of this is typed, right now it would return ``string | undefined``. Should it throw an error if undefined? (That scenario should never happen if the input types were enforced) Or it should return ``string | undefined`` even though the rest of the methods do not handle this ``undefined`` scenario?

simonadomnisoru: I would propose to follow the pattern of the rest of the methods in the class. If you look at `convertUrl` or `convertEmail`, they both return `''` for `falsey` values (`undefined` included).  

In your case it can become:
```
    convertOrgUnit(value: any): string {
        return value?.id || '';
    }
```

WDYT?

SferaDev: Sure thing, I see you're not using the Nullish coalescing operator (??) so I'll update now with your proposal. Thanks @simonadomnisoru!

SferaDev:
Any news on this @simonadomnisoru

SferaDev:
Should I rebase onto https://github.com/dhis2/capture-app/pull/1876/files?

simonadomnisoru:
> Any news on this @simonadomnisoru

Hey, 
A couple of weeks ago I mentioned this PR at a tech meeting. @JoakimSM had some concerns regarding these changes in the program rules. I'll let him explain it in more detail. 

Thank you!

JoakimSM:
Hey @SferaDev,

Really sorry about the late reply. I think we have some challenges if we are going to support this properly. We also need to think about support for this in the backend rules engine. I've made a JIRA-ticket that I hope we can follow up on in the near future: [https://jira.dhis2.org/browse/DHIS2-13058](https://jira.dhis2.org/browse/DHIS2-13058)

PR #1698, opened 2021-05-18
fix: [DHIS2-8998] delay strings initiation after the translation config loads

simonadomnisoru:


huocha (approved):

src/core_modules/capture-core/components/Pages/EditEvent/DataEntry/EditEventDataEntry.component.js
```diff
@@ -319,7 +319,7 @@ type DataEntrySection = {
     name: string,
 };
 
-const dataEntrySectionDefinitions = {
+const dataEntrySectionDefinitions = () => ({
```
JoakimSM, line 322: It seems like we are not invoking this function? (and some of the other functions in this PR).

simonadomnisoru: I added the missing invoking. Thanks for the feedback!

JoakimSM (approved):
LGTM

PR #1728, opened 2021-05-24
fix: husky lost hooks

simonadomnisoru:
After the [migration to husky v6](https://github.com/dhis2/capture-app/commit/593144402bd175b22975f095e23256a5616b88a7) the git hooks stop working. This PR adds the hooks using the [patterns from v6.](https://typicode.github.io/husky/#/?id=create-a-hook)


EirikHaugstulen (approved):
LGTM 👍

package.json
```diff
@@ -74,13 +74,9 @@
...
-    "verifyCacheVersion": "node scripts/verifyCacheVersion.js"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "d2-app-scripts i18n extract && git add ./i18n/",
-      "pre-push": "yarn linter:check && yarn flow:check"
-    }
+    "verifyCacheVersion": "node scripts/verifyCacheVersion.js",
+    "prepare": "husky install",
+    "i18n:add": "d2-app-scripts i18n extract && git add ./i18n/"
```
JoakimSM, line 79: Do you think it makes sense to add: `"postinstall": "yarn prepare"` to ensure husky is installed when doing `yarn (install)`?

simonadomnisoru: Hello,
From what I understood it will be automatically run as part of the yarn scripts lifecycle every time after your packages are installed (when you run yarn). It is what the Husky documentation recommends https://typicode.github.io/husky/#/?id=install. Therefore we don’t have to run anything manually.

JoakimSM: Thanks, did a second test and it seems to work 🤔

JoakimSM (approved):
👍

package.json
simonadomnisoru: After all, I use the `postinstall` step you suggested because it has support in both `yarn` versions. Thank you!

PR #1738, opened 2021-05-26
DHIS2-11069: Rule Engines

huocha:


src/core_modules/capture-core/components/Pages/Enrollment/hooks.js
```diff
@@ -11,3 +13,42 @@ export const useEnrollmentInfo = (enrollmentId: string, programId: string) => {
...
+        .reduce((acc, curr) => {
+            curr.programStageDataElements.forEach((stage) => {
+                acc[stage.dataElement.id] = { ...stage.dataElement };
+            });
+            return acc;
+        }, {});
+};
+
+// $FlowFixMe
+export const useRunRulesForEnrollement = (orgUnit, program, programStagesData, teiAttributes) => {
```
simonadomnisoru, line 30: Hey,
Missing types for `orgUnit, program, programStagesData, teiAttributes` parameters
Thank you!

src/core_modules/capture-core/components/Pages/Enrollment/hooks.js
```diff
@@ -11,3 +13,42 @@ export const useEnrollmentInfo = (enrollmentId: string, programId: string) => {
 
     return { programHasEnrollments, enrollmentsOnProgramContainEnrollmentId, tetId };
 };
+
+
+const extractDataElements = (data) => {
```
simonadomnisoru, line 18: Hey,
Missing types for `data` parameter
Thank you!

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
```diff
@@ -1,26 +1,60 @@
...
+
+    const { data: programStagesData, error: programStageError } = useDataQuery(programsQuery);
+    const { data: teiAttributes, error: teiError } = useDataQuery(teiAttributesQuery);
+
+    if (programStageError || teiError) {
+        log.error(errorCreator('Enrollment page could not be loaded')({ programStageError, teiError }));
+    }
+
+    const rules = useRunRulesForEnrollement(orgUnit, program, programStagesData, teiAttributes);
+    console.log({ rules });
```
simonadomnisoru, line 56: You forgot a `console.log` here. Thanks!

src/core_modules/capture-core/rules/actionsCreator/runRulesForEnrollmentPage.js
```diff
@@ -0,0 +1,43 @@
...
+    // returns an array of effects that need to take place in the UI.
+    return RulesEngine.programRuleEffectsForEnrollment(
+        { programRulesVariables, programRules, constants },
+        allEventsData,
+        orgUnit,
+        dataElementsInProgram,
+        teiValues,
+        trackedEntityAttributes,
+        enrollmentData,
+        // $FlowFixMe[prop-missing] automated comment
```
simonadomnisoru, line 40: The `optionSets: OptionSets` is set bellow in `static programRuleEffectsForEnrollment`, I wonder then why the `$FlowFixMe` comment is needed. Can you help me understand? Thanks!

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
huocha: this is intentional since Eirik will need it to do his task. thanks!

src/core_modules/capture-core/rules/actionsCreator/runRulesForEnrollmentPage.js
JoakimSM: @simonadomnisoru This is something we should add to our Flow vs Typescript comparison! The input object is a type that contains all the properties the param expects, but it also contains some additional properties. I would expect this to work when the type defined for the param is inexact (meaning more properties than those declared should be accepted). Tried to do some magic with the $Shape utility type, but also ran into problems with that approach.

I would say keep the $FlowFixMe for now, but maybe @huocha can add the label TypescriptCompare (or something similar) so that we remember to check how this is working with Typescript when the time comes.

JoakimSM (commented):
Hey @huocha, thanks for your effort here, this is quite complicated! I suggest we set up a call to discuss my comments in this one (start of next week, let's agree up on a time on slack). 

Hope you're not too discouraged by the number of comments here, we will figure it out 🙂 

I would encourage you to create some rules on your local instance and see if you get the expected output from the rules engine. I can help you get this up and running. A second step would be to create some unit tests.

JoakimSM (commented):
@huocha Please let me know if/when you have any questions regarding my comments here. This is quite complicated, so if any of my comments doesn't make 100% sense to you, it's better that we talk about them (GitHub, Slack or Meet)

I also wanted to give some input about the file structure: I would suggest that we move files that are only used in 'EnrollmentPageDefault' to that folder. For example 'runRulesForEnrollment', 'useEnrollmentsAndAttributes', 'useProgramMetadata'.

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/runRulesForEnrollment.js
```diff
@@ -0,0 +1,109 @@
...
+            program,
+            orgUnit,
+            trackedEntityAttributes,
+            dataElementsInProgram: dataElements,
+            teiValues: teiAttributesValues,
+            eventsData: getEventsDataFromEnrollment(enrollment, dataElements),
+            enrollmentData: getEnrollmentData(enrollment),
+        });
+    }
+    return undefined;
```
simonadomnisoru, line 108: Hey, I am wondering if it makes sense to return something relevant here and maybe show it to the user. WHYT?

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/runRulesForEnrollment.js
```diff
@@ -0,0 +1,109 @@
...
+        enrollmentId: event.enrollment,
+        enrollmentStatus: event.enrollmentStatus,
+        status: event.status,
+        eventDate: convertDate(event.dueDate),
+        dueDate: convertDate(event.dueDate),
+        ...getEventValuesFromEvent(enrollment, event.event, dataElements),
+    }));
+
+    const groupByProgramStageId = allEvents.reduce((acc, currentEvent) => {
+        acc[currentEvent.programStageId] = { ...currentEvent };
```
JoakimSM, line 63: we need an array of events here (multiple events can have the same programStageId)

JoakimSM (commented):
Almost there @huocha 🙂 

Couple of new comments and also we need to resolve these:
https://github.com/dhis2/capture-app/pull/1738#discussion_r644055773
https://github.com/dhis2/capture-app/pull/1738#discussion_r644071190
https://github.com/dhis2/capture-app/pull/1738#discussion_r644050257

src/core_modules/capture-core/components/Pages/Enrollment/runRulesForEnrollment.js
```diff
@@ -0,0 +1,111 @@
...
+        const dataValueList = flatDataValuesFromEvents(enrollments[0].events);
+        const dataElements = getDataElementsFromProgram(programMetadata, dataValueList);
+
+        const trackedEntityAttributes = attributes.reduce((acc, item) => {
+            acc[item.attribute] = { id: item.attribute, valueType: item.valueType };
+            return acc;
+        }, {});
+
+        const teiAttributesValues = attributes?.reduce((acc, item) => {
+            acc[item.attribute] = item.value;
```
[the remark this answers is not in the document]
huocha: Hi @JoakimSM, I wonder why this is clientToServer? I suppose that `attributes` is data from server and we want to convert to client so serverToClient?

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/runRulesForEnrollment.js
huocha: Hey @simonadomnisoru, thanks for the comment. I think this should run in the background and not showing to the user. Maybe we can add a log for debugging purpose but I will let other team members decide @JoakimSM WDYT?

src/core_modules/capture-core/components/Pages/Enrollment/runRulesForEnrollment.js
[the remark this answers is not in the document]
JoakimSM: Ah, serverToClient is correct (typo) 🙂

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/runRulesForEnrollment.js
JoakimSM: I think the logic here is sound. The function returns undefined until we have all the required data.

JoakimSM (approved):
Thanks @huocha for putting up with all my comments on this one 🙂

PR #1744, opened 2021-05-27
TECH-426: Refactor UNSAFE lifecycle

j-nguyenn:
This is my approach to refactor UNSAFE lifecycle 

1. `UNSAFE_componentWillMount` 
This  s invoked just before mounting occurs, ideally should be moved to `constructor`. In case it needs to await for a function, I use `async componentDidMount`.

2. `UNSAFE_componentWillReceiveProps`
This is more tricky and mostly depends on each case. [The official website](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change) suggests to use `getDerivedStateFromProps` or `componentDidUpdate` .
In our case if we want to update the state when the props change without overwriting the current state when state update, I choose to use `componentDidUpdate`.

Please suggest if you have other ideas or want a discussion.

### DIFF
--- a/src/core_modules/capture-ui/PolygonField/DeleteControl.component.js
+++ b/src/core_modules/capture-ui/PolygonField/DeleteControl.component.js
@@ -14,7 +14,7 @@ type Props = {
 
 // $FlowFixMe[prop-missing] automated comment
 class DeleteControlPlain extends MapControl<any, Props> {
-    omponentDidMount() {
+    componentDidMount() {
         const deleteControl = L.control({ position: 'topright' });
         const text = i18n.t('Delete polygon');
         const jsx = (
--- a/src/core_modules/capture-ui/PolygonField/DeleteControl.component.js
+++ b/src/core_modules/capture-ui/PolygonField/DeleteControl.component.js
@@ -14,7 +14,7 @@ type Props = {
 
 // $FlowFixMe[prop-missing] automated comment
 class DeleteControlPlain extends MapControl<any, Props> {
-    UNSAFE_componentWillMount() {
+    omponentDidMount() {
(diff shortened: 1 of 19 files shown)


src/core_modules/capture-ui/PolygonField/DeleteControl.component.js
```diff
@@ -14,7 +14,7 @@ type Props = {
 
 // $FlowFixMe[prop-missing] automated comment
 class DeleteControlPlain extends MapControl<any, Props> {
-    UNSAFE_componentWillMount() {
+    omponentDidMount() {
```
simonadomnisoru, line 17: Hey,
There is a typo with a missing `c `in this name. 
Thanks!

simonadomnisoru:
Hey @huocha,

Congrats on this PR. This kind of migration is always tricky 🙌.

I was curious what is the testing plan for the components affected by the refactoring. How can we ensure that the logic and functionality are not affected? 
Secondly, I wanted to suggest removing the rules from [eslint config ](https://github.com/dhis2/capture-app/blob/master/.eslintrc#L135) that allow the UNSAFE lifecycle in the project. 
Thank you!

simonadomnisoru:
> Hey @huocha,
> 
> Congrats on this PR. This kind of migration is always tricky 🙌.
> 
> I was curious what is the testing plan for the components affected by the refactoring. How can we ensure that the logic and functionality are not affected?
> Secondly, I wanted to suggest removing the rules from [eslint config ](https://github.com/dhis2/capture-app/blob/master/.eslintrc#L135) that allow the UNSAFE lifecycle in the project.
> Thank you!

Hey @huocha, let me know the testing/QA strategy for this refactor and I can approve the PR. Thank you!

PR #1754, opened 2021-05-31
feat: DHIS2-11236_Shared Hook for the information widgets on the enrollment dashboard

EirikHaugstulen:
This is a suggestion of how we can handle the filtering of rules returned from the `RulesEngine` for rendering in the information widgets. (Feedback, Error, Warning, Indicators) 
The hook takes the rule object from the RulesEngine as a input and returns an object of the filtered data.
Using destructuring, you can get the values directly, i.e.
`const { showError, showWarning } = useFilteredWidgetData(rules);`

I think this might increase reusability.
Thanks!


simonadomnisoru:
Hey @EirikHaugstulen, 
Great job on this PR!  I have 2 suggestions:
Firstly, I feel that the code will be clear if you move all the values for `rule.id` and `rule.type` in 2 const objects. 
Secondly, I personally prefer to use the Spread syntax to append items in an array instead of the `push` method. If you want to use it then the code can become something like this `showWarning = [...showWarning, rule];`
Let me know what is your opinion regarding these suggestions. Thanks!

EirikHaugstulen:
@simonadomnisoru Thanks, updated with the spread operator now! 
Could you elaborate a bit on what you meant with `rule.id` & `rule.type` in an object? 
Not sure I understood. Do you mean saving all values in an Array and only iterating this array? 
Thank you!

simonadomnisoru:
Hey @EirikHaugstulen, 

You compare `rule.type` with the following strings: `DISPLAYTEXT` `SHOWWARNING` `SHOWERROR` `DISPLAYKEYVALUEPAIRS` `DISPLAYTEXT` and rule.id with `general` `feedback` `indicators`. In my opinion, is clear if you move these strings to 2 `const` objects. You can see an example [here](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetEnrollment/Status/Status.component.js#L13). 
Thanks!

EirikHaugstulen:
@simonadomnisoru Thank you for the feedback! 
I think I've implemented your changes, but be sure to let me know if I haven't.

Converting this to a draft for now as I suspect there are more features that has to be implemented.

Link to ticket: https://jira.dhis2.org/browse/DHIS2-11236

src/core_modules/capture-core/components/Pages/Enrollment/hooks.js
```diff
@@ -11,3 +21,49 @@ export const useEnrollmentInfo = (enrollmentId: string, programId: string) => {
...
+            }
+        } else if (rule.id === ruleIDs.feedback) {
+            feedback = [...feedback, rule?.displayText || rule?.displayKeyValuePair];
+        } else if (rule.id === ruleIDs.indicators) {
+            indicator = [...indicator, rule?.displayText || rule?.displayKeyValuePair];
+        }
+    });
+
+    return {
+        showWarning,
```
JoakimSM, line 64: suggestion: showWarning -> warnings, showError => errors, feedback => feedbacks, indicator => indicators

EirikHaugstulen:
Thanks, should be updated! 👍

JoakimSM (approved):
LGTM 👍

PR #1757, opened 2021-06-01
feat: DHIS2-10981_DataConversionForWidgetFeedback

EirikHaugstulen:
This PR introduces a stripped down version of the WidgetFeedback that only takes two input-parameters; "displayText" & "displayKeyValue". The component then renders the items in a list form or returns a string of no feedback. 

https://jira.dhis2.org/browse/DHIS2-10981

image


src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,63 @@
...
+const styles = {
+    container: {
+        padding: `0px ${spacersNum.dp16}px`,
+    },
+    unorderedList: {
+        paddingLeft: '15px',
+        marginTop: '0px',
+        lineHeight: '1.375',
+        fontSize: '14px',
+        color: '#212934',
```
huocha, line 18: Hello I think this can be `colors.grey900` from dhis2-ui

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,63 @@
...
+    noFeedbackText: {
+        color: '#6B7280',
+        paddingLeft: '15px',
+        fontSize: '14px',
+        marginTop: '0px',
+    },
+    listItem: {
+        marginBottom: '5px',
+        '&::marker': {
+            color: '#A0ADBA',
```
huocha, line 29: this can be `colors. grey500`

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,63 @@
...
+const WidgetFeedbackContentComponent = ({ displayText, displayKeyValue, classes }: Props) => {
+    if ((!displayText || displayText.length <= 0) && (!displayKeyValue || displayKeyValue.length <= 0)) {
+        return <p className={classes.noFeedbackText}>{i18n.t('No feedback for this enrollment yet')}</p>;
+    }
+
+    return (
+        <div className={classes.container}>
+            <ul className={classes.unorderedList}>
+                {displayText && displayText.length > 0 ? displayText.map(item => (
+                    item.message && (
```
huocha, line 43: I prefer extract this render block into function just for readability and avoid shadow name

huocha (commented):
Great work. I add some minor comments to reuse the [colors constants](https://github.com/dhis2/ui/blob/master/packages/constants/src/colors.js) from dhis2-ui

EirikHaugstulen:
@huocha Thanks, much appreciated 👍
Implemented all changes!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,73 @@
...
+        marginBottom: '5px',
+        '&::marker': {
+            color: colors.grey500,
+        },
+    },
+};
+
+
+const WidgetFeedbackContentComponent = ({ displayText, displayKeyValue, classes }: Props) => {
+    if ((!displayText || displayText.length <= 0) && (!displayKeyValue || displayKeyValue.length <= 0)) {
```
simonadomnisoru, line 37: Hey, 
I think you can use [Optional chaining ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and the code will be more clear. WDYT?
`displayText?.length <= 0 && displayKeyValue?.length <= 0`
Thanks!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,73 @@
...
+        return <p className={classes.noFeedbackText}>{i18n.t('No feedback for this enrollment yet')}</p>;
+    }
+
+    const renderText = () => (
```
simonadomnisoru, line 41: You already check above for `displayText.length` and `displayKeyValue.length`, there is no need to check again inside `renderKeyValue` and `renderText`

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,73 @@
+// @flow
+
+import React, { type ComponentType } from 'react';
+import { spacersNum, colors } from '@dhis2/ui';
+import i18n from '@dhis2/d2-i18n';
+import { withStyles } from '@material-ui/core';
+import type { Props } from '../WidgetFeedback.types';
+
+const styles = {
```
simonadomnisoru, line 9: Hey,
Can you help me understand why you used `5px`, `14px`, `15px` instead of the [SpacerNum const ](https://ui.dhis2.nu/demo/?path=/story/using-ui-useful-constants--page)?

Thank you!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
EirikHaugstulen: Yep, definitely increased readability 👍

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
EirikHaugstulen: Not to familiar with dhis/ui, but will be sure to study it in the coming days! 
I just used the pixel-sizes from the sketch-inspector. I've replaced most of them with the nearest SpacerNum-constant.
(Did not seem to work on font-sizes, but I will investigate it further)
Thanks!

simonadomnisoru (approved):

EirikHaugstulen:
Edit: changed the behavior of the widget to only take one prop as input. This prop is an Array of either string, object with id&message, or an object with id, key and value. 
I think this would both increase reusability and make it easier to use. 
This content-component will also be used in the indicator-widget.

EirikHaugstulen:
Added feature: After discussion with Markus and Joakim, we spoke about why we could not hide the feedback-widget if there were no messages to be displayed. Markus introduced a use case where the feedback widget would be used to display essential health data about a patient (Allergies, medicine and so on). If we removed the feedback widget entirely, this would raise the question if there really are no feedbacks or if the end user just could not find it. 
The solution to this is that we check whether there are `Program Rule Actions` that can potentially result in a displayed message in the feedback widget. If there are no such action, we can assume that the widget is not particularly important, and therefore hide it from the dashboard, saving space.

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.types.js
```diff
@@ -1,5 +1,23 @@
 // @flow
 
+export type filteredText = {|
```
JoakimSM, line 3: I suggest we use PascalCase for the type declarations (FilteredText)

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.js
```diff
@@ -0,0 +1,89 @@
...
+        '&::marker': {
+            color: colors.grey500,
+        },
+    },
+};
+
+
+const WidgetFeedbackContentComponent = ({ widgetData, classes }: contentType) => {
+    if (!widgetData?.length) {
+        return <p data-test={'widget-content'} className={classes.noFeedbackText}>{i18n.t('No feedback for this enrollment yet')}</p>;
```
JoakimSM, line 38: I think we must pass in this emptyText to the component (will be another text when we're looking at one specific event)

JoakimSM (commented):
Thanks @EirikHaugstulen. Looks pretty good 👍 . Some comments that I hope you can look into.

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
```diff
@@ -39,14 +40,22 @@ export const EnrollmentPageDefault = () => {
...
     useEffect(() => {
         const effects = runRulesForEnrollment({ orgUnit, program, programMetadata, enrollment, attributes });
         if (effects) {
+            // $FlowFixMe
             setRuleEffects(effects);
         }
+        const flattenedRuleActionLocations = program.programRules.map(item => item.programRuleActions
+            .map(rule => rule.location || null))
+            .flat();
+        setHideFeedbackWidget(!flattenedRuleActionLocations.includes('feedback'));
```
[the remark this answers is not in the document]
EirikHaugstulen: Yup, thanks!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.types.js
EirikHaugstulen: Sorry, didn't know. Thanks for enlightening me 👍

JoakimSM (approved):
Thanks @EirikHaugstulen , Looks good 👍

PR #1771, opened 2021-06-02
feat: DHIS2-11044_ErrorWidget

EirikHaugstulen:
This PR is a suggestion of how to use the `WidgetErrorAndWarningContent`-component to render the error & widget-content. 
The widget takes array of either strings or objects and a type.
The content-component will then return the list with styling based on the input given.

https://jira.dhis2.org/browse/DHIS2-11044

image
image

Important edit!
This PR is heavily dependent on:
[DHIS2-10981](https://github.com/dhis2/capture-app/pull/1757)


src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.component.js
```diff
@@ -1,23 +1,29 @@
...
+
+export const WidgetFeedback = ({ feedback, emptyText }: Props) => {
+    const [openStatus, setOpenStatus] = useState(true);
 
-export const WidgetFeedback = ({ className }: Props) => {
-    const [openStatus, setOpenStatus] = useState(false);
     return (
         <div
-            className={className}
+            data-test={'feedback-widget'}
```
simonadomnisoru, line 13: Hey, there is no need to wrap the data test id inside the `{ }`. Thanks!

EirikHaugstulen: Thanks, updated it in the [Indicator-PR](https://github.com/dhis2/capture-app/pull/1831) 👍

simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetErrorAndWarning/content/WidgetErrorAndWarningContent.types.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import type { Rule, RuleObject } from '../WidgetError/WidgetError.types';
```
JoakimSM, line 2: 1. I would define these types here instead of importing from WidgetError (and import the types defined here in WidgetError, so, the other way around)
2. I would change the naming here. The data that are passed in to the Widgets doesn't have to come from the rules engine and is really just a collection of messages (errors or warnings).

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.types.js
```diff
@@ -13,6 +13,7 @@ export type Props = {|
     widgetEffects: ?Object,
     hideWidgets: HideWidgets,
     onDelete: () => void,
+    widgetEffects: Object,
```
JoakimSM, line 16: maybe the structure here can be imported from somewhere?

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
```diff
@@ -50,6 +51,7 @@ export const EnrollmentPageDefaultPlain = ({
                 <WidgetStagesAndEvents stages={program.stages} />
             </div>
             <div className={classes.rightColumn}>
+                <WidgetError error={widgetEffects?.errors} />
```
JoakimSM, line 54: I would suggest we hide this if there are no error effects at the moment. `widgetEffects?.errors && <WidgetError..`

JoakimSM (commented):
Thanks @EirikHaugstulen. Looks quite good. Couple of comments.

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
EirikHaugstulen: So remove the hiding-logic from inside the component? Or do we still keep that on as a fallback if the widget is used elsewhere?

JoakimSM: Lets keep your suggested logic (discussed in meeting)

PR #1774, opened 2021-06-04
chore: temporary remove flaky locked selector tests

JoakimSM:
Let's temporary remove these flaky tests until the locked selector is refactored here: https://jira.dhis2.org/browse/TECH-580


simonadomnisoru (approved):
Great idea!

PR #1782, opened 2021-06-08
Tech 614 - Non Collapsible Widget

EirikHaugstulen:
This PR introduces a collapsible and non-collapsible widget. 
The component is still called `Widget`, but if `collapsible` is not passed as prop, the widget will default to the collapsible-widget. The widget can also take a `color`-prop that will set the background color if needed. This will revert to white of nothing is passed in.
There is no need to implement any changes to the already existing widgets.

https://jira.dhis2.org/browse/TECH-614

### DIFF
--- a/src/core_modules/capture-core/components/Widget/Widget.component.js
+++ b/src/core_modules/capture-core/components/Widget/Widget.component.js
@@ -4,15 +4,15 @@ import { WidgetCollapsible } from './WidgetCollapsible.component';
 import { WidgetNonCollapsible } from './WidgetNonCollapsible.component';
 import type { Props } from './widget.types';
 
-export const Widget = ({ collapsible = true, className, ...passOnProps }: Props) => (
+export const Widget = ({ collapsible = true, ...passOnProps }: Props) => (
     collapsible ? (
-        <div className={className}>
+        <div>
             { /* $FlowFixMe */ }
             <WidgetCollapsible
                 {...passOnProps}
             />
         </div>) :
-        (<div className={className}>
+        (<div>
             { /* $FlowFixMe */ }
             <WidgetNonCollapsible
                 {...passOnProps}
--- a/src/core_modules/capture-core/components/Widget/Widget.component.js
+++ b/src/core_modules/capture-core/components/Widget/Widget.component.js
@@ -6,11 +6,12 @@ import type { Props } from './widget.types';
 
 export const Widget = ({ collapsible = true, className, ...passOnProps }: Props) => (
     collapsible ? (
-        // $FlowFixMe doesn't work due to destructering :(
         <div className={className}>
-        <WidgetCollapsible
-            {...passOnProps}
-        /></div>) : (
+            { /* $FlowFixMe */ }
+            <WidgetCollapsible
+                {...passOnProps}
+            />
+        </div>) : (
             // $FlowFixMe doesn't work due to destructering :(
             <WidgetNonCollapsible
                 {...passOnProps}
--- a/src/core_modules/capture-core/components/Widget/Widget.component.js
+++ b/src/core_modules/capture-core/components/Widget/Widget.component.js
@@ -4,8 +4,8 @@ import { WidgetCollapsible } from './WidgetCollapsible.component';
 import { WidgetNonCollapsible } from './WidgetNonCollapsible.component';
 import type { Props } from './widget.types';
 
-export const Widget = ({ collapsible = true, ...passOnProps }: Props) => (
-    collapsible ? (
+export const Widget = ({ noncollapsible = false, ...passOnProps }: Props) => (
+    !noncollapsible ? (
         <div>
             { /* $FlowFixMe */ }
             <WidgetCollapsible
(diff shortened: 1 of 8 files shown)


src/core_modules/capture-core/components/Widget/Widget.component.js
```diff
@@ -1,144 +1,21 @@
...
-            setChildrenVisibility(true);
-        } else {
-            hideChildrenTimeoutRef.current = setTimeout(() => {
-                setChildrenVisibility(false);
-            }, 200);
-        }
-    }, [open, animationsReady]);
-
-    return (
+export const Widget = ({ collapsible = true, ...passOnProps }: Props) => (
```
simonadomnisoru, line 7: Hey @EirikHaugstulen, 

I am a bit confused about the usage of the prop `collapsible`. I think it can be misleading to have a missing prop default value set to `true`. How about setting it to `false`? You will just have to add it as a prop to the components that already use `Widget` and it will work. WDYT?

EirikHaugstulen: I can see how this might be misleading, and I'm not sure of what is the best solution, but I can at least explain what I think is the thought-process behind this:

At least my own mental model on using the widget component is that it should be collapsible. If it's not collapsible, it's just a basic tile or div with a header. This collapsible-prop would therefore be a way to override this if necessary. (To increase component reusability). I also think that based on the design sketches of the enrollment dashboard, we use the collapsible widget a whole lot more than the non-collapsible widget and therefore this would be an edge case. (error & warning mostly)

Highly open for refactoring this if you still think it is better to default the prop to false. ☺️
Thanks!

simonadomnisoru: How about using a prop named `noncollapsible`? If `noncollapsible` is undefined or false then the default behavior will apply and the widget will be collapsible. For the edge cases when we need to override de default behavior we can pass the prop `noncollapsible` to the Widget. Do you think it will be less confusing?

EirikHaugstulen: Yep, that sounds like a good idea. 
I'll implement the changes suggested!

JoakimSM: Great discussion here @EirikHaugstulen @simonadomnisoru (I was the one introducing `collapsible = true` before @EirikHaugstulen inherited this task). Naming variables are really hard sometimes 🙂 Gave it some thought and I agree with you @simonadomnisoru. I think what they say here makes sense: https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables. TLDR; Avoid negative names. But ALWAYS choose names that enable setting “false” by default.

src/core_modules/capture-core/components/Widget/WidgetNonCollapsible.component.js
```diff
@@ -0,0 +1,43 @@
...
+        alignItems: 'center',
+        padding: spacersNum.dp16,
+        fontWeight: 500,
+        fontSize: 16,
+        color: colors.grey800,
+    },
+};
+
+const WidgetNonCollapsiblePlain = ({ header, children, color, classes }: WidgetNonCollapsiblePropsPlain) => (
+    <div className={classes.container}>
```
JoakimSM, line 25: Any reason why we don't set `style={{ backgroundColor: color || colors.white }}` here (instead of setting it twice)?

src/core_modules/capture-core/components/Widget/widgetCollapsible.types.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import type { Node } from 'react';
+
+export type WidgetCollapsibleProps = {|
+    header: Node,
+    children: Node,
+    open: boolean,
+    onOpen: () => void,
+    onClose: () => void,
+    color?: ?string,
```
JoakimSM, line 10: Can we do `color?: string` instead. This way we don't allow `null` to be passed in and the benefit is that we can assign default values in the function declarations. For example in `WidgetNonCollapsible.component.js ` we can declare the function as this:
`const WidgetNonCollapsiblePlain = ({ header, children, color = colors.white, classes }: WidgetNonCollapsiblePropsPlain) =>`

src/core_modules/capture-core/components/Widget/widget.types.js
```diff
@@ -1,10 +1,17 @@
...
-export type Props = {|
-    header: React$Node,
-    children: React$Node,
-    open: boolean,
-    onOpen: () => void,
-    onClose: () => void,
-    ...CssClasses,
+type CollapsibleProps = {|
+    ...WidgetCollapsibleProps,
+    color?: ?string,
```
JoakimSM, line 7: `color` is part of the `WidgetCollapsibleProps` so we don't need to readd it here. Same below.

JoakimSM (commented):
Looks good @EirikHaugstulen. Couple of comments I hope you can look into.

EirikHaugstulen:
Thanks for the feedback! Should be updated now.

simonadomnisoru (approved):
LGTM! Congrats!

JoakimSM (approved):
Good work @EirikHaugstulen!

PR #1788, opened 2021-06-10
feat: [DHIS2-10222] Stages & Events

huocha:
Recreate from DHIS2-10218 due to incorrect ticket name


src/core_modules/capture-core/components/WidgetStagesAndEvents/types/common.types.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+type DataValue = {
```
simonadomnisoru, line 3: Hey,

Do you think it makes sense to reuse `type ApiDataValue` and `type ApiTEIEvent` from `getEnrollmentEvents.js` instead of creating new types? 
Thanks!

huocha: @simonadomnisoru  thanks for the suggestion, I didn't know this type exist :) updated the branch

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/helpers.js
```diff
@@ -0,0 +1,84 @@
+// @flow
+import i18n from '@dhis2/d2-i18n';
+import moment from 'moment';
+import type { ApiTEIEvent } from 'capture-core/events/getEnrollmentEvents';
+import { convertValue as convertClientToView } from '../../../../converters/clientToView';
```
JoakimSM, line 5: I suggest we use `clientToList` instead

JoakimSM (commented):
Hey @huocha! Thanks for the PR, there's a lot to like in here 👍 

Some logic we will need to change though. Apart from the comments in the code, we also need to make some separate api requests to fetch value data we don't get from the main request. An example is data elements of type organisation unit, the value we get from the main request for these are the org unit id, but we would like to display the name. This file is a great example of how to do this: https://github.com/dhis2/capture-app/blob/DHIS2-10222/src/core_modules/capture-core/components/Pages/MainPage/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/getTeisWithSubvalues.js. The file is for TEIs (tracked entity instances) though and we need to make some changes. So I suggest, for now, that you try to use the same principle for events and create a new file in the stages and events widget.

My suggestion for where we should do these supplementary api requests: Let's continue to pass in the server values to the module (better if other apps will use this Widget in the future). As one of the first operations inside the Widget we will need to grab this supplementary data.

My comments here are probably not crystal clear, so let's have a talk when you want to start digging into this 🙂

huocha:
@JoakimSM  thank you for your feedback. As discussed, I added the subvalue for Organisation unit, file source and image [here](https://github.com/dhis2/capture-app/pull/1788/files#diff-6a1b0af5f0764c0a9c1f45e3f9427aefa731e1889a26fdc46e40975a7e8cb90f) and also resolve other comments.

I agree with the idea of making Widget with style props options (In the mockup the widget still comes with a border bottom, but maybe later we want something completely borderless).
Please let me know if there is anything else to change.

simonadomnisoru (approved):

JoakimSM (commented):
Sorry about the really long comment in here, I had to get my thoughts on the matter written down in a structured way. We should talk about that one!

JoakimSM:
The borderless mode for the Widget is merged to master just now. Can we update the UI to be more in line with the sketch? PR: https://github.com/dhis2/capture-app/pull/1843
previous comment: https://github.com/dhis2/capture-app/pull/1788#discussion_r652626492

huocha:
Thank you for your feedback @JoakimSM, as discussed I changed the data structure and batch fetch the subvalue, 
the options and sorting will be done in separate PRs. Can. you have a look?

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/StageDetail.component.js
```diff
@@ -0,0 +1,188 @@
...
+    DataTableHead,
+    DataTableFoot,
+    DataTable,
+    DataTableRow,
+    DataTableCell,
+    DataTableColumnHeader,
+    Button,
+} from '@dhis2/ui';
+import { sortDataFromEvent } from '../hooks/sortFuntions';
+import { useComputeDataFromEvent, useComputeHeaderColumn } from '../hooks/useEventList';
```
JoakimSM, line 17: Can we move these files closer to where they are used? Meaning, since these files are only used in this file I would move the files to the StageDetail folder.

JoakimSM (commented):
We're getting there 🙂  We can have a chat tomorrow or something if anything is unclear.

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/sortFuntions.js
```diff
@@ -0,0 +1,81 @@
...
+
+const sortTime = (timeA: string, timeB: string, direction: string) => {
+    if (direction === SORT_DIRECTION.ASC) {
+        return moment(timeA).unix() - moment(timeB).unix();
+    }
+    if (direction === SORT_DIRECTION.DESC) {
+        return moment(timeB).unix() - moment(timeA).unix();
+    }
+    return 0;
+};
```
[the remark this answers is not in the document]
huocha: do you mean using server value? because this should be in client value already (same as display in UI)

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/StageDetail.component.js
huocha: the helpers also used in `Stage`, but as you suggested, I move all folder to `StageDetail`

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/StageDetail.component.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+        return dataSource
+            .sort((a, b) => {
+                const { type } = headerColumns.find(col => col.id === columnName) || {};
+                // $FlowFixMe
+                return sortDataFromEvent(a[columnName], b[columnName], type, sortDirection);
+            })
+            .slice(0, displayedRowNumber)
+            .map((row, index) => {
+                const cells = Object.keys(row)
```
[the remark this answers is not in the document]
JoakimSM: I'll admit this isn't that important since it's working as it is. I'll close this one.

JoakimSM (approved):
Thanks @huocha! Good job getting this over the finish line 🎉

PR #1790, opened 2021-06-10
fix: wait for field validations on save

JoakimSM:
When updating our app to use a 2.37 test instance I encountered a problem where a form could be submitted before the initial field validations were complete (required for compulsory fields). I would have loved to rework this from scratch, but the best solution I found for now is to wait if this happens (I don't think you should be able to experience this if you're not running this using a testing engine)


JoakimSM:
superseded by https://github.com/dhis2/capture-app/pull/1791

PR #1791, opened 2021-06-10
fix: TECH-622 use 2.37 backend for tests

JoakimSM:
Use 2.37 backend for tests.

There is also a fix in here. Ideally I would have created a separate PR for the fix, but the 2.36 test backend was updated and caused some tests to fail and I didn't want to fix those tests temporarily and then change them again after updating to use the 2.37 backend. The fix is: When updating I encountered a problem where a form could be submitted before the initial field validations were complete (required for compulsory fields). I would have loved to rework this from scratch, but the best solution I found for now is to wait if this happens (I don't think you should be able to experience this if you're not running this using a testing engine)

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
 NODE_PATH=src/D2App/core_modules
 REACT_APP_CACHE_VERSION=$npm_package_cacheVersion
 REACT_APP_VERSION=$npm_package_version
-REACT_APP_DHIS2_API_VERSION=36
+REACT_APP_DHIS2_API_VERSION=37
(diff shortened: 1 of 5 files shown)


simonadomnisoru:
Hey @JoakimSM, 

I have some suggestions/questions. I think that another approach for this fix is to disable the form submit button until the initial field validations are complete. Do you think this approach will provide a better UI experience for our users? 

When the user is on a slow connection and it may take a while to run the field validations, how are we letting them know that we are processing the request and preventing multiple clicks on the Submit button? 

Thank you!

simonadomnisoru (approved):

PR #1803, opened 2021-06-15
feat: [DHIS2-11208] Enrolment event form using the view/edit single event component

simonadomnisoru:
[DHIS2-11208](https://jira.dhis2.org/browse/DHIS2-11208)

Display the enrollment event form using the existing single event form. 

To trigger the proper Redux reducers and get the data I added the new enrollmentEventEdit page to `withAppUrlSync.js`. 

`ViewEventDataEntry` , `EditEventDataEntry` and `getRulesActionsForEvent` were adapted to work for a program of type `TrackerProgram`. One of the main differences between the programs is how the stage is stored. The `EventProgram` has only one stage and it is retrieved in the code using `program.stage`. Meanwhile, a `TrackerProgram` has many stages and needs to be retrieved using the `stageId`.


src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
```diff
@@ -18,12 +23,33 @@ export const EnrollmentEventPage = () => {
...
     const programStage = [...program.stages?.values()].find(
         item => item.id === stageId,
     );
+
     return (
-        <EnrollmentEventPageComponent mode={mode} programStage={programStage} />
+        <EnrollmentEventPageComponent
+            mode={showEditEvent ? pageMode.EDIT : pageMode.VIEW}
+            programStage={programStage}
+            onEdit={() => onOpenEditEvent()}
```
JoakimSM, line 35: I think we should approach this more like you did at first. We should let the consumer of the `WidgetEventEdit` set the mode and then we deal with the complexity internally in the Widget. For example, the action `startShowEditEventDataEntry` we can call from an `useEffect` if the mode is set to edit. We still need the mode and onEdit callback here, but these will be values from useState.

src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
```diff
@@ -18,12 +23,33 @@ export const EnrollmentEventPage = () => {
...
+        />
     );
 };
+
+const mapStateToProps = (state: ReduxState) => ({
+    showEditEvent: state.viewEventPage?.eventDetailsSection?.showEditEvent,
```
JoakimSM, line 41: Maybe we can move the location of the state here to something like `state.widgetEventEdit`?

src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
```diff
@@ -18,12 +23,33 @@ export const EnrollmentEventPage = () => {
...
+        dispatch(startShowEditEventDataEntry());
+    },
+});
+
+// $FlowSuppress
+// $FlowFixMe[missing-annot] automated comment
+export const EnrollmentEventPage = connect(
+    mapStateToProps,
+    mapDispatchToProps,
+)(EnrollmentEventPagePlain);
```
JoakimSM, line 55: This might be changed / moved based on another comment of mine, but it's great if we use `useDispatch` and `useSelector` instead of the legacy connect.

cypress/integration/EnrollmentPage/WidgetEnrollment.feature
```diff
@@ -1,25 +1,25 @@
 Feature: User interacts with Enrollment Widget
 
-    Scenario: User can close the Enrollment Widget
-        Given you open the enrollment page
-        When you click the enrollment widget toggle open close button
-        Then the enrollment widget should be closed
+    # Scenario: User can close the Enrollment Widget
```
JoakimSM, line 3: Did you intend to remove all these tests? (this file and TeiWorkingListsUser.feature)

JoakimSM (commented):
Looks quite good @simonadomnisoru 👍  Hope you can have a look into my comments.

src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
simonadomnisoru: If I change anything in the Redux state path then the old redux actions and reducers will not work anymore. Therefore if we don't want a big refactor I think is best to use the same path that are being used in the single event form

simonadomnisoru:
> Looks quite good @simonadomnisoru 👍 Hope you can have a look into my comments.

Hey @JoakimSM,
I implemented the suggestions from your comments. Can you have a look? 
Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
JoakimSM: Ok, let's leave it for now.

JoakimSM:
Looks good, I think you missed this one though: https://github.com/dhis2/capture-app/pull/1803#discussion_r662268754

I would encourage you to look into that one a bit, would make it easier to reason about the Widget.

src/core_modules/capture-core/components/Pages/EnrollmentEvent/EnrollmentEventPage.container.js
simonadomnisoru: Hey @JoakimSM,

 The switch between the form edit/view mode comes from the redux store inside `state.viewEventPage.eventDetailsSection.showEditEvent`. This value is currently modified inside:

1. `EditEventDataEntry` with the `Save` button and the `Cancel` button
2. `WidgetEventEdit` with the Edit button (`dispatch` of `startShowEditEventDataEntry`)

If I understand you correctly, you suggest that the page `EnrollmentEventPage` should handle the form edit/view mode and store it inside its state. In this case, I would need to change the logic for the `Save` button and the `Cancel` button inside the `EditEventDataEntry`. I will probably need to remove their current logic that modifies `state.viewEventPage.eventDetailsSection.showEditEvent` and add it inside `EnrollmentEventPage` using callbacks or dispatches. The old logic will also have to be maintained as long as we still use this form for single events. 

I prefer to have `EditEventDataEntry`, `ViewEventDataEntry` and their parent `WidgetEventEdit` handle the form edit/view mode without involving the page. Let's have a chat via Slack when you are back to check what is the best strategy. 
Thank you!

JoakimSM (approved):
Great. Looks good 👍

PR #1809, opened 2021-06-17
fix: Cypress post upgrade

JoakimSM:
Changes:
- Automatically start the app when running `yarn cy:open` and `yarn cy:run`
- Some time ago we made some custom scripts in our app to deal with some Cypress edge cases. An example would be be if you create a `.env.development` or `.env.development.local` file locally and then run `yarn cy:open`; in this case you would, without the scripts, run against your local instance while the api requests made in your tests will run against your test instance. To fix this we specify our Cypress config in the `.env.cypress` file and the server specified here will always be used. Updated the variable names in this file instead of `cypress.json`.
- Fixed some tests after updating the DB

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,5 +1,4 @@
 const {
-    networkShim,
     chromeAllowXSiteCookies,
     cucumberPreprocessor,
 } = require('@dhis2/cypress-plugins');
@@ -8,7 +7,6 @@ const getCypressEnvVariables = require('./getCypressEnvVariables');
 const path = require('path');
 
 module.exports = (on, config) => {
-    networkShim(on);
     chromeAllowXSiteCookies(on);
     cucumberPreprocessor(on, config);
     on('before:browser:launch', (browser, launchOptions) => {
--- a/cypress/support/index.js
+++ b/cypress/support/index.js
@@ -1,5 +1,3 @@
-import { enableNetworkShim } from '@dhis2/cypress-commands';
-
 // Add additional support functions here
 import './helpers';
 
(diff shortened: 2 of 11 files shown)


cypress/plugins/index.js
```diff
@@ -8,7 +7,6 @@ const getCypressEnvVariables = require('./getCypressEnvVariables');
 const path = require('path');
 
 module.exports = (on, config) => {
-    networkShim(on);
```
JoakimSM, line 11: @simonadomnisoru I don't think we need this, but I might be wrong. Any insights?

cypress/support/index.js
```diff
@@ -12,4 +10,3 @@ beforeEach(() => {
     */
     cy.clearCookies({ domain: null });
 });
-enableNetworkShim();
```
JoakimSM, line 15: @simonadomnisoru same as other comment.

package.json
```diff
@@ -72,8 +72,8 @@
     "flow:check": "./node_modules/.bin/flow check ./src",
     "flow:addTypes": "flow-typed install",
     "lint": "npm run flow:check && npm run linter:check",
-    "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
-    "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
+    "cy:open": "concurrently --kill-others \"yarn start:forCypress\" \"wait-on 'http-get://localhost:3000' && cypress open\"",
```
simonadomnisoru, line 75: Good idea!

cypress/plugins/index.js
simonadomnisoru: You're right. We don't need it. We use live network requests therefore there is no need to record them and play them back.

simonadomnisoru (approved):

PR #1868, opened 2021-07-05
feat: [DHIS2-11465] load events filter by program stage id

huocha:


src/core_modules/capture-core/components/TeiWorkingLists/Setup/TeiWorkingListsSetup.component.js
```diff
@@ -197,6 +197,7 @@ export const TeiWorkingListsSetup = ({
             onLoadView={useInjectDataFetchingMetaToLoadList(defaultColumns, filtersOnly, onLoadView)}
             onUpdateList={useInjectDataFetchingMetaToUpdateList(defaultColumns, filtersOnly, onUpdateList)}
             programId={program.id}
+            programStageId=""
```
simonadomnisoru, line 200: Hey, why is the `programStageId` an empty string here? Thank you!

huocha: hey @simonadomnisoru, I didn't intend to add programStageId for TrackerEvent but it uses the same props and EventTracker in the WorkingList. I pushed a change to make it optional in the props.

simonadomnisoru (approved):
LGTM!

PR #1870, opened 2021-07-06
feat: [DHIS2-11420] handle display of options

huocha:


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/helpers.js
```diff
@@ -99,14 +99,17 @@ const getAllFieldsWithValue = async (
...
-
     return columns
         .filter(({ isPredefined }) => !isPredefined)
         .reduce((acc, { id }) => {
-            const { type } = dataElements.find(el => el.id === id) ?? {};
+            const { type, options } = dataElements.find(el => el.id === id) ?? {};
             if (type && records[id]) {
                 const value = dataElementsByType.find(item => item.type === type).ids[id];
-                acc[id] = formatValueForView(value, type);
+                if (options && options[value]) {
```
simonadomnisoru, line 108: Hey @huocha,

In the [example](https://github.com/dhis2/capture-app/blob/0a82ec3bfd72839ae53172d2f60577f111b96b59/src/core_modules/capture-core/components/Pages/MainPage/WorkingListsCommon/hooks/useDataSource.js#L35) provided in the [DHIS2-11420](https://jira.dhis2.org/browse/DHIS2-11420), there is a `log.error` when the option is not found. Do you think we should add it also here? 
Thanks!

huocha: hey @simonadomnisoru, thank you for the comment. I will add the log

simonadomnisoru (approved):

PR #1875, opened 2021-07-07
DHIS2-11422: sort function

huocha:


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/sortFuntions.js
```diff
@@ -0,0 +1,65 @@
...
+    [dataElementTypes.NUMBER]: sortText,
+    [dataElementTypes.INTEGER]: sortText,
+    [dataElementTypes.INTEGER_POSITIVE]: sortText,
+    [dataElementTypes.INTEGER_NEGATIVE]: sortText,
+    [dataElementTypes.INTEGER_ZERO_OR_POSITIVE]: sortText,
+    [dataElementTypes.BOOLEAN]: sortText,
+    [dataElementTypes.TRUE_ONLY]: sortText,
+    [dataElementTypes.DATE]: sortTime,
+    [dataElementTypes.DATETIME]: sortTime,
+    [dataElementTypes.TIME]: sortText,
```
simonadomnisoru, line 42: For `dataElementTypes.TIME` why do you use `sortText` instead `sortTime`? Thanks!

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/sortFuntions.js
```diff
@@ -0,0 +1,65 @@
...
+    [dataElementTypes.PERCENTAGE]: sortText,
+    [dataElementTypes.URL]: sortText,
+    [dataElementTypes.AGE]: sortText,
+    [dataElementTypes.ORGANISATION_UNIT]: sortText,
+    [dataElementTypes.COORDINATE]: sortText,
```
simonadomnisoru, line 47: As far as I remember `dataElementTypes.COORDINATE` is an object with the shape 
```
{
  longitude: number, 
  latitude: number
}
```
I think we need another helper function to be able to sort by the coordinates. What do you think?

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/sortFuntions.js
huocha: Thank you for spotting it, I changed that

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/hooks/sortFuntions.js
huocha: hey @simonadomnisoru, I'm aware of the rawObject of coordinate. in this case the value parsed here is already formatted in string.

simonadomnisoru (approved):

PR #1876, opened 2021-07-08
refactor: [TECH-641] Rules engine refactoring

superskip:
Moved the generic parts of the rules engine (i.e. most of it) to capture-core-utils.
Removed the static functions from the RulesEngine class.
Refactored "processTypes" out of rulesEffectsProcessor (thus it's possible this PR takes care of TECH-619).


src/core_modules/capture-core-utils/rulesEngine/RulesEngine.js
```diff
@@ -0,0 +1,346 @@
...
+    * @param {*} currentEvent selected event (null if not applicable)
+    * @param {*} eventsContainer list of all relevant events, or null
+    * @param {*} dataElements all data elements (metadata)
+    * @param {*} selectedEntity selected TEI (null if not applicable)
+    * @param {*} trackedEntityAttributes all tracked entity attributes (metadata)
+    * @param {*} selectedEnrollment selected enrollment (null if not applicable)
+    * @param {*} selectedOrgUnit selected OrgUnit
+    * @param {*} optionSets all option sets
+    */
+    getProgramRuleEffects(
```
huocha, line 229: Do you think it's better to use the object here instead of multiple params? It could be helpful when developers don't remember the order of parameters.

src/core_modules/capture-core/rules/converters/outputConverter.js
```diff
@@ -1,68 +1,106 @@
...
     }
+
     convertIntegerZeroOrPositive(value: number): string {
         return value.toString();
     }
-    convertPercentage(value: number): string {
-        return value.toString();
+
+    convertTrackerAssociate(value: string): any {
+        log.warn('convertTrackerAssociate not implemented', value);
```
huocha, line 85: hey @superskip, is this something we will implement in the future or what is the plan?

huocha (commented):
Great work @superskip 👍 I just have few comments

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/VariableService.js
```diff
@@ -66,12 +67,17 @@ export class VariableService {
             OptionSetHelper.getName(optionSets[trackedEntityAttributes[trackedEntityAttributeId].optionSetId].options, value)
             : value;
     }
-    static dateUtils = getDateUtils(momentConverter);
+
+    static dateUtils: any;
```
simonadomnisoru, line 71: I think `Object` is a better type for `dateUtils`. WDYT?

src/core_modules/capture-core-utils/rulesEngine/rulesEngine.types.js
```diff
@@ -252,34 +267,39 @@ export type D2FunctionParameters = {
...
-    'd2:checkControlDigits': D2FunctionParameters,
-    'd2:round': D2FunctionParameters,
-    'd2:lastEventDate': D2FunctionParameters,
     'd2:left': D2FunctionParameters,
     'd2:right': D2FunctionParameters,
     'd2:substring': D2FunctionParameters,
     'd2:split': D2FunctionParameters,
-    'd2:zScoreWFA': D2FunctionParameters,
     'd2:length': D2FunctionParameters,
+    // d2:inOrgUnitGroup(<orgunit_group_code>)
```
simonadomnisoru, line 293: Why did you comment these D2 functions ? Thanks!

src/core_modules/capture-core-utils/rulesEngine/rulesEngine.types.js
```diff
@@ -57,15 +57,16 @@ export type ProgramRuleAction = {
...
     trackedEntityAttributeId?: ?string,
+    programRuleVariableId?: ?string,
     optionGroupId: ?string,
     optionId: ?string,
     style?: ?Object,
 };
 
 export type ProgramRule = {
     id: string,
-    name: string;
```
simonadomnisoru, line 67: I wonder why didn't the linter detect these typos. Do you have any idea @superskip?

src/core_modules/capture-core/rules/converters/outputConverter.js
superskip: That particular function? Probably not from what I've heard, but the function needs a definition because of the interface that RulesValueConverter implements. The interface is defined such that each _value type_ a data element can have has one converter, and _Tracker Associate_ is one such value type.

src/core_modules/capture-core-utils/rulesEngine/RulesEngine.js
superskip: So you're saying it's a pattern to pass one object containing named parameters instead of passing a comma separated list? Then one would have to remember or check the names instead of the parameter order though. But that's probably easier in this case where there are more than four parameters and the ordering is quite arbitrary.

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/VariableService.js
superskip: I looked up the difference between `Object` and `any`, but apparently they are the same thing in flow. The NOTE in the last section on [this page](https://flow.org/en/docs/types/objects/) says quite explicitly that it's better to keep `any` if the alternative is to change it into Object, actually.

src/core_modules/capture-core-utils/rulesEngine/rulesEngine.types.js
superskip: I compared the list of D2-functions that can be selected in a program rule expression with the corresponding list in the rules engine, and found that the rules engine was lacking a few. I inserted the missing functions as comments for later reference.

src/core_modules/capture-core-utils/rulesEngine/rulesEngine.types.js
superskip: It must be lacking a detection rule for that sort of thing, can't imagine any other explanation.

src/core_modules/capture-core-utils/rulesEngine/RulesEngine.js
huocha: Yes, exactly!

JoakimSM (approved):
Looks very good @superskip

PR #1877, opened 2021-07-09
refactor: [TECH-580] ScopeSelector in enrollment page - part 1

simonadomnisoru:
I created a new component `ScopeSelector` and copied the files from `LockerSelector`. I used the ScopeSelector instead of the LockerSelector on the enrollment page. Nothing else changed. 

 I prefer to split the refactor into 2 parts to be able to differentiate between the logic that was copy-pasted from the  `LockerSelector` and the new refactored hooks. The real changes are implemented in https://github.com/dhis2/capture-app/pull/1878

### DIFF
--- a/src/core_modules/capture-core/components/ScopeSelector/QuickSelector/OrgUnitSelector.component.js
+++ b/src/core_modules/capture-core/components/ScopeSelector/QuickSelector/OrgUnitSelector.component.js
@@ -1,0 +1,185 @@
+// @flow
+import React, { Component } from 'react';
+import i18n from '@dhis2/d2-i18n';
+import { withStyles } from '@material-ui/core/styles';
+import Paper from '@material-ui/core/Paper';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogContentText from '@material-ui/core/DialogContentText';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import IconButton from '@material-ui/core/IconButton';
+import ClearIcon from '@material-ui/icons/Clear';
+import { Button } from '../../Buttons';
+import { OrgUnitField } from './OrgUnitField.container';
+
+const styles = (theme: Theme) => ({
+    paper: {
+        padding: 8,
+        backgroundColor: theme.palette.grey.lighter,
+    },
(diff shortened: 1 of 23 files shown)


src/core_modules/capture-core/components/ScopeSelector/QuickSelector/OrgUnitSelector.component.js
```diff
@@ -0,0 +1,185 @@
...
+    showWarning: boolean,
+    selectedOrgUnit: Object,
+    classes: Object,
+};
+
+type State = {
+    open: boolean,
+};
+
+class OrgUnitSelectorPlain extends Component<Props, State> {
```
huocha, line 85: Hey @simonadomnisoru , is there any reason why you use class component instead of function component? I think we are moving toward using function component.

src/core_modules/capture-core/components/ScopeSelector/ScopeSelector.container.js
```diff
@@ -0,0 +1,195 @@
...
+                    orgUnitId: routerOrgUnitId,
+                },
+                pathname,
+            },
+        },
+    }) =>
+        ({
+            selectedProgramId: routerProgramId || selectedProgramId,
+            selectedOrgUnitId: routerOrgUnitId || selectedOrgUnitId,
+            pathname,
```
huocha, line 52: should we add `shallowEqual` here? WDYT?

src/core_modules/capture-core/components/ScopeSelector/QuickSelector/OrgUnitSelector.component.js
simonadomnisoru: Hey @huocha,

I wrote in the PR description that the files in this PR are copy-pasted from the `LockerSelector` component. class component was used in these files I didn't change the logic.
Thanks!

src/core_modules/capture-core/components/ScopeSelector/ScopeSelector.container.js
simonadomnisoru: In this PR I just copied old logic inside a new folder/component. The real changes are implemented in https://github.com/dhis2/capture-app/pull/1878. There the `currentSelections` logic is removed. Thanks!

JoakimSM (approved):
LGTM

PR #1881, opened 2021-07-12
DHIS2-10330 viewEvent URL link for category options

EirikHaugstulen:
Small change to fix the bug where posting the link would break the capture app. 

https://jira.dhis2.org/browse/DHIS2-10330

### DIFF
--- a/src/core_modules/capture-core/components/Pages/ViewEvent/epics/getCategoriesDataFromEvent.js
+++ b/src/core_modules/capture-core/components/Pages/ViewEvent/epics/getCategoriesDataFromEvent.js
@@ -10,12 +10,12 @@ function getCategoryOptionsAsync(optionIds: string) {
             fields: 'id,displayName,categories~pluck,access',
             filter: `id:in:[${optionIds}]`,
         })
-        .then(response => response && response.categoryOptions);
+        .then(response => response?.categoryOptions);
 }
 
 
 export async function getCategoriesDataFromEventAsync(event: CaptureClientEvent): Promise<?Array<Object>> {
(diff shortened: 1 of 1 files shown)


src/core_modules/capture-core/components/Pages/ViewEvent/epics/getCategoriesDataFromEvent.js
```diff
@@ -10,12 +10,12 @@ function getCategoryOptionsAsync(optionIds: string) {
...
             filter: `id:in:[${optionIds}]`,
         })
-        .then(response => response && response.categoryOptions);
+        .then(response => response?.categoryOptions);
 }
 
 
 export async function getCategoriesDataFromEventAsync(event: CaptureClientEvent): Promise<?Array<Object>> {
-    const optionIdsFromEvent = event.attributeCategoryOptions;
+    const optionIdsFromEvent = event.attributeCategoryOptions?.replace(/;/g, ',');
```
simonadomnisoru, line 18: Hey @EirikHaugstulen, 

Do you want to replace all the occurrences of`;` with`,`?
I think that `;` needs to be escaped like this `.replace(/\;/g, ",")`. You can also use `replaceAll(";", ",")` and simplify the regex syntax. 🤔

EirikHaugstulen: Hey @simonadomnisoru! 
Thanks for reviewing. I also thought the `;` should be escaped, but ESLint does not agree.
Do you still want me to change the code?

image

simonadomnisoru (approved):

PR #1887, opened 2021-07-13
feat: [TECH-643] create custom hook for fetching and storing enrollment data to Redux

simonadomnisoru:
The `useEnrollment` hook fetch new enrollment data if: no enrollment data exists in the Redux store or the enrollment id from the Redux store doesn't match the enrollment id from the URL.


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.js
```diff
@@ -54,3 +56,6 @@ export const deleteEnrollment = ({ enrollmentId }: { enrollmentId: string }) =>
     actionCreator(enrollmentPageActionTypes.DELETE_ENROLLMENT)({
         enrollmentId,
     });
+
+export const saveEnrollment = (enrollmentSite: EnrollmentData) =>
```
huocha, line 60: hey @simonadomnisoru, I wonder if this `saveEnrollment` is used only once when fetching data for the first time?
Will it be confused with the save function after editing enrollment?

simonadomnisoru: Yes for now it is only used once. But in the next PRs, it may be used after a new event is added to the enrollment or after an existing event is edited. I agree there are many functions with similar names and it can be confusing. Do you have a better name suggestion that I can use? Thanks!

huocha: how about `setEnrollment` if it's used to set the enrollment data in the store?

huocha (approved):
LGTM!

PR #1907, opened 2021-07-19
feat: [DHIS2-11393] add event page - display form

huocha:


src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.component.js
```diff
@@ -40,30 +55,36 @@ const EnrollmentAddEventPagePain = ({
...
+                                        />
+                                    </div>
+                                )}
+                                <span> {stageForm.name} </span>
+                            </div>
+                        }
+                        noncollapsible
+                    >
+                        <div data-test="edit-event-form">
+                            <EditEventDataEntry formFoundation={programStage.stageForm} />
```
simonadomnisoru, line 80: Hey, 
The `EditEventDataEntry` is the form component used in the edit page. For the new event page the old components are `SingleEventRegistrationEntry` and `NewEventDataEntry`. I think you should use these ones, not the ones from the edit form. The changes in the `.actions`, `.epics`, and rules engine helpers should also happen for the add event components, not the edit ones. WHYT?
Thanks!

PR #1880, opened 2021-07-12
DHIS2-11388: create new routing for Enrollment Add Event

huocha:

PR #1908, opened 2021-07-19
feat: [DHIS2-11393] add event display form

huocha:
Reopen the PR displaying form using `SingleEventRegistrationEntry`

### DIFF
--- a/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
+++ b/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
@@ -1,5 +1,6 @@
 Feature: User interacts with Enrollment Add event page
   Scenario: The user can land on the enrollment add event page.
-    Given you land on the enrollment event page by having typed #/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg&teiId=EaOyKGOIGRp&enrollmentId=IMEGtAd4DOj&stageId=ZzYYXq4fJie&eventId=vDGrlVeyE2Q
+    Given you land on the enrollment add event page by having typed /#/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg8&teiId=tIJu6iqQxNV&enrollmentId=CCBLMntFuzb&stageId=A03MvHHogjR&eventId=V1CerIi3sdL
     Then you see the following Enrollment: New Event
-    And you see the widget header Baby Postnatal
+    And you see the widget header Birth
+    And you see the add event form details
--- a/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
+++ b/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
@@ -1,6 +1,6 @@
 Feature: User interacts with Enrollment Add event page
   Scenario: The user can land on the enrollment add event page.
-    Given you land on the enrollment add event page by having typed /#/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg8&teiId=tIJu6iqQxNV&enrollmentId=CCBLMntFuzb&stageId=A03MvHHogjR&eventId=V1CerIi3sdL
+    Given you land on the enrollment add event page by having typed /#/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg8&teiId=tIJu6iqQxNV&enrollmentId=CCBLMntFuzb&stageId=A03MvHHogjR
     Then you see the following Enrollment: New Event
     And you see the widget header Birth
     And you see the add event form details
--- a/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation/index.js
+++ b/cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation/index.js
@@ -2,7 +2,7 @@ beforeEach(() => {
     cy.loginThroughForm();
 });
 
-Given(/^you land on the enrollment event page by having typed (.*)$/, (url) => {
+Given(/^you land on the enrollment add event page by having typed (.*)$/, (url) => {
     cy.visit(url);
 });
 
@@ -16,5 +16,12 @@ And(/^you see the widget header (.*)$/, (name) => {
             cy.get('[data-test="widget-contents"]').should('exist');
             cy.get('[data-test="widget-header"]').should('exist');
             cy.contains(name).should('exist');
+        });
+});
+
+And('you see the add event form details', () => {
+    cy.get('[data-test="add-event-enrollment-page-content"]')
+        .within(() => {
+            cy.get('[data-test="edit-event-form"]').should('exist');
         });
 });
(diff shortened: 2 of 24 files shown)


src/core_modules/capture-core/metaData/helpers/EventProgram/getEventProgramEventAccess.js
```diff
@@ -1,12 +1,25 @@
 // @flow
-import { getEventProgramThrowIfNotFound } from './getEventProgramThrowIfNotFound';
+import log from 'loglevel';
+import { errorCreator } from 'capture-core-utils';
+import { getProgramFromProgramIdThrowIfNotFound } from '../getProgramFromProgramIdThrowIfNotFound';
+import { EventProgram } from '../../Program';
 
 export function getEventProgramEventAccess(
```
simonadomnisoru, line 7: Hey, 
If you make `getEventProgramEventAccess` support both programs types then I recommend renaming the file and the function to something more suitable like `getProgramEventAccess`. Moreover, I think that the file should be moved outside the `EventProgram` folder into a more generic location. WDYT?
Thanks!

cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation/index.js
```diff
@@ -18,3 +18,10 @@ And(/^you see the widget header (.*)$/, (name) => {
             cy.contains(name).should('exist');
         });
 });
+
+And('you see the add event form details', () => {
+    cy.get('[data-test="add-event-enrollment-page-content"]')
+        .within(() => {
+            cy.get('[data-test="edit-event-form"]').should('exist');
```
simonadomnisoru, line 25: Hey, 
Can you add more strict checks for this feature? I think that besides the form id we can also check that some specific elements inside the form are being displayed.
Thank you!

cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
```diff
@@ -1,5 +1,6 @@
 Feature: User interacts with Enrollment Add event page
   Scenario: The user can land on the enrollment add event page.
-    Given you land on the enrollment event page by having typed #/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg&teiId=EaOyKGOIGRp&enrollmentId=IMEGtAd4DOj&stageId=ZzYYXq4fJie&eventId=vDGrlVeyE2Q
+    Given you land on the enrollment add event page by having typed /#/enrollmentEventNew?programId=IpHINAT79UW&orgUnitId=DiszpKrYNg8&teiId=tIJu6iqQxNV&enrollmentId=CCBLMntFuzb&stageId=A03MvHHogjR&eventId=V1CerIi3sdL
```
simonadomnisoru, line 3: Hey, 

I just noticed you have the `eventId` in the URL. When we create a new event, we don't know the `eventId` yet therefore it should not be part of the page URLs parameters. You can check the correct pattern in the [design docs](https://docs.google.com/document/d/1I9-xc1oA95cWb64MHmIJXTHXQnxzi1SJ3RUmiuzSh78/edit#heading=h.r4doz564mcx5). I am sorry I didn't notice this in your previous PR https://github.com/dhis2/capture-app/pull/1880 😅
Thank you!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.component.js
```diff
@@ -60,8 +73,26 @@ const EnrollmentAddEventPagePain = ({
                     }
                     noncollapsible
                 >
-                    [event details]
+                    <div data-test="add-event-form">
```
simonadomnisoru, line 76: Hey, 
Here, I think you should use `Widget` and moved all the form-related logic inside it. The same pattern is used on the [Enrollment](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js#L53) page. The page only handles the context and passes it to different Widgets. You can also look at `WidgetEventEdit` for some ideas. 
Thanks!

src/core_modules/capture-core/components/Pages/ViewEvent/ViewEventComponent/viewEvent.actions.js
```diff
@@ -22,7 +22,7 @@ export const actionTypes = {
 
 export const viewEventFromUrl = (data: Object) =>
     actionCreator(actionTypes.VIEW_EVENT_FROM_URL)({
-        eventId: data.nextProps.viewEventId,
+        eventId: data.nextProps.viewEventId || data.nextProps.eventId,
```
simonadomnisoru, line 25: Hey, 
I don't think this change is needed anymore as the URL will not contain the parameter `eventId`. 
Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.component.js
huocha: hey @simonadomnisoru, thank you for the comment, I moved the form to `WidgetAddEvent`

cypress/integration/EnrollmentAddEventPage/EnrollmentAddEventPageNavigation.feature
simonadomnisoru: Hey, 
Everything looks good now! I just have one last comment. I think that `eventId` should also be removed from the Cypress scenario URL. 
Thank you for your patience with all my suggestions! 😁

huocha:
@simonadomnisoru  thank you for all comments, I made the adjustments based on your suggestions.

src/core_modules/capture-core/components/WidgetAddEvent/WidgetAddEvent.component.js
```diff
@@ -0,0 +1,81 @@
...
+                            />
+                        </div>
+                    )}
+                    <span> {stageForm.name} </span>
+                </div>
+            }
+
+        >
+            <div data-test="add-event-form">
+                {scopeType === scopeTypes.TRACKER_PROGRAM && <EnrollmentRegistrationEntry
```
simonadomnisoru, line 58: Hey, 
I am not sure `EnrollmentRegistrationEntry` is the component to use here. `EnrollmentRegistrationEntry` is used to register a new **enrollment** and this feature is about a new **event**. 
I think you should use `SingleEventRegistrationEntry` and `NewEventDataEntry` (or something similar from the single event page) and make them work for an enrollment's event (tracker program).  
In the Cypress tests, you should check for the **event** inputs fields just like you did a few comments back https://github.com/dhis2/capture-app/pull/1908/commits/6b44aee44b09c0398638b842613e28305a6691b4 . 
You can check how this feature works in the old Tracker capture app.  

Screenshot 2021-07-20 at 16 01 34

Screenshot 2021-07-20 at 16 03 26

Let me know if this makes sense. Thank you! 😄

huocha: sorry I was confused by the fact that there isn't possible to create any new event on the main page, only enrolment as you said. I will make the `SingleEventRegistrationEntry` usable for both Tracker and Event program.

PR #1910, opened 2021-07-20
DHIS2 - Widgets in Add Event Page

EirikHaugstulen:
This is a shared PR for the following Jira-issues:
[DHIS2-11448](https://jira.dhis2.org/browse/DHIS2-11448)
[DHIS2-11449](https://jira.dhis2.org/browse/DHIS2-11449)
[DHIS2-11450](https://jira.dhis2.org/browse/DHIS2-11450)
[DHIS2-11452](https://jira.dhis2.org/browse/DHIS2-11452)
[DHIS2-11455](https://jira.dhis2.org/browse/DHIS2-11455)
[DHIS2-11456](https://jira.dhis2.org/browse/DHIS2-11456)

This introduces the different widgets to the `EventNew`-page and connects it to the redux store.

It is also branched out of both https://github.com/dhis2/capture-app/pull/1915 & https://github.com/dhis2/capture-app/pull/1923, so these needs to be merged.


simonadomnisoru:
Hey @EirikHaugstulen, 

Great job on this PR!
Can you add Cypress tests for all the new Widgets? You should be able to reuse most of the features inside the `EnrollmentPage` folder.
We will probably need the same tests for these widgets in the `EnrollmentEditEvent` page. Can you investigate if it makes sense to use `Scenario Outline` and move the widget's features outside the `EnrollmentPage` to a more generic location?
Thank you!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,28 +1,54 @@
...
         }) => ({
             programId: query.programId,
             stageId: query.stageId,
+            teiId: query.teiId,
+            enrollmentId: query.enrollmentId,
+            orgUnitId: query.orgUnitId,
         }),
         shallowEqual,
     );
+    const outputEffects = useWidgetDataFromStore('singleEvent-newEvent');
```
simonadomnisoru, line 29: I am curious if this new hook `useWidgetDataFromStore` can replace `useFilteredWidgetData` in the [Enrollment](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js#L45) page? Thanks!

EirikHaugstulen: Hey @simonadomnisoru,
The `useWidgetDataFromStore`-hook fetches the messages from the redux-store, while in the EnrollmentPageDefault, the messages are stored in an Object directly from the rulesEngine. The `useFilteredWidgetData` is then used to extract the correct messages and fiilter them. This way, one of the hooks fetches and the other filters. 
Do you still think they could be the same? Let me know if something is unclear or if I have misunderstood your question ☺️

simonadomnisoru: Now I see the difference. Thank you for explaining it! 😄 
This approach assumes that the messages are always stored. I see some edge cases to this strategy. For example, what happens when the user goes directly to the Add Event page via a URL link. Or if the user clears the cache (redux store) and refreshes the page (they usually do it to get the latest metadata)? 🤔
Do you think is a good idea to check that the messages exist in the redux-store before filtering them? And if they don't exist should `useWidgetDataFromStore` be used to fetch them?
Thanks!

EirikHaugstulen: Yes, I understand.
The `NewEvent`-page is going to use the same component as used in the `/new`-page. This page dispatches an action called `OpenNewEventInDataEntryActionsBatch`. This is the action that prepares the redux-store for the effects, and is also the one that imports the initial effects after the rulesEngine is run on load. Since these widgets will run side-by-side with this component, the initial effects should be run simultaneously, even if the user goes directly through the URL. If not, the hook will return the object with empty arrays and the widgets will not show. (Except for feedback & indicators which will show a "No feedback/indicator for this event yet"-message. The profile and enrollment will still work as normal. In the `/EventEdit`-page I am a bit more uncertain. I see that in the `/new`-page, there is an action called `BATCHING_REDUCER.BATCH` with a `UpdateRulesEffects`-payload that does the same thing. Will this work in the same way as before @simonadomnisoru? ☺️

@huocha Feel free to correct me if I'm wrong about something!

simonadomnisoru: Hey, 
I am yet not very familiar with how the rules will work on the `EnrolmentEditEvent` page. 
Today I started looking at this ticket https://jira.dhis2.org/browse/TECH-632. After it is implemented, the rules engine (for a tracker program) will be is called with different data than today. The `EnrollmentAddEvent` page has a similar ticket inside its epic https://jira.dhis2.org/browse/TECH-635. 
After I dig more into this I will have more feedback. Thank you!

EirikHaugstulen: Okay, I understand! ☺️ I`ll put this issue on ice for a bit until it is decided how the rules will be run. Thanks for all comments @simonadomnisoru!

simonadomnisoru: Hey @EirikHaugstulen,

I spent the last couple of days investigating the rules engine. From my findings, the logic of when (on load) and where (in the redux store object) we save the initial effects of the rules will not change (for the `EnrolmentEditEvent` at least). Therefore all my doubts about this PR are clarified.

Thank you!

huocha:
hey @EirikHaugstulen, I updated the branch Add event to fix the issue of duplication of some widgets reported [here](https://github.com/dhis2/capture-app/pull/1915#issuecomment-891691350). If you merge my branch into your branch it will be great.

simonadomnisoru (approved):

JoakimSM (approved):

PR #1917, opened 2021-07-21
feat: [TECH-633] Update event in redux after edit

simonadomnisoru:
- Moved the `useEnrollment` hook, actions, and epics from the `Enrollment` Page to a more generic location. 
- Added commit and rollback actions to the save event form.


JoakimSM (approved):
Looks good @simonadomnisoru!

PR #1923, opened 2021-07-26
DHIS2-11399

huocha:
Implement `Complete`, `Save without completing` buttons on Add new event form
- Modify the text only in Add new event form (not the main page)
- I have troubles using existing`withCompleteButton` since it requires the D2Form instance but can not get the ref from another HOC, so I add directly the button in `withMainButton`


src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/getConvertedNewSingleEvent.js
```diff
@@ -43,6 +43,44 @@ export const getNewEventServerData = (state: ReduxState, formFoundation: RenderF
...
+    }
+    if (mainDataServerValues.status === 'COMPLETED') {
+        mainDataServerValues.completedDate = getFormattedStringFromMomentUsingEuropeanGlyphs(moment());
+    }
+
+    return {
+        events: [
+            {
+                ...mainDataServerValues,
+                program: programId,
```
simonadomnisoru, line 67: Hey, 
There is a helper named [getServerValuesToSaveFromMainEvent](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/DataEntry/common/convertEventToServerValues.js#L5) inside `DataEntry` component. Do you think it can be reused to map the keys from client to server shape?

Thanks!

simonadomnisoru:
Hey @huocha, 
How do you plan to test this feature? 

Currently, in the Tracker programs table, we are missing the delete functionality. To ensure that the Cypress test cleans the database, I think we need the delete functionality before being able to properly test the 'save new event' feature. 
Despite this limitation, you should be able to test at least that the program rules are being triggered properly without saving the new event. 
Thanks!

src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/getConvertedNewSingleEvent.js
huocha: I don't want to mess up with the main event so I made a separate function

PR #1928, opened 2021-07-29
feat:  [TECH-632] Invoked the rules engine with the correct enrollment events data

simonadomnisoru:
Invoked the rules engine with data from all events for the current enrollment and TEI stored in Redux


JoakimSM (approved):
Looks really good 👍

PR #1930, opened 2021-07-30
refactor: [TECH-605] CircularProgress migrated to CircularLoader

simonadomnisoru:


src/core_modules/capture-core/components/LoadingMasks/LoadingMaskForButton.component.js
```diff
@@ -1,15 +1,7 @@
...
-
-};
-
-export const LoadingMaskForButton = (props: Props) => (
-    // $FlowFixMe[cannot-spread-inexact] automated comment
-    <LoadingMask
-        size={24}
-        {...props}
-    />
+export const LoadingMaskForButton = () => (
```
huocha, line 5: NIT: Since it's no longer `LoadingMask` do you think we should rename to `CircularLoaderForButton`

src/core_modules/capture-core/components/LoadingMasks/LoadingMaskForPage.component.js
```diff
@@ -22,9 +22,7 @@ const LoadingMaskForPagePlain = (props: Props) => (
     <div
         className={props.classes.container}
     >
-        <LoadingMask
-            size={50}
-        />
+        <CircularLoader />
```
huocha, line 25: NIT: Since it's no longer `LoadingMask` do you think we should rename to `CircularLoaderForPage`

src/core_modules/capture-core/components/LoadingMasks/LoadingMaskForButton.component.js
simonadomnisoru: Hey,

The components in the folder `LoadingMasks` use`CircularLoader` + different other elements in their `JSX`. That is why I prefer the term `LoadingMasks...` as a generic namespace for whatever we want to render,  not only the `CircularLoader`. 

Thank you for your feedback!

src/core_modules/capture-core/components/LoadingMasks/LoadingMaskForPage.component.js
simonadomnisoru: Same explanation as the one in the `LoadingMaskForButton.component.js` file

simonadomnisoru:
I split the big migration ticket https://jira.dhis2.org/browse/TECH-605 into smaller tasks for each one of the components. The changes from this PR were moved to https://github.com/dhis2/capture-app/pull/1931

PR #1941, opened 2021-08-05
chore(deps): update app-runtime to latest

ismay:
https://jira.dhis2.org/browse/LIBS-200. Updating to the new app-runtime release (currently in pre-release, don't merge until this is using a stable version).

- [ ] Before this can be updated the breaking changes in cli-app-scripts should be addressed (i.e. dhis2/ui should be updated). Until then this can't be merged
- [ ] Remove resolutions field once upgraded to stable

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,11 @@
   "cacheVersion": "1",
   "serverVersion": "37",
   "private": false,
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


JoakimSM:
Superseded by https://github.com/dhis2/capture-app/pull/2227

PR #1944, opened 2021-08-09
feat: [DHIS2-11572] implement navigation from overview page

huocha:


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/StageDetail.component.js
```diff
@@ -65,6 +82,18 @@ const StageDetailPlain = ({ events, eventName, dataElements, classes }: Props) =
         });
     };
 
+    const handleViewAll = () => {
+        history.push(
+            `/enrollment/stageEvents?${urlArguments({ orgUnitId, programId, stageId })}`,
```
simonadomnisoru, line 87: Hey @huocha, 

I think both navigations handlers should be moved at the Enrollment page level.  
The `WidgetsStagesAndEvents` can expose callbacks when the user clicks the `New {{ eventName }} event` and `Go to full {{ eventName }}` buttons. Using this strategy `WidgetsStagesAndEvents` is decoupled from the `Enrollment` page routing logic. The widget can be easier reused in other pages that will handle their own navigation. 

What are your thoughts on this? Thank you!

huocha: hey @simonadomnisoru, I don't think it's possible to handle the navigation from the enrollment page since it needs the `stageId` which only specified in the `WidgetStagesAndEvents`

simonadomnisoru: The `stageId` can be passed as a parameter to the callback function exposed from `WidgetStagesAndEvents` and used on the page. 


1) In the `Enrollment` page:

`const handleOnClick = (stageid) => history.push(/enrollment/stageEvents?${urlArguments({ orgUnitId, programId, stageId })}`
....
`<WidgetStagesAndEvents callbackOnClick={callbackOnClick} .... /> `


2) In the WidgetStagesAndEvents/Stages/Stage/StageDetail/StageDetail.component.js the callback will be used like this.
```
  const handleViewAll = () => {
        const {callbackOnClick} = props;
        callbackOnClick(stageId)
}
```

What do you think?

huocha: I can not imagine these buttons to behave differently as their labels said "Go to this page" or "Create new event", it also attached to the widget with the `stageId` so I don't think pass the callback 3 levels above worth it. 
But I will let the team decides if we should do it or not
@JoakimSM @EirikHaugstulen WDYT?

huocha: hey @simonadomnisoru, after discussing with @JoakimSM we agree to handle it with the callback.
I will change the implementation.

simonadomnisoru (approved):
LGTM!

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
```diff
@@ -72,6 +72,17 @@ export const EnrollmentPageDefault = () => {
...
+    const onViewAll = (stageId) => {
+        history.push(
+            `/enrollment/stageEvents?${urlArguments({ orgUnitId, programId, stageId })}`);
+    };
+
+    const onCreateNew = (stageId) => {
+        history.push(
+            `/enrollmentEventNew?${urlArguments({ orgUnitId, programId, teiId, enrollmentId, stageId })}`,
+        );
+    };
```
JoakimSM, line 84: For performance reasons, I would recommend that we make it a habit to wrap callbacks in `useCallback`.

JoakimSM (approved):
Looks good. One minor comment

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
huocha: do you mean we should add `useCallback` both here and in the [Stage file](https://github.com/dhis2/capture-app/pull/1944/files#diff-aa98c87a1bdf7b1b522ffd413b516c5423e36342210531a0d0e054a6c0df8789R73)?

JoakimSM: Yes, that makes sense to me

PR #1959, opened 2021-08-10
fix: [TECH-643] useEffect to keep the sideEffects out of the render method

simonadomnisoru:
Address feedback for previous commit https://github.com/dhis2/capture-app/commit/27434937ed9f3bab8bb803c4093c07c02b686238#diff-d08ab005f39f6c780009bc1709e65f0a06df17d340a8a4651cc953ca1e2b2e4bR48-R50


JoakimSM (approved):
LGTM

PR #1970, opened 2021-08-12
feat: [TECH-620]: sort rules engine events by date

superskip:
Make sure the event arrays that come in to the rules engine are sorted in ascending order wrt. registration date.


JoakimSM (approved):
LGTM 👍

PR #1987, opened 2021-08-19
fix: [DHIS2-11084] Refactor search form value converters of TEA

EirikHaugstulen:
Refactor the search form value conversion on the search for TEI page. 
Many of the data element types used as a Tracked Entity Attribute in the search form is not converted to the values that the api accepts. 
This does not fix every data type, especially because some is not supported by the API. (Coordinates, polygon, age and so on).


simonadomnisoru (approved):
LGTM!

j-nguyenn (approved):
👍

jasminenguyennn (approved):

JoakimSM (approved):
Thanks @EirikHaugstulen, looks good.

Can you also do a back port of this to v37?

PR #1992, opened 2021-08-20
refactor: [TECH-659] move Top bar actions from Scope Selector

simonadomnisoru:
- Create a new component with the top bar actions logic.
- Use it on the `Enrollment` page
- Clean the code related with the actions form the `ScopeSelector`


j-nguyenn (commented):
Thank you for your great work!

src/core_modules/capture-core/components/TopBarActions/TopBarActions.container.js
```diff
@@ -0,0 +1,108 @@
...
+    selectedProgramId,
+    selectedOrgUnitId,
+    isUserInteractionInProgress = false,
+    customActionsOnProgramIdReset = [],
+}: Props) => {
+    const [openStartAgainWarning, setOpenStartAgainWarning] = useState(false);
+    const dispatch = useDispatch();
+    const { reset } = useReset();
+    const { setOrgUnitId } = useSetOrgUnitId();
+    const dontShowWarning = () => !isUserInteractionInProgress;
```
j-nguyenn, line 33: hey @simonadomnisoru, why do you invoke a function here? 

nit: I see you only refactor file without changing it, but will we have plan to clean up this later? I think a good pattern for the naming here would be `shouldNotShowWarning`? wdyt?

j-nguyenn (commented):
Thank you for your great work!

src/core_modules/capture-core/components/TopBarActions/TopBarActions.container.js
simonadomnisoru: Hey, 
I decided to use the `isUserInteractionInProgress` prop directly without storing it in any new variable. Let me know wdyt. 
Thanks!

j-nguyenn (approved):
👍

JoakimSM (approved):
LGTM 👍

PR #2009, opened 2021-08-24
feat: [DHIS2-11576] improve sorting function

j-nguyenn:
### Feature
 Improve sorting functions
- Sort by status desc by default
- Secondarily sort by eventDate or dueDate
- Empty values will be placed in the bottom


Bekkalizer (commented):
Tested now and this seems to work very well. There is only one very small thing I found when trying the different sorting out. The default sort for the list works very well on almost all columns: The first click on a column will make the sorting descending, then the second click will make the sort descending, and the third click will go to default sorting - which means that "Status" will be sorted ascending(nice).
The "Status" column works the same way, but the effect might be a little bit unexpected. When already in the default sorting state("Status" ascending), and then clicking the Status sorting button, the filter is deselected, and you get to the default state again. This might be slightly strange, as you where in that state already("Status" sorted ascending both before and after the click). It would have been slightly better if this first column only has the two options, ascending and descending, and the "blank"/default state was not triggered when clicking the Status column specifically. 

----
Edit: This would require a bigger change in the component, and we will not do anything with it in this PR.

Bekkalizer (approved):

PR #2010, opened 2021-08-24
feat: [DHIS2-11631] Prevent adding event if stage is not repeatable and there is already event

j-nguyenn:
In the stages and events Widget, prevent the user from adding more than one event when the programstage is not repeatable
- disable the button
- on hover, show a ui tooltip with content: This stage can only have one event


simonadomnisoru (approved):
LGTM Great job!

Bekkalizer (approved):
LGTM!

PR #2014, opened 2021-08-25
feat: [DHIS2-11652] show the add button for empty event stage

j-nguyenn:
### Feature
In case there is no event in the list, there should be a button to add new event. Clicking this button will navigate to `Add new event` page

Screenshot 2021-08-25 at 07 39 53


Bekkalizer (approved):
LGTM!

PR #2022, opened 2021-08-26
feat: [DHIS2-11199][DHIS2-11389] add ScopeSelector in EnrollemntEventEdit page & EnrollemntEventNew page

simonadomnisoru:
- And the `ScopeSelector` in the `EnrollmpentEventEdit` page. If the user tries to leave the page, the warning pop up is triggered if the page is in edit mode
- And the `ScopeSelector` in the `EnrollmpentEventNew` page. If the user tries to leave the page after it has interacted with the form, the warning pop-up is triggered.


src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.js
```diff
@@ -46,7 +47,14 @@ export const EnrollmentEditEventPage = () => {
         dispatch(deleteEnrollment({ enrollmentId }));
     };
     const onGoBack = () => history.push(`/enrollment?${urlArguments({ orgUnitId, programId, teiId, enrollmentId })}`);
-    useEnrollment(teiId);
+    const enrollmentSite = useEnrollment(teiId).enrollment;
+    const { teiDisplayName } = useTeiDisplayName(teiId, programId);
+    const { trackedEntityName } = getScopeInfo(enrollmentSite.trackedEntityType);
+    const enrollmentsAsOptions = buildEnrollmentsAsOptions([enrollmentSite], programId);
+    const event = enrollmentSite?.events?.find(item => item.event === eventId);
```
j-nguyenn, line 54: hey @simonadomnisoru how do we handle the case when `enrollmentSite` or `events` is undefined?

simonadomnisoru: Hey, 
Nice catch! I added checks for when the data is `undefined` in the [useEffect](https://github.com/dhis2/capture-app/pull/2022/files#diff-eeac4f86e764871045b77688d009436440c3f98f09af8a1c2b7d18b44d642105R58) bellow. 
Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.component.js
```diff
@@ -47,60 +59,137 @@ const EnrollmentEditEventPagePain = ({
...
+                        ready={pageStatus !== pageStatuses.MISSING_DATA}
+                        onClear={() => resetEventId('enrollment')}
+                        options={[
+                            {
+                                label: eventDate || '',
+                                value: 'alwaysPreselected',
+                            },
+                        ]}
+                        selectedValue="alwaysPreselected"
+                        title={i18n.t('date of visit')}
```
JoakimSM, line 145: (Sorry for repeating myself)

We have some metadata that we can use for this title. It can be found here: programStage.stageForm.getLabel('eventDate')

(and yes we need to refactor where it resides at some point)

JoakimSM (commented):
Great work on this @simonadomnisoru 👍 

One minor comment. I also have a comment about the pageStatus, but I think it's better to have a quick chat about that one 🙂  I will ping you.

JoakimSM (approved):
Perfect!

PR #2024, opened 2021-08-27
feat: [DHIS2-11630] visualize due date in status tag

j-nguyenn:
### Feature
In the stages and events Widget we should visualise the due date in the status tag for scheduled events.

- `Overdue: due xxx ago`
- `Scheduled: due DDD` or `Scheduled: due xxx ago`

### Tests
can not add cypress test as it requires changing db for now. 

Screenshot 2021-08-26 at 12 03 15


Bekkalizer:
Looks nice! As we discussed on slack, it would be great if we can only show whole days of difference - and avoid "due in X hours" or "overdue by Y hours". An event that has due date on 2021-08-31 should ideally be status "Scheduled" for that entire day, and only become "overdue by 1 day" on the next day, 2021-09-01.

Bekkalizer (approved):
I cant test right now but the changes looks sensible!

PR #2033, opened 2021-08-30
feat: [DHIS2-11467] List records in program regardless of OrgUnit

EirikHaugstulen:
Working list with events and enrollments in a program regardless of selecting OrgUnit. This list will be shown on the MainPage when program is selected without ou.

Jira: https://jira.dhis2.org/browse/DHIS2-11467


Bekkalizer (commented):
Tested and it seems to work very well! Nice! 

There was one missing functionality when it comes to navigating to the old tracker capture with a returnUrl parameter. It seems that the &all part of the URL is not appended to the returnUrl that is submitted to tracker capture, so when opening an enrollment and then navigating back, the url is not including &all and the user has to click to reopen the list they came from when they left Capture.

JoakimSM (approved):

PR #2062, opened 2021-09-06
fix: [DHIS2-11669][DHIS2-11746] save new enrollment event validation & statuses

simonadomnisoru:
- Trigger the error popup when the user clicks the "Complete" button in the new enrollment event form. [DHIS2-11669](https://jira.dhis2.org/browse/DHIS2-11669). 
- send the correct event status when the user clicks the "Save without competing"

### DIFF
--- a/src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
+++ b/src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
@@ -14,6 +14,7 @@ import { getDataEntryKey } from '../common/getDataEntryKey';
 import { type RenderFoundation } from '../../../metaData';
 import { MessagesDialogContents } from './MessagesDialogContents';
 import { makeGetWarnings, makeGetErrors } from './withSaveHandler.selectors';
+import { addEventSaveTypes } from '../../WidgetEnrollmentEventNew/DataEntry/addEventSaveTypes';
 
 type Props = {
     classes: Object,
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
```diff
@@ -184,18 +185,17 @@ const getSaveHandler = (
...
-            const { onSave, itemId, id, calculatedFoundation } = this.props;
-            onSave(
-                itemId,
-                id,
-                calculatedFoundation,
-                saveType,
-            );
+            const { onSave, itemId, id, calculatedFoundation, warnings, errors } = this.props;
+            if (saveType === addEventSaveTypes.COMPLETE && ((errors && errors.length > 0) || (warnings && warnings.length > 0))) {
+                this.showMessagesPopup(saveType);
```
jasminenguyennn, line 195: hi @simonadomnisoru  Why do we only show the message when the save type is COMPLETE?
should it show if there is error and warning even when save type is SAVE_WITHOUT_COMPLETING?

simonadomnisoru: Hey, 
If the user clicks the button "Save without completing", we allowed for the event to be saved without validating the fields. The error popup will be shown only when the user clicks the "Complete" button. [DHIS2-11669](https://jira.dhis2.org/browse/DHIS2-11669) 
Thanks!

src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
```diff
@@ -184,18 +185,18 @@ const getSaveHandler = (
...
 
         handleSave = (saveType?: ?string) => {
-            const { onSave, itemId, id, calculatedFoundation } = this.props;
-            onSave(
-                itemId,
-                id,
-                calculatedFoundation,
-                saveType,
-            );
+            console.log('handleSave', saveType);
```
jasminenguyennn, line 193: is there any reason why we should keep this console?

simonadomnisoru: Good catch! I forgot to remove it. I updated the code. Thank you

jasminenguyennn (approved):

PR #2067, opened 2021-09-07
feat: [DHIS2-11616] Widget Event Comment

jasminenguyennn:
### Feature
Implement Widget comment in Event page  
- `WidgetComment` is a stand alone component takes `notes` and `onAddNote` as parameters
- `WidgetEventComment` for event edit page (view/edit mode)
Screenshot 2021-09-08 at 13 10 38
Screenshot 2021-09-08 at 13 10 43


src/core_modules/capture-core/components/WidgetEventComment/WidgetEventComment.epics.js
```diff
@@ -0,0 +1,76 @@
...
+    removeEventNote,
+} from '../Pages/ViewEvent/ViewEventComponent/editEvent.actions';
+
+import {
+    addNote,
+    removeNote,
+} from '../DataEntry/actions/dataEntry.actions';
+import { getCurrentUser } from '../../d2/d2Instance';
+
+export const addNoteForEventEpic = (action$: InputObservable, store: ReduxStore) =>
```
simonadomnisoru, line 22: Hey, 
I think we can take advantage of this refactor of `Comment` Widget and get rid of the `epics`, `actions`, and `reducers`. You can use instead the [DHIS2 application runtime hooks](https://runtime.dhis2.nu/#/hooks/). WHDY?
Thanks!

jasminenguyennn: hey @simonadomnisoru I don't know that we have plan to get rid of the epics and actions, they are still in used when get from the widget

Bekkalizer:
We should try to stick to "Comment", and eliminate the word "Note".

src/core_modules/capture-core/components/WidgetComment/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,164 @@
...
+            </div>
+
+            {isEditing && <div className={classes.newCommentButtonContainer} data-test="comment-buttons-container">
+                <Button
+                    dataTest="add-note-btn"
+                    onClick={onAddNote}
+                    className={classes.addCommentContainer}
+                    primary
+                >
+                    {i18n.t('Save note')}
```
Bekkalizer, line 152: Lets change this to 'Save comment'

i18n/en.pot
```diff
@@ -851,6 +851,18 @@ msgstr ""
 msgid "open the Tracker Capture app"
 msgstr ""
 
+msgid "This event doesn't have any notes yet"
```
Bekkalizer, line 854: Lets switch the wording to "comments" instead of "notes"

src/core_modules/capture-core/components/WidgetComment/CommentSection/CommentSection.js
jasminenguyennn: Sure will we update the design too. 
I think the `notes` still used in many place design here https://www.sketch.com/s/95c3d586-218c-4685-968b-a388270da949/a/R1ApWrl

Bekkalizer: Yes there seems to be some inconsistencies with the wording also in the designs. Great if we can make sure the texts we use right now is "comment", then we can circle back and make sure all the docs is in line with this :)

simonadomnisoru (approved):

Bekkalizer (approved):

PR #2068, opened 2021-09-07
test: d2-app-scripts not found

simonadomnisoru:
Sometimes the UT job runs before the install job and throws the error `d2-app-scripts not found`.  For example:

- https://github.com/dhis2/capture-app/runs/3525428607
- https://github.com/dhis2/capture-app/runs/3526136145
- https://github.com/dhis2/capture-app/runs/3531656562

I moved the UT job to ensure that all the scripts are correctly installed.


jasminenguyennn (approved):
👍

PR #2071, opened 2021-09-08
test: d2-app-scripts not found

simonadomnisoru:
Add the `install` job as a dependency for the `unit-test` job


JoakimSM (approved):
Thanks for looking into this 👍

PR #2074, opened 2021-09-08
DHIS2- Change reducer name in widgets-hook

EirikHaugstulen:
Small change to make widgets in add event page work.


src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -55,7 +55,7 @@ export const EnrollmentAddEventPage = () => {
     // Ticket: https://jira.dhis2.org/browse/TECH-669
     const { program } = useProgramInfo(programId);
     const programStage = [...program.stages.values()].find(item => item.id === stageId);
-    const outputEffects = useWidgetDataFromStore('singleEvent-addEvent');
+    const outputEffects = useWidgetDataFromStore('enrollmentEvent-newEvent');
```
jasminenguyennn, line 58: I believe we also have to change the name here 
maybe we can make it a constant or pass down from the page to keep the name in sync? 
https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/AccessVerification/AccessVerification.container.js#L22

simonadomnisoru (approved):

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
EirikHaugstulen: Thanks, @jasminenguyennn 👍

jasminenguyennn (approved):

PR #2077, opened 2021-09-09
chore: [DHIS2-11576] handle visited status

jasminenguyennn:


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/hooks/helpers.js
```diff
@@ -20,6 +20,10 @@ const getEventStatus = (event: ApiTEIEvent) => {
     if (isEventOverdue(event)) {
         return { status: statusTypes.OVERDUE, options: daysUntilDueDate ? dueDateFromNow : undefined };
     }
+    // DHIS2-11576: VISITED status is treated as ACTIVE
+    if (event.status === 'VISITED') {
```
simonadomnisoru, line 24: Hey, does it make sense to move the `VISITED` string inside the `statusTypes` const as `statusTypes.VISITED`?

jasminenguyennn: hey @simonadomnisoru we supposed to remove the status `VISITED` so I think it's better not to include inside statusTypes

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/hooks/helpers.js
```diff
@@ -50,7 +54,7 @@ const convertStatusForView = (event: ApiTEIEvent) => {
     return {
         isNegative,
         isPositive,
-        text: translatedStatusTypes(options)[status],
+        text: translatedStatusTypes(options)[status] ?? event.status,
```
JoakimSM, line 57: I think we can safely remove the `?? event.status`? Also, I see that we now have two files called statusTypes (possibly due to some merge conflicts). Can we use the file in the events folder instead and delete the one from the DataElements folder?

j-nguyenn: thanks @JoakimSM. I wanted to add `?? event.status` in case we have any unhandled statuses but if we are sure then it's not necessary. will remove it

jasminenguyennn: thanks @JoakimSM. I wanted to add `?? event.status` in case we have any unhandled statuses but if we are sure then it's not necessary. will remove it

simonadomnisoru (approved):

PR #2083, opened 2021-09-10
fix: [DHIS-11745][TECH-634] Update the enrollments collection when saving new event

simonadomnisoru:
- When the save request starts we update the enrollments collection In Redux. Use a `uid` until we get the `eventId` from the API
- If the save request succeeded, add the `eventId` to the enrollments collection.
- If the save request failed, remove the event from the enrollments collection and display a message using `FeedbackBar`.


jasminenguyennn (commented):
hey @simonadomnisoru, great work 👍 
do we have the strategy to test this?

simonadomnisoru:
> hey @simonadomnisoru, great work 👍
> do we have the strategy to test this?

Hey @jasminenguyennn, 

As for today, there are no Cypress tests for the new event enrollment form therefore I also didn't add any for my fix. The strategy to test this fix is with manual QA. The steps are detailed in the [jira ticket](https://jira.dhis2.org/browse/DHIS2-11745) and will be:
1. Go to the new event enrollment page
2. Add a new event
3. Enter values and Save.
4. The user is taken to the enrollment dashboard page.
4. Check the new event is not shown under the stage lists.

Thank you!

jasminenguyennn (approved):

PR #2086, opened 2021-09-13
[TECH-672] Navigate to enrollment overview based on dataStore

EirikHaugstulen:
https://jira.dhis2.org/browse/TECH-672

Migrate from old `navigateToTrackedEntityDashboard`-function, make API-call to fetch the dataStore/userDataStore and navigate the user to the correct dashboard.
Users can now try out the new enrollment dashboard with a config-json in the dataStorage with a "program-specific" enabling.  

Steps to test:
Make a POST-request to /api/dataStore/capture/useNewDashboard
Make API-body JSON with [programId]: "true/false"
The capture app should then navigate the user if the program is set to true.
There is also functionality to override the general config with a "per-user" enabling by making the same request as above in the /userDataStore-endpoint. This will then take precedence over the global dataStore-storage.


src/core_modules/capture-core/components/DataStore/DataStore.epics.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { ofType } from 'redux-observable';
+import { mergeMap, catchError } from 'rxjs/operators';
+import { EMPTY } from 'rxjs';
+import { fetchDataStore } from './DataStore.actions';
+import { getApi } from '../../d2';
+import type { UseNewDashboard } from './DataStore.actions';
+import { appStartActionTypes } from '../../../../components/AppStart';
+
+function getDataStoreFromApi() {
```
simonadomnisoru, line 10: Hey, 
Does it make sense to use the [DHIS2 hooks ](https://runtime.dhis2.nu/#/hooks/) to get/send the data from/to the API? 
Thanks!

src/core_modules/capture-core/actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.actions.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import { actionCreator } from '../actions.utils';
+
+export const actionTypes = Object.freeze({
+    NAVIGATE_TO_ENROLLMENT_OVERVIEW: 'enrollmentNavigation.navigateToEnrollmentOverview',
+});
+
+type NavigateToEnrollmentOverviewProps = {|
```
simonadomnisoru, line 9: Hey, 
In order to maintain homogeneity with the rest of the folders structure, I think is more sensible to move the type declaration from `navigateToEnrollmentOverview.actions.js` file to `navigateToEnrollmentOverview.types.js`.
Thanks!

src/core_modules/capture-core/components/DataStore/DataStore.actions.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import { actionCreator } from '../../actions/actions.utils';
+
+export const actionTypes = {
+    FETCH_DATA_STORE: 'useNewDashboard.FetchDataStore',
+};
+
+export type UseNewDashboard = {|
```
simonadomnisoru, line 8: Same thing here regarding the `type` declaration to be moved in a `.types` file

src/core_modules/capture-core/components/DataStore/DataStore.epics.js
EirikHaugstulen: Hey,
I might be wrong, but since this is in an Epic I thought we couldn't use hooks? Since it's not a React function nor a component? Thanks!

src/core_modules/capture-core/actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.actions.js
EirikHaugstulen: I agree, thanks!

src/core_modules/capture-core/components/DataStore/DataStore.epics.js
```diff
@@ -0,0 +1,45 @@
...
+    action$.pipe(
+        ofType(
+            appStartActionTypes.APP_LOAD_SUCESS,
+        ),
+        mergeMap(async () => {
+            const apiDataStore: UseNewDashboard = await getDataStoreFromApi();
+            // $FlowFixMe
+            return fetchDataStore({ dataStore: apiDataStore });
+        }),
+        catchError(() => EMPTY),
```
simonadomnisoru, line 31: Does it make sense to log the error here using `log.error `and `errorCreator`?

src/core_modules/capture-core/components/DataStore/DataStore.epics.js
```diff
@@ -0,0 +1,45 @@
...
+    action$.pipe(
+        ofType(
+            appStartActionTypes.APP_LOAD_SUCESS,
+        ),
+        mergeMap(async () => {
+            const apiUserDataStore: UseNewDashboard = await getUserDataStoreFromApi();
+            // $FlowFixMe
+            return fetchDataStore({ userDataStore: apiUserDataStore });
+        }),
+        catchError(() => EMPTY),
```
simonadomnisoru, line 44: same doubt here about the logging of the error.

src/core_modules/capture-core/actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.epics.js
```diff
@@ -0,0 +1,61 @@
...
+        // TODO This will be removed when the link between capture and tracker capture is not relevant
+        const { baseUrl } = config;
+        const { search, pathname } = dependencies.history.location;
+        const {
+            programId: queryProgramId,
+            trackedEntityTypeId: queryTrackedEntityTypeId,
+        } = store.value.router.location.query;
+
+        const instanceBaseUrl = baseUrl.split('/api')[0];
+        const scopeHierarchy = queryProgramId ? 'PROGRAM' : 'TRACKED_ENTITY_TYPE';
```
simonadomnisoru, line 50: Can you extract `PROGRAM` and `TRACKED_ENTITY_TYPE` to const? Thank you!

src/core_modules/capture-core/components/DataStore/DataStore.epics.js
EirikHaugstulen: Since this is an "experimental feature", we don't really want to show this as an error in the console. The most common thing is that this dataStorage is not available since not many programs will use the new dashboard with this version. @Bekkalizer feel free to correct me if I'm wrong 👍

Bekkalizer: Yeah the key will usually not be in the dataStore, and we don't need a warning/error in the logs I think...

simonadomnisoru: Ok, thanks for the explanation!

simonadomnisoru (approved):

src/core_modules/capture-core/actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.epics.js
EirikHaugstulen: Yes, good catch, changed! Thanks!

PR #2089, opened 2021-09-13
feat: rules engine support for "enrollment event" page

superskip:


JoakimSM (approved):
Good work @superskip !

PR #2091, opened 2021-09-13
[TECH-681] Refactor api-calls to the /events-endpoint from the EventWorkingList

EirikHaugstulen:
https://jira.dhis2.org/browse/TECH-681

On initial load, the eventWorkingList adds a `programStageId=eventProgramStage`-filter. This does nothing and should probably be removed.
The /events-endpoint has now been refactored and will require an ouMode if orgUnit is not selected. Add this to the API-call from the EventWorkingList.

### DIFF
--- a/src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.js
+++ b/src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.js
@@ -177,6 +177,12 @@ export const getEventListData = async (
     categoryCombinationId?: ?string,
 ) => {
     const mainColumns = getMainColumns(columnsMetaForDataFetching);
+
+    // Remove programStageId from the API if program is single event
+    if (queryArgs?.programStageId === 'EventProgramStage') {
+        queryArgs.programStageId = undefined;
+    }
+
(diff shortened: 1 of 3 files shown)


src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.js
```diff
@@ -177,6 +177,11 @@ export const getEventListData = async (
     categoryCombinationId?: ?string,
 ) => {
     const mainColumns = getMainColumns(columnsMetaForDataFetching);
+    // Remove programStageId from the API if program is single event
+    if (queryArgs?.programStageId === 'EventProgramStage') {
+        queryArgs.programStageId = undefined;
```
simonadomnisoru, line 182: Hey, 
I am curious to know why you don't remove `programStageId` from the object and prefer to use `undefined`?  
Thanks!

EirikHaugstulen: Just a matter of old habits I think. Changed it now, thanks! 👍

simonadomnisoru (approved):

Bekkalizer (approved):

PR #2095, opened 2021-09-14
fix: [DHIS2-11745] add orgUnitName to the new enrollment event in Redux

simonadomnisoru:
add `orgUnitName` to the new enrollment event in Redux to be able to display it in the Stage events table

### DIFF
--- a/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/validated.epics.js
+++ b/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/validated.epics.js
@@ -68,10 +68,13 @@ export const saveNewEnrollmentEventEpic = (action$: InputObservable, store: Redu
                 enrollmentId,
                 completed,
             });
+            const orgUnit = serverData.events[0]?.orgUnit || '';
+            const orgUnitName = state.organisationUnits[orgUnit]?.name || '';
+            const eventData = { ...serverData.events[0], orgUnitName };
 
             onSaveExternal && onSaveExternal(serverData);
             return batchActions([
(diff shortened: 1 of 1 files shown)


src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/validated.epics.js
```diff
@@ -68,10 +68,13 @@ export const saveNewEnrollmentEventEpic = (action$: InputObservable, store: Redu
                 enrollmentId,
                 completed,
             });
+            const orgUnit = serverData.events[0]?.orgUnit || '';
```
Bekkalizer, line 71: I see this is an inherited assumption, but why do we ask for the first event in the array? Will there always be one event in this list?

simonadomnisoru: Hey, the function [getAddEventEnrollmentServerData](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/getConvertedNewSingleEvent.js#L64) returns the array `events` with one event. Therefore, I think is safe to assume there will be only one event in [serverData.events](https://github.com/dhis2/capture-app/blob/a3f839d02db9066497487e75943f79d4241c1f4f/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/validated.epics.js#L61)

Bekkalizer: Ok, thanks - just checking :)

Bekkalizer (approved):

PR #2110, opened 2021-09-15
fix: [DHIS2-11783] fetch orgUnit if none present

jasminenguyennn:


simonadomnisoru:
Hey, 
Does it make sense to write a Cypress test to check that the effects of the rules are executed and displayed correctly? 
Thanks!

jasminenguyennn:
@simonadomnisoru I don't think cypress test in pressing time like this case makes sense.

simonadomnisoru (approved):

PR #2115, opened 2021-09-16
fix: [DHIS2-11791] widget comments don't show in edit mode

jasminenguyennn:
### Issue
[Ticket](https://jira.dhis2.org/browse/DHIS2-11791): Comments don't load in edit mode

### Solution

This is because `dataEntryNotes` is not set when loadDataEntry so I add the `dataEntryNotes` and removed unused param `foundation`

### DIFF
--- a/cypress/integration/WidgetsForEnrollmentPages/WidgetEventComment/index.js
+++ b/cypress/integration/WidgetsForEnrollmentPages/WidgetEventComment/index.js
@@ -3,12 +3,10 @@ Then('the enrollment widget should be loaded', () => {
 });
 
 When('you click edit mode', () => {
-    cy.get('[data-test="widget-enrollment-event"]').within(() => {
-        cy.get('[data-test="dhis2-uicore-button"]')
-            .contains('Edit event')
-            .click();
-        cy.wait(100);
-    });
+    cy.get('[data-test="dhis2-uicore-button"]')
+        .contains('Edit event')
+        .click();
+    cy.contains('Enrollment: Edit Event').should('exist');
 });
 
 When(/^you fill in the comment: (.*)$/, (comment) => {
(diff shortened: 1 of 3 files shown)


cypress/integration/WidgetsForEnrollmentPages/WidgetEventComment/index.js
```diff
@@ -2,6 +2,14 @@ Then('the enrollment widget should be loaded', () => {
     cy.contains('The enrollment event data could not be found').should('not.exist');
 });
 
+When('you click edit mode', () => {
+    cy.get('[data-test="widget-enrollment-event"]').within(() => {
+        cy.get('[data-test="dhis2-uicore-button"]')
+            .contains('Edit event')
+            .click();
+        cy.wait(100);
```
simonadomnisoru, line 10: Hey, 
Why is there a need to wait for 100 ms? Maybe you can wait for some text to appear on the page rather than have a fixed waiting time 🤔. 
The similar step is used in [EnrollmentEditEventPageForm](https://github.com/dhis2/capture-app/blob/master/cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageForm/index.js#L9) and it works without `cy.wait`. 
Thanks!

jasminenguyennn: hey it needs to wait for the widget to be loaded and redux data added before accessing the form, I tried to wait for a text but doesn't work so keep it like that

Bekkalizer: @jasminenguyennn can you have a second look at this one. I want us to try very hard to avoid having waiting statements in the code, we introduce an unreliable test here that might fail if the test execution takes longer than usual.

jasminenguyennn: @Bekkalizer sure. I updated it, if we are not rushing for merging as QA wanted.

simonadomnisoru (approved):

JoakimSM (approved):
Sorry for leaving this PR hanging. I was not able to look at the other PRs for the comments Widget so I'm not really up to speed on this one. But from my limited understanding, this seems like a good fix  👍

PR #2119, opened 2021-09-17
fix: [DHIS2-11745][DHIS2-11797] add missing data & save event app crash

simonadomnisoru:
- send and display `dueDate` and `orgUnitName`
- disable the event row in the Stage table while we wait for the API response.
- move `saveNewEventSucceededEpic`, `saveNewEventFailedEpic` and `updateEnrollmentEventsWithoutId` logic from the `WidgetEnrollmentEventNew` folder to the `EnrollmentAddEventPage` page folder
- fix the `enrollmentDomain` reducers 

The cypress tests failures are related to the instability of the live test instance


src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/getConvertedAddEvent.js
```diff
@@ -44,6 +46,8 @@ export const getAddEventEnrollmentServerData = ({
                 orgUnit: orgUnitId,
                 trackedEntityInstance: teiId,
                 enrollment: enrollmentId,
+                dueDate: mainDataServerValues.eventDate,
+                orgUnitName: organisationUnits[orgUnitId].name,
```
JoakimSM, line 50: I think I have removed every dependency to data that are being set in the app sync for the new enrollment event page. So, if we can get the organisation unit name from here instead it would be great: https://github.com/dhis2/capture-app/blob/7daf0e63d0fcbb23278dc0af0c4111db057cb603/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/Validated.container.js#L31

(yes, we should try to remove the app sync soon for this page 🙂 )

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -38,9 +35,13 @@ export const EnrollmentAddEventPage = () => {
         dispatch(navigateToEnrollmentPage(programId, orgUnitId, teiId, enrollmentId));
     }, [dispatch, programId, orgUnitId, teiId, enrollmentId]);
 
-    const handleSave = useCallback(() => {
-        dispatch(navigateToEnrollmentPage(programId, orgUnitId, teiId, enrollmentId));
-    }, [dispatch, programId, orgUnitId, teiId, enrollmentId]);
+    const handleSave = useCallback(
+        (data, uid) => {
+            uid && dispatch(updateEnrollmentEventsWithoutId(uid, data.events[0]));
```
JoakimSM, line 40: Can we annotate the `uid` instead of doing `uid && ...`?
```suggestion
        (data, uid: string) => {
           dispatch(updateEnrollmentEventsWithoutId(uid, data.events[0]));
```

JoakimSM (commented):
Thanks @simonadomnisoru. Couple of comments, but it looks really good overall 👍

jasminenguyennn (changes requested):
@simonadomnisoru can you add cypress test for this fix. thanks

simonadomnisoru:
> @simonadomnisoru can you add cypress test for this fix. thanks

Hey @jasminenguyennn, 

As for today, there are no Cypress tests for the new event enrollment form therefore I also didn't add any for my fix. Until we are able to delete an event from UI, there cannot be any cypress tests for this form. The strategy to test this fix is with manual QA. 

Thanks!

jasminenguyennn:
@simonadomnisoru I thought you previously mentioned that it will be easier for the board control to approve fix with test coverage, that also apply for new pull request. Also I think the failure from old tests should not affect the result of the new test.

simonadomnisoru:
> @simonadomnisoru I thought you previously mentioned that it will be easier for the board control to approve fix with test coverage, that also apply for new pull request. Also I think the failure from old tests should not affect the result of the new test.

Hey @jasminenguyennn, 

Because we test against a live instance, a Cypress test must leave the database clean after it is executed. In this case, as I mentioned before, we don't have the delete event functionality implemented yet. Therefore if I add a new event, currently there is no way to delete it via Cypress and the database will be altered. I don't think this is the best approach. 
I will leave it up to @JoakimSM to decide what is best and if we should write Cypress tests for the new enrollment event form.

Thank you!

JoakimSM (approved):
Thanks 👍

JoakimSM:
Sorry, didn't notice your discussion. Let's add tests in another PR/ticket: https://jira.dhis2.org/browse/TECH-686

jasminenguyennn (approved):

PR #2135, opened 2021-09-22
fix: [DHIS2-11814][DHIS2-11822] add fallback for tei missing name

simonadomnisoru:


JoakimSM (approved):
Looks good @simonadomnisoru 👍

PR #2136, opened 2021-09-22
[DHIS2-11795] - Hide Non-searchable filters when in ACCESSIBLE-mode

EirikHaugstulen:
https://jira.dhis2.org/browse/DHIS2-11795

Fix permission issue when using the ACCESSIBLE-mode in WorkingList-component.

### DIFF
--- a/src/core_modules/capture-core/components/ListView/Filters/Filters.component.js
+++ b/src/core_modules/capture-core/components/ListView/Filters/Filters.component.js
@@ -29,7 +29,7 @@ type Props = {
 const getValidElementConfigsVisiblePrioritized = (columns: Array<Column>) =>
     new Map(
         columns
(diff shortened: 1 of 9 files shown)


simonadomnisoru (approved):
LGTM!

JoakimSM (commented):
Good job @EirikHaugstulen. I know we talked about creating this `filterHidden` prop, but looking at the code: does this do the same job as `filterDisabled`? If so, maybe we remove `filterDisabled` and use `filterHidden` instead? `filterHidden` seems to be the better name for the functionality here.

EirikHaugstulen:
Yep, seems reasonable @JoakimSM! 
Removed `filterDisabled` and changed the current implementations to `filterHidden` 👍

JoakimSM (approved):
👍

PR #2138, opened 2021-09-23
feat: [DHIS2-11830] implement auto select enrollment

jasminenguyennn:
### Issue:
Ticket: [DHIS2-11830](https://jira.dhis2.org/browse/DHIS2-11830) Auto-select enrollment when opening enrollment dashboard

### DIFF
--- a/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
+++ b/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
@@ -17,6 +17,7 @@ import { useEnrollmentInfo } from './useEnrollmentInfo';
 import { enrollmentPageStatuses } from './EnrollmentPage.constants';
 import { getScopeInfo } from '../../../metaData';
 import { buildEnrollmentsAsOptions } from '../../ScopeSelector';
+import { setEnrollmentSelection } from '../../LockedSelector/LockedSelector.actions';
 
 const useComponentLifecycle = () => {
     const dispatch = useDispatch();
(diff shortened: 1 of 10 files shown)


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
```diff
@@ -28,11 +29,12 @@ const useComponentLifecycle = () => {
...
 
     const { scopeType } = useScopeInfo(programId);
-    const { programHasEnrollments, enrollmentsOnProgramContainEnrollmentId } = useEnrollmentInfo(enrollmentId, programId);
+    const { programHasEnrollments, enrollmentsOnProgramContainEnrollmentId, autoEnrollmentId } = useEnrollmentInfo(enrollmentId, programId);
     useEffect(() => {
         const selectedProgramIsTracker = programId && scopeType === scopeTypes.TRACKER_PROGRAM;
-
-        if (selectedProgramIsTracker && programHasEnrollments && enrollmentsOnProgramContainEnrollmentId) {
+        if (enrollmentId === 'AUTO' && autoEnrollmentId) {
+            dispatch(setEnrollmentSelection({ enrollmentId: autoEnrollmentId }));
```
JoakimSM, line 36: Can we, instead of going with the epics, use the hook that is defined here:
https://github.com/dhis2/capture-app/blob/0772559f4fd54f53b44d9fc1478c36e07716b027/src/core_modules/capture-core/components/ScopeSelector/hooks/useSetEnrollmentId.js#L6

We need to make one change though; use history.replace instead of history.push (if not, you will get a hard time going back with the browser back button; you will be forwarded again). So, make it an option to replace instead of pushing.

JoakimSM (commented):
Thanks @jasminenguyennn, I would like us to make one change; see review comment. But looks good overall 👍

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.container.js
jasminenguyennn: 👍  thanks @JoakimSM  I update the changes as suggested

JoakimSM (approved):
Thanks 👍

PR #2150, opened 2021-09-27
fix: [DHIS2-11850] Tracker fails to load when opening TEI without orgUnit

EirikHaugstulen:
Fetch the ownerOrgUnit from the API and push with the navigation to TC.

https://jira.dhis2.org/browse/DHIS2-11850


simonadomnisoru:
Hey @EirikHaugstulen,
Are there any tests you can add to ensure that the bug is fixed? 
Thanks!

EirikHaugstulen:
Hey @simonadomnisoru,
This does not do any UI-changes, and is scheduled to be removed in 2.38, so I don't really have any tests in mind.
Please feel free to nudge me in the right direction if there's anything specific you were thinking about.
Thanks!

simonadomnisoru (approved):

Bekkalizer (approved):

PR #2165, opened 2021-10-04
feat: [DHIS2-11617] Implement switch tab in add event page

jasminenguyennn:
### Feature
[DHIS2-11617 ](https://jira.dhis2.org/secure/RapidBoard.jspa?rapidView=80&view=detail&selectedIssue=DHIS2-11617&quickFilter=230): Mode for scheduling event

### DIFF
--- a/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/WidgetEnrollmentEventNew.container.js
+++ b/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/WidgetEnrollmentEventNew.container.js
@@ -1,15 +1,32 @@
 // @flow
-import React, { useMemo, useState, useRef } from 'react';
+import React, { useMemo, useState, useRef, type ComponentType } from 'react';
 import i18n from '@dhis2/d2-i18n';
-import { TabBar, Tab } from '@dhis2/ui';
+import withStyles from '@material-ui/core/styles/withStyles';
+import { TabBar, Tab, spacersNum, colors } from '@dhis2/ui';
+import { Widget } from '../Widget';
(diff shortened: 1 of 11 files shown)


src/core_modules/capture-core/components/WidgetEnrollmentEventNew/WidgetEnrollmentEventNew.container.js
```diff
@@ -18,13 +42,69 @@ export const WidgetEnrollmentEventNew = ({ programId, stageId, onSave, ...passOn
 
     const formFoundation = stage.stageForm;
 
+    const onHandleSwitchTab = (newMode) => {
+        setWarningVisible(true);
```
JoakimSM, line 46: We should only show the warning if the user has interacted with the tab contents (e.g, changed any form values).

JoakimSM (commented):
Thanks for the PR @jasminenguyennn.

It would be great if we can change our approach a bit here.

I suggest we create self-contained widgets for the contents of each tab; i.e. the «Report» tab will contain the WidgetEnrollmentEventNew widget, the «Schedule» tab will contain a schedule widget and the «Refer» tab will contain a refer widget (all self-contained). The idea is that we later can use these widgets in other apps.

So, with this approach, the code for the TabBar should be moved from the Widget to the Page itself (a new component within the page would be preferred).

We can have a chat about this tomorrow if you like.

EirikHaugstulen:
> @EirikHaugstulen oh I forgot we have Quick Action, I can add a new query param `&tab=REPORT`. WDYT?

Yep, that sounds great! 
Thank you 👍

PR #2151, opened 2021-09-28
fix: [DHIS2-11797] clean the dataElements sent to the server

simonadomnisoru:
Remove the dataElements that don't belong to the program stage when creating a new event. The tests will be added in 
[TECH-686 ](https://jira.dhis2.org/browse/TECH-686) as we previously agreed upon in https://github.com/dhis2/capture-app/pull/2119

Cypress tests failures are related to a BE change see more in [Cypress known issues](https://github.com/dhis2/capture-app/wiki/Pull-Request-Guidelines#cypress-known-issues)

PR #2167, opened 2021-10-04
fix: [DHIS2-11797] Prevent out of scope effects to trigger

JoakimSM:
Superseeds https://github.com/dhis2/capture-app/pull/2151

Utilising the ideas from https://github.com/dhis2/capture-app/pull/2089 to prevent out of scope effect to trigger.


simonadomnisoru (approved):

superskip:
> Thanks @superskip. Yes, I agree that the name is not the best, too generic. Not sure that I'm completely sold on "metaDataType" either though. Tough one. Maybe "outputType" could do it?

"targetMetaDataType"? Naming it "outputType" makes it confusingly similar to the already existing "type" property.

PR #2174, opened 2021-10-05
feat: [DHIS2-11633][DHIS2-11878] edit profile widget

simonadomnisoru:
- get the data from API
-  build the `formFundation` 
- Call the `RulesEngine` when the form opens and when a value changes
- trigger the proper redux actions to open and show the form with the values
- Display the program rules effects


src/core_modules/capture-core/components/WidgetProfile/DataEntry/buildForm.js
```diff
@@ -0,0 +1,75 @@
...
+        convertedProgramTrackedEntityAttributes = [...convertedProgramTrackedEntityAttributes, asyncForEach];
+    }
+
+    if (!convertedProgramStages.asyncForEach) {
+        convertedProgramStages = programStages.map(programStage => ({
+            ...programStage,
+            programStageDataElements: { ...programStage.programStageDataElements, asyncForEach },
+        }));
+        convertedProgramStages = [...convertedProgramStages, asyncForEach];
+    }
```
simonadomnisoru, line 41: This hack is needed for the Cypress tests. I could not found why but `asyncForEach` property from Array prototype is undefined. This error is happening only in Cypress, localhost works ok. 
Let me know if you have another idea on how to fix it. 
Thank you!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/DataEntry.container.js
```diff
@@ -0,0 +1,105 @@
...
+                            onGetValidationContext={() => {}}
+                            saveAttempted={undefined}
+                        />
+                    </ModalContent>
+                    <ModalActions>
+                        <ButtonStrip end>
+                            <Button onClick={onCancel} secondary>
+                                {i18n.t('Cancel without saving')}
+                            </Button>
+                            <Button onClick={() => {}} primary>
```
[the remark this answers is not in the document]
simonadomnisoru: I added the TODO infos. Thanks!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
```diff
@@ -0,0 +1,196 @@
...
+    cachedOptionSets: Map<string, CachedOptionSet>,
+|};
+
+export class DataElementFactory {
+    static errorMessages = {
+        TRACKED_ENTITY_ATTRIBUTE_NOT_FOUND:
+            'TrackedEntityAttributeId missing from programTrackedEntityAttribute or trackedEntityAttribute not found',
+    };
+
+    static buildtetFeatureType(featureType: 'POINT' | 'POLYGON') {
```
jasminenguyennn, line 34: is it `tei` instead of `tet`

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
```diff
@@ -0,0 +1,196 @@
...
+        const dataElement = new DataElement((o) => {
```
jasminenguyennn, line 35: I think we agree not to use single character naming. thanks

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
```diff
@@ -0,0 +1,196 @@
...
+                                },
+                            };
+                        }
+                    }
+                }
+
+                let requestPromise;
+                if (o.scope === dataElementUniqueScope.ORGANISATION_UNIT) {
+                    const orgUnitId = contextProps.orgUnitId;
+                    requestPromise = getApi().get('trackedEntityInstances', {
```
jasminenguyennn, line 80: I thought we are moving away from `getApi`

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    optionSetFactory: OptionSetFactory;
+    cachedTrackedEntityAttributes: Map<string, CachedTrackedEntityAttribute>;
+    constructor({ cachedTrackedEntityAttributes, cachedOptionSets }: ConstructorInput) {
+        this.cachedTrackedEntityAttributes = cachedTrackedEntityAttributes;
+        this.optionSetFactory = new OptionSetFactory(cachedOptionSets);
+    }
+
+    // eslint-disable-next-line complexity
```
jasminenguyennn, line 120: why do we need to disable complexity here, is there any way to split the function to smaller ones?

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/EnrollmentFactory.js
```diff
@@ -0,0 +1,200 @@
...
+    }: ConstructorInput) {
+        this.trackedEntityTypeCollection = trackedEntityTypeCollection;
+        this.cachedTrackedEntityTypes = cachedTrackedEntityTypes;
+        this.dataElementFactory = new DataElementFactory({
+            cachedTrackedEntityAttributes,
+            cachedOptionSets,
+        });
+    }
+
+    _buildTetFeatureTypeField(trackedEntityTypeId: ?string) {
```
jasminenguyennn, line 61: same here is it `tei`?

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -1,68 +1,56 @@
...
+};
 
-    const {
-        loading: programsLoading,
-        data: programsData,
-        error: programsError,
-    } = useDataQuery(programsQuery);
+const WidgetProfilePlain = ({ teiId, programId, showEdit = false, orgUnitId = '', classes }: Props) => {
+    const [open, setOpenStatus] = useState(true);
+    const [toggleEditModal, setToggleEditModal] = useState(false);
```
jasminenguyennn, line 30: should this take the `showEdit` as initial state?

simonadomnisoru: No, `showEdit` and `toggleEditModal` are different variables and in charge of different functionality. `toggleEditModal` opens and closes the edit profile modal, while `showEdit` enable the edit profile feature. Thanks

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
simonadomnisoru: I think `tet` is correctly used here because the code is about the `tracked entity type`. Thanks!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/EnrollmentFactory.js
simonadomnisoru: Same as above, I think `tet` is correctly used here because the code is about the `tracked entity type`. Thanks!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
simonadomnisoru: Hey, 
I was not aware we are moving away from `getApi`. What are we supposed to use instead? 
In this case, I cannot use the `useDataQuery` hook from `@dhis2/app-runtime` because I am not inside a react component. What do you suggest to use insted of `getApi`?
Thank you!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/factory/DataElementFactory.js
jasminenguyennn: so can we make it camelCase? TET or Tet? Thanks

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
jasminenguyennn: but if there is no showEdit button then how can it open the Edit modal?

simonadomnisoru: As mention above the variables and in charge of different functionality.

- `showEdit === false`
       When there is no `showEdit` button then the feature will not be displayed at all. The value of `toggleEditModal` does not matter in this case. 

- `showEdit === true`
      In this case a conflict will appear if `showEdit` is set as the initial state for `toggleEditModal`. The modal will be opened from the initialization. This is not the desired behavious for this feature. 

Thanks!

jasminenguyennn (approved):
👍

src/core_modules/capture-core/components/WidgetProfile/DataEntry/hooks/useLifecycle.js
```diff
@@ -0,0 +1,123 @@
...
+            programAPI?.programTrackedEntityAttributes?.reduce(
+                (acc, currentValue) => ({
+                    ...acc,
+                    [currentValue.trackedEntityAttribute.id]: currentValue.trackedEntityAttribute,
+                }),
+                {},
+            ),
+        [programAPI],
+    );
+    const optionSets: OptionSets = useMemo(
```
JoakimSM, line 52: We should also include the optionSets from the dataElements in the program stages here.

src/core_modules/capture-core/components/WidgetProfile/DataEntry/hooks/useLifecycle.js
```diff
@@ -0,0 +1,123 @@
...
+
+    useEffect(() => {
+        if (
+            rulesContainer &&
+            Object.entries(rulesContainer).length > 0 &&
+            formFoundation &&
+            Object.entries(formFoundation).length > 0
+        ) {
+            dispatch(
+                getOpenDataEntryActions({
```
JoakimSM, line 91: Maybe we can split getOpenDataEntryActions into two parts: One for making the form itself ready and another one for running the initial rules execution. Reason coming to mind: When orgUnit/orgUnitId changes, we would like to rerun rules, but not clear the form. In our enrollment dashboard overview this problem probably won't surface, but if we think two steps ahead and want to use the Widget elsewhere, changing orgUnit while the form is displayed might be a possibility.

src/core_modules/capture-core/components/WidgetProfile/DataEntry/hooks/useLifecycle.js
```diff
@@ -0,0 +1,123 @@
...
+            mergedAttributes.reduce(
+                (acc, currentValue) => ({ ...acc, [currentValue.reactKey]: currentValue.value }),
+                {},
+            ),
+        [mergedAttributes],
+    );
+
+    useEffect(() => {
+        if (!loadingProgramRules && constants && Object.entries(formFoundation).length === 0) {
+            buildFormFoundation(programAPI, setFormFoundation);
```
JoakimSM, line 73: This is mostly nitpicking, but could we abstract away the building of the formFoundation? Meaning we do something like this: `const formFoundation = useFormFoundation(.....)` and the state and the effects would reside In the useFormFoundation hook.

src/core_modules/capture-core/components/WidgetProfile/DataEntry/ProgramRules/getRulesActionsForTEI.js
```diff
@@ -0,0 +1,53 @@
...
+    enrollmentData?: ?Enrollment,
+    teiValues?: ?TEIValues,
+    trackedEntityAttributes: ?TrackedEntityAttributes,
+    optionSets: OptionSets,
+    rulesContainer: ProgramRulesContainer,
+}) => {
+    const effects = rulesEngine.getProgramRuleEffects({
+        programRulesContainer: rulesContainer,
+        currentEvent: null,
+        eventsContainer: null,
```
JoakimSM, line 43: 1. We should include the events here. 
2. I've changed the the rules engine interface a bit.

I think it's better to talk about this one.

src/core_modules/capture-core/components/WidgetProfile/DataEntry/FormFoundation/DataElement.js
```diff
@@ -0,0 +1,228 @@
+// @flow
+/* eslint-disable no-underscore-dangle */
+import log from 'loglevel';
+import i18n from '@dhis2/d2-i18n';
+import { pipe, errorCreator } from 'capture-core-utils';
+
+import type {
+    CachedProgramTrackedEntityAttribute,
+    CachedTrackedEntityAttribute,
+    CachedOptionSet,
```
JoakimSM, line 10: I think we should define our own types for this particular context. There can be differences between what is in the cache and the api data we use here.

JoakimSM (commented):
This is coming together nicely, good work 👍

There are some things I would like us to look into though. I think it would be a good idea to start with a chat so I can elaborate on some of the comments.

JoakimSM:
A quick follow up to: https://github.com/dhis2/capture-app/pull/2308#discussion_r772485061

Apart from changing from mergedAttributes to the new object being created in that PR, you should be aware of and deal with:

- The new object created will have values in client format (if my suggestion is implemented). This is the correct format for the rules engine. For the form data itself, the values will need to be converted from client to form format.
- Another aspect we didn't talk about: The value of the options (the code) in the option sets will need to be converted from api/server format to client format for both the rulesEngine and the RenderFoundation object.

simonadomnisoru:
>  

  

> A quick follow up to: [#2308 (comment)](https://github.com/dhis2/capture-app/pull/2308#discussion_r772485061) 

>  

> Apart from changing from mergedAttributes to the new object being created in that PR, you should be aware of and deal with: 

>  

> * The new object created will have values in client format (if my suggestion is implemented). This is the correct format for the rules engine. For the form data itself, the values will need to be converted from client to form format. 

> * Another aspect we didn't talk about: The value of the options (the code) in the option sets will need to be converted from api/server format to client format for both the rulesEngine and the RenderFoundation object. 

Hey @JoakimSM ,    

Thanks for the review! I implemented the changes you suggested. Regarding the 2 points you mentioned afterward:    
1. I will wait for @jasminenguyennn  [PR](https://github.com/dhis2/capture-app/pull/2308#discussion_r772485061 ) to be merged to be able to use the TEI attributes subValues.  
2. I dug inside `serverToClient`, `convertOptionSetsToRulesEngineFormat` and `metaDataMemoryStores/optionSets.store` and I could not find any example on how to convert the value of the options (the code) in the option sets. Let's have a chat about it to clarify.

Thank you!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/ProgramRules/rulesContainer.js
```diff
@@ -0,0 +1,79 @@
...
+
+    program.programRules = mainRules;
+    Object.keys(rulesByStage).forEach((stageKey) => {
+        const rulesForStage = rulesByStage[stageKey];
+        const programStage = program.getStage(stageKey);
+        if (programStage) {
+            // $FlowFixMe[prop-missing] automated comment
+            programStage.programRules = rulesForStage;
+        }
+    });
```
JoakimSM, line 40: I think the result of this is passed directly into the rulesEngine, if that is the case, we should just create a flat structure of rules here.

JoakimSM (commented):
Thanks @simonadomnisoru , we are almost there I think 👍 Just need to put the finishing touches to some details. Let's have a chat about my recent comments and the option sets you asked about (hopefully tomorrow).

src/core_modules/capture-core/components/WidgetProfile/DataEntry/FormFoundation/RenderFoundation.js
```diff
@@ -0,0 +1,162 @@
...
+    return section;
+};
+
+const buildElementsForSection = async (
+    programTrackedEntityAttributes: ?Array<CachedProgramTrackedEntityAttribute>,
+    trackedEntityAttributes: Array<CachedTrackedEntityAttribute>,
+    optionSets: Array<CachedOptionSet>,
+    section: Section,
+) => {
+    await programTrackedEntityAttributes?.forEach(async (trackedEntityAttribute) => {
```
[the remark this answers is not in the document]
simonadomnisoru: Hey, 
The `asyncForEach` error still persists in Cypress. You can see the error in the below screenshot. 

Screenshot 2022-01-18 at 16 19 31

To avoid this strange error I decided to use the `for-of` loop and now all the promises are correctly handled.

Thanks!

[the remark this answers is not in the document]
JoakimSM: Investigated a bit and it seems data retrieved from the api is missing the asyncForEach when running the Cypress tests.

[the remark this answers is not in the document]
JoakimSM: Makes sense to do `for-of` as you did 👍

JoakimSM (approved):
Nice! Good job

PR #2195, opened 2021-10-09
feat: [DHIS2-11953] Quick actions-widget in enrollment dashboard

eirikhaugstulen:
https://jira.dhis2.org/browse/DHIS2-11953

Quick actions widget with navigation to "New event", "Make referral" and "Schedule an event".


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentQuickActions/QuickActionButton/QuickActionButton.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+type Props = {|
+    icon: Element<any>,
+    label: string,
+    onClickAction: () => void,
+    dataTest?: string,
+    disable?: ?boolean,
+    ...CssClasses,
+|}
```
JoakimSM, line 22: Can we move this to a separate file? Also, I've found that it is, in most cases, better to use the annotation `disable?: boolean` to not allow null values; the benefit being that we can assign default values in the function parameters if we need it.

JoakimSM (commented):
Looks good overall @EirikHaugstulen 👍 A couple of comments

EirikHaugstulen:
Thanks @JoakimSM! 
Made some changes based on your review, give it a look when you have the time. ☺️

simonadomnisoru (approved):
LGTM!

JoakimSM (approved):

PR #2204, opened 2021-10-18
fix: [DHIS2-11986] fix error opening enrollment page

jasminenguyennn:
### Issue 
[DHIS2-11986](https://jira.dhis2.org/browse/DHIS2-11986): Error opening enrollment overview page
- Add new layer to check if the `programId` is defined to prevent redirecting


jasminenguyennn:
Close this PR after the discussion during FE meeting (the page will be done as part of @EirikHaugstulen validation)

PR #2216, opened 2021-10-21
feat: [DHIS2-11864] Schedule event using redux

jasminenguyennn:
### Feature:
[DHIS2-11864](): Schedule event using redux and offline service


src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.container.js
```diff
@@ -41,28 +35,35 @@ export const WidgetEventSchedule = ({
...
-            enrollment: enrollmentId,
-            program: programId,
-            programStage: stageId,
-            status: 'SCHEDULE',
-            notes: [],
-        }] });
-    };
+    useEffect(() => {
+        dispatch(startEditingScheduleEvent());
+    }, []); // eslint-disable-line
```
simonadomnisoru, line 43: Hey,
The best practice for `useEffect` is to either pass an array with all the dependencies or pass nothing. You can read more about it [here](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528). Therefore, I am curious to know why did you disable the `lint` checks in this case? 
Thanks!

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.container.js
```diff
@@ -41,28 +35,35 @@ export const WidgetEventSchedule = ({
...
+
+    useEffect(() => {
+        if (!scheduleDate && suggestedScheduleDate) { setScheduleDate(suggestedScheduleDate); }
+    }, [suggestedScheduleDate]); // eslint-disable-line
```
simonadomnisoru, line 47: Same doubt here, why don't you pass all the dependencies to the array? Thanks!

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.container.js
jasminenguyennn: hey because I only want to run it once on mount. Thanks

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.container.js
jasminenguyennn: hey, because I only what to update the schedule date once when on mount. Thanks

simonadomnisoru (approved):
LGTM!

JoakimSM (commented):
Thanks for the RP @jasminenguyennn. Looks good overall 👍 Couple of comments in the code. Also, can you look into why the "Stages and Events" Widget doesn't show the newly scheduled event before refreshing the enrollment dashboard page? The main workflow should be there, but for some reason it's not working.

JoakimSM (approved):
Thanks @jasminenguyennn. Looks really good now.

PR #2220, opened 2021-10-27
chore: [TECH-357] import moment es6

jasminenguyennn:
### Issue
Replace import moment in es6 way


JoakimSM (approved):
Thanks, looks good to me 👍

PR #2224, opened 2021-10-27
fix: [TECH-776] use URL params from useLocationQuery & history

eirikhaugstulen:
Suggestion as to how we might solve the issue where react-router updates with a new set of values before react-redux. 
This would make `connected-react-router` redundant and fix random crashes when opening different capture-pages.


src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,44 +1,39 @@
...
 export const EnrollmentAddEventPage = () => {
     const { programId, stageId, orgUnitId, teiId, enrollmentId } = useLocationQuery();
-
+    const history = useHistory();
     const dispatch = useDispatch();
 
     const handleCancel = useCallback(() => {
-        dispatch(navigateToEnrollmentPage(programId, orgUnitId, teiId, enrollmentId));
-    }, [dispatch, programId, orgUnitId, teiId, enrollmentId]);
+        history.push(`/enrollment?${buildUrlQueryString({ orgUnitId, programId, teiId, enrollmentId })}`);
```
jasminenguyennn, line 22: is there a reason why don't we keep the `dispatch` and modify the `navigateToEnrollmentPage` function?

simonadomnisoru:
Hey @eirikhaugstulen, 
I think this PR makes a strong case to have all the pathnames from inside `AppPages.components.js` moved to a const. Something like this: 
```
const pathnames = {
    VIEW_EVENT: '/viewEvent',
    SEARCH: '/search',
    NEW: '/new',
    ENROLLMENT_EVENT_EDIT: '/enrollmentEventEdit',
    ENROLLMENT_EVENT_NEW: '/enrollmentEventNew',
    ENROLLMENT: '/enrollment',
    ENROLLMENT_STAGE_EVENTS: '/enrollment/stageEvents',
    MAIN_PAGE: '/',
    MAIN_PAGE_KEYS: '/:keys',
}
Object.freeze(pathnames);
```

The const can be exported and used across the whole app in `AppPages.components.js`, `history.push(...)`, `pageToPush` etc. You can also have a look at the existing `const pageKeys` from `withAppUrlSync` HOC and maybe build on it. 
WDYT?

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
eirikhaugstulen: I think my reasoning for this change is that we do not have access to the useHistory-hook in `EnrollmentAddEventPage.actions`, and therefore it would be easier to just remove it. We could pass in the history object, but this would require some rewrite. Does this make sense, or is it something I'm forgetting? 
Thanks!

eirikhaugstulen:
Hey, @simonadomnisoru.
I agree that we should implement the code changes suggested, but after a quick chat with Joakim, I think it would make sense to extract this as a separate issue. Created [TECH-889](https://jira.dhis2.org/browse/TECH-889) and assigned it in my name. Thanks!

src/core_modules/capture-core/components/Pages/ViewEvent/epics/viewEvent.epics.js
```diff
@@ -134,18 +137,22 @@ export const backToMainPageEpic = (action$: InputObservable, store: ReduxStore)
...
+export const backToMainPageLocationChangeEpic = (action$: InputObservable, store: ReduxStore, { history }: ApiUtils) =>
     action$.pipe(
         ofType(viewEventActionTypes.START_GO_BACK_TO_MAIN_PAGE),
         map(() => {
+            // TODO - This should probably be replaced by URL args
             const state = store.value;
             const programId = state.currentSelections.programId;
             const orgUnitId = state.currentSelections.orgUnitId;
             const showaccessible = state.currentSelections.showaccessible;
+            // const { programId, orgUnitId, showaccessible } = deriveURLParamsFromHistory(history);
```
simonadomnisoru, line 149: Hey,
Is there any reason why the ids cannot be currently replaced by URL args?
Thanks!

src/core_modules/capture-core/utils/routing/deriveURLParamsFromLocation.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+export const deriveURLParamsFromLocation = (): {| [key: string]: string |} => {
+    const urlSearch = window.location.hash.split('?');
+    return [...new URLSearchParams(urlSearch[1]).entries()].reduce((accParams, entry) => {
```
simonadomnisoru, line 5: Hey,
I think that in order to have better code readability it would be useful if `urlSearch[1]`, `entry[0]`, `entry[1]` were stored in variables with a meaningful name. WDYT?
Thanks!

src/core_modules/capture-core/components/Pages/ViewEvent/epics/viewEvent.epics.js
eirikhaugstulen: Hey,
This is because it's used in the viewEvent-page. The viewEvent-url does not follow the conventional standard of having the id's in the URL, instead it's just `viewEventId=`. This should probably be rewritten if we want to move away from currentSelections. (although in a different issue)
Should I create one?
Thanks.

src/core_modules/capture-core/utils/routing/deriveURLParamsFromLocation.js
eirikhaugstulen: Yep, should be better now. 
Thanks!

src/core_modules/capture-core/components/Pages/ViewEvent/epics/viewEvent.epics.js
simonadomnisoru: This makes sense, thank you for the explanation!
There is no need to create another ticket, when we will remove the `currentSelections` logic this refactor will come organically.

simonadomnisoru (approved):

jasminenguyennn (approved):
👍

JoakimSM (approved):
Thanks @eirikhaugstulen , look good to me 👍

PR #2227, opened 2021-11-03
chore: [TECH-770] update dhis2 dependencies

JoakimSM:
Updates our dhis2 dependencies (ui, app-runtime, app-scripts etc). 

Caveats:

- You will need node version 14 (or higher)
- Sometimes when switching branches, yarn will mess up the dependency tree. I will create alignment PRs for the v37 and v36 branches. Workaround: manually delete the node_modules folder.


simonadomnisoru (approved):
LGTM!

PR #2232, opened 2021-11-05
fix: [DHIS2-12077] fix i18n string

jasminenguyennn:
### Issue:
[DHIS2-12077](https://jira.dhis2.org/browse/DHIS2-12077): Non-translatable fields in Capture

**Note**
"Complete" has two different meanings but the string is translated just once.
Complete in Save event (as a verb) and Complete as a status in the Enrollment widget
will considere "Completed" as the status in the Enrollment widget


simonadomnisoru (approved):

JoakimSM (approved):
LGTM

PR #2233, opened 2021-11-05
fix: prevent sending empty data values to the api

JoakimSM:
When doing a POST or PUT to the events endpoint, it's unnecessary to send in empty data values (the reason this is unnecessary for the PUT, is that the old values are not preserved)


simonadomnisoru (approved):

jasminenguyennn (approved):
👍

PR #2236, opened 2021-11-09
fix: [DHIS2-11084] - Bug in TE-Attribute search

EirikHaugstulen:
Fix bug that returns undefined in some cases.


simonadomnisoru (approved):

jasminenguyennn (approved):
👍

PR #2253, opened 2021-11-15
fix: refetch dependency causing infinite loop

JoakimSM:
Up app-runtime to utilize the fix here: https://github.com/dhis2/app-runtime/pull/1071

Prior to this fix the refetch function (from useDataQuery) was recreated on each render; this caused infinite loops in some of our fetching code (when refetch was added to the useEffect dependency array the hook would run on every render)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "serverVersion": "37",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


simonadomnisoru (approved):

jasminenguyennn (approved):
👍

PR #2256, opened 2021-11-16
fix: [DHIS2-12122] fix overdue condition

jasminenguyennn:
#### Note:
Test will be added when Schedule function finished


simonadomnisoru (approved):
LGTM!

PR #2265, opened 2021-11-18
chore: [TECH-637] add tooltip and greyout disable event

jasminenguyennn:
### Issue
[TECH-637](https://jira.dhis2.org/browse/TECH-637): Events that are not yet preserved to the server should be disabled/non clickable

- Add tooltip and greyout effect when hovering disabled event
Screenshot 2021-11-17 at 15 46 32


simonadomnisoru (approved):

JoakimSM (commented):
I found that the tooltip only appeared when hovering the actual value of a cell (if the cell was empty the tooltip would not show). Started digging into it and found it wasn't so easy to fix, so I pushed commit. Please review, there is value in understanding what is going on here.

We could have added the tooltip to the row instead, but I found that the message could appear quite far from my cursor so opted to keep using the cell.

(I'm also using the dhis2-ui component instead of material-ui)

simonadomnisoru (approved):
Hey Joakim, 
Great job! It looks better now with the DHIS2-UI component.

jasminenguyennn:
@JoakimSM thanks! I saw other places are still using Tooltip from material-ui, maybe we can migrate it to dhis2-ui later?

PR #2267, opened 2021-11-19
[DHIS2-11610] fix: convert the orgUnit attribute to server values

simonadomnisoru:


src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js
```diff
@@ -41,7 +41,7 @@ const standardGeoJson = (geometry) => {
 const deriveAttributesFromFormValues = (formValues = {}) =>
     Object.keys(formValues)
         .filter(key => !geometryType(key))
-        .map(key => ({ attribute: key, value: formValues[key] }));
+        .map(key => ({ attribute: key, value: typeof formValues[key] === 'object' ? formValues[key]?.id : formValues[key] }));
```
simonadomnisoru, line 44: I am not sure if this check is enough. Maybe is better to create a convertor and go one by one for each `dataElementTypes`. WDYT?

jasminenguyennn: @simonadomnisoru  that makes sense, not all `dataElementTypes` are handled the same

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js
```diff
@@ -92,13 +92,15 @@ const deriveEvents = ({ stages, enrollmentDate, incidentDate, programId, orgUnit
 export const startSavingNewTrackedEntityInstanceEpic: Epic = (action$: InputObservable, store: ReduxStore) =>
     action$.pipe(
         ofType(registrationFormActionTypes.NEW_TRACKED_ENTITY_INSTANCE_SAVE_START),
-        map(() => {
+        map((action) => {
             const { currentSelections: { orgUnitId, trackedEntityTypeId }, formsValues } = store.value;
             const values = formsValues['newPageDataEntryId-newTei'];
+            const formFoundation = action.payload?.formFoundation;
+            const formServerValues = formFoundation?.convertValues(values, convertToServer);
```
JoakimSM, line 99: This is a good start. We need to convert to client first though and then we can convert to server format.

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js
```diff
@@ -124,18 +126,20 @@ export const completeSavingNewTrackedEntityInstanceEpic: Epic = (action$: InputO
 export const startSavingNewTrackedEntityInstanceWithEnrollmentEpic: Epic = (action$: InputObservable, store: ReduxStore) =>
     action$.pipe(
         ofType(registrationFormActionTypes.NEW_TRACKED_ENTITY_INSTANCE_WITH_ENROLLMENT_SAVE_START),
-        map(() => {
+        map((action) => {
             const { currentSelections: { orgUnitId, programId }, formsValues, dataEntriesFieldsValue } = store.value;
             const { incidentDate, enrollmentDate, geometry } = dataEntriesFieldsValue['newPageDataEntryId-newEnrollment'] || { };
```
JoakimSM, line 131: Not really a part of this issue, but it would be great to fix it anyway. `incidentDate` and `enrollmentDate` should also be converted (formToClient then clientToServer). The reason is that we can display the date in the format DD-MM-YYYY (we have a setting for this), but the format sent to the server should always be YYYY-MM-DD. Doing the conversions should fix the potential problem here.

JoakimSM (commented):
Thanks @simonadomnisoru, good start to this. I've suggested a couple of improvements.

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js
simonadomnisoru: Hey, 
Yes, it makes sense to fix `incidentDate` and `enrollmentDate`. I updated the code with your suggestions. 
Thank you!

JoakimSM (approved):
Thanks, LGTM 👍

PR #2270, opened 2021-11-22
fix: [TECH-669] Validate input params on EnrollmentEventNew

eirikhaugstulen:
Basic validation when opening the `EnrollmentEventNew`-page from the URL.


src/core_modules/capture-core/utils/cachedDataHooks/useValidatedIDsFromCache.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+        if (orgUnitId) {
+            keys.push({ id: orgUnitId, type: IdTypes.ORG_UNIT });
+            promises.push(getCachedSingleResourceFromKeyAsync(userStores.ORGANISATION_UNITS_BY_PROGRAM, programId));
+        }
+
+        Promise.all(promises)
+            .then((adapterResponses) => {
+                if (orgUnitId) {
+                    adapterResponses[1] = adapterResponses[1]?.organisationUnits[orgUnitId];
```
simonadomnisoru, line 39: I am not sure that mutating the `adapterResponses` promise response is the best practice here. Maybe you can have a convertor function inside the `keys` object and apply it. Something like this:
```
keys.push({ id: orgUnitId, type: IdTypes.ORG_UNIT, convert: value => value?.organisationUnits[orgUnitId] });
keys.push({ id: programId, type: IdTypes.PROGRAM, convert: value => value });
```
And it can be used later like this
```
const adapterResponseObjects = keys.map(({ id, type, convert }, index) => ({
     id,
     valid: convert(adapterResponses[index]),
     type,
}));
```
WDYT?

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,126 +1,74 @@
...
-                hideWidgets={hideWidgets}
-                widgetReducerName={widgetReducerName}
-                rulesExecutionDependencies={rulesExecutionDependencies}
-                pageFailure={Boolean(commonDataError)}
-                ready={Boolean(enrollment)}
-                dataEntryHasChanges={dataEntryHasChanges}
+    switch (pageStatus) {
+    case EnrollmentAddEventPageStatuses.LOADING:
+        return <LoadingMaskForPage />;
+    case EnrollmentAddEventPageStatuses.MISSING_REQUIRED_VALUES:
```
simonadomnisoru, line 35: Hey, 
Do you think it makes sense to use the [NoticeBox](https://ui.dhis2.nu/demo/?path=/story/data-display-notice-box--error) component to display all these warnings messages? 
Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.js
```diff
@@ -0,0 +1,127 @@
...
+        handleResetEventId,
+        teiDisplayName,
+        trackedEntityName,
+        enrollmentsAsOptions,
+        teiSelectorFailure,
+        userInteractionInProgress,
+    } = useEnrollmentAddEventTopBar(teiId, programId, enrollment);
+
+    if (stageId && !selectedProgramStage) {
+        return <p style={{ color: 'red' }}>{i18n.t('Program stage is invalid')}</p>;
```
simonadomnisoru, line 83: Same doubt here about the usage of NoticeBox.

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,126 +1,74 @@
...
+    const { valid: validIds, error: validatedIdsError } = useValidatedIDsFromCache({ programId, orgUnitId });
     const {
         enrollment,
         attributeValues,
         error: commonDataError,
     } = useCommonEnrollmentDomainData(teiId, enrollmentId, programId);
+    const pageIsInvalid = (validIds && !validIds?.every(({ valid }) => valid)) || commonDataError || validatedIdsError;
+    const pageStatus = useMemo(() => {
+        if (!programId || !enrollmentId || !teiId) return EnrollmentAddEventPageStatuses.MISSING_REQUIRED_VALUES;
+        if (pageIsInvalid && validIds[0]?.valid && !validIds[1]?.valid) return EnrollmentAddEventPageStatuses.ORG_UNIT_INVALID;
```
simonadomnisoru, line 25: Hey, 
I think the code would be easier to understand if `validIds[0]` and `validIds[1]` were variables with a descriptive name. Currently, you have to dig inside `useValidatedIDsFromCache` to understand what they are. Do you think naming them will improve the code readability?
Thanks!

src/core_modules/capture-core/utils/cachedDataHooks/useValidatedIDsFromCache.js
```diff
@@ -0,0 +1,69 @@
...
+            promises.push(containsKeyInStorageAsync(userStores.PROGRAMS, programId));
+        }
+        if (orgUnitId) {
+            keys.push({ id: orgUnitId, type: IdTypes.ORG_UNIT });
+            promises.push(getCachedSingleResourceFromKeyAsync(userStores.ORGANISATION_UNITS_BY_PROGRAM, programId));
+        }
+
+        Promise.all(promises)
+            .then((adapterResponses) => {
+                if (orgUnitId) {
```
simonadomnisoru, line 38: Is `orgUnit` response always be inside `adapterResponses[1]`. How about the case when `programId` is `null` or `undefined` and there is only one promise to fulfill in the array? 🤔

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.js
```diff
@@ -0,0 +1,127 @@
...
+        handleResetStageId,
+        handleResetEventId,
+        teiDisplayName,
+        trackedEntityName,
+        enrollmentsAsOptions,
+        teiSelectorFailure,
+        userInteractionInProgress,
+    } = useEnrollmentAddEventTopBar(teiId, programId, enrollment);
+
+    if (stageId && !selectedProgramStage) {
```
simonadomnisoru, line 82: Does it make sense to move the `stageId` validation inside `src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js` near the rest of the ids checks?

eirikhaugstulen: Yes, I believe that this would be sensible, although this would require us to have the whole program accessible in `EnrollmentAddEventPage.container.js`. To get the program info we would use the `useProgramInfo`-hook, and this is a hook that we try to avoid in the container. (Since it throws an error if programId is missing or invalid, not a controlled message to the user). Let me know if you want me to take a look at any solutions to this. 👍

src/core_modules/capture-core/utils/cachedDataHooks/useValidatedIDsFromCache.js
eirikhaugstulen: Thanks, I rewrote the hook to match your solution. Works like a charm ☺️

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
eirikhaugstulen: Yep, absolutely agree! Made some changes, hope it's better this time.

src/core_modules/capture-core/utils/cachedDataHooks/useValidatedIDsFromCache.js
eirikhaugstulen: This should not be a problem. The orgUnit-search is in a program-context, so it should fail if a programId is missing or not valid. Either way, this has been changed, so it should not be a worry.

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,126 +1,74 @@
...
-                rulesExecutionDependencies={rulesExecutionDependencies}
-                pageFailure={Boolean(commonDataError)}
-                ready={Boolean(enrollment)}
-                dataEntryHasChanges={dataEntryHasChanges}
+    switch (pageStatus) {
+    case EnrollmentAddEventPageStatuses.LOADING:
+        return <LoadingMaskForPage />;
+    case EnrollmentAddEventPageStatuses.MISSING_REQUIRED_VALUES:
+        return (
+            <p style={{ color: 'red' }}>
```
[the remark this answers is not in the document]
eirikhaugstulen: As suggested, this was changed to using the NoticeBox-component. Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.container.js
```diff
@@ -1,126 +1,74 @@
...
         attributeValues,
         error: commonDataError,
     } = useCommonEnrollmentDomainData(teiId, enrollmentId, programId);
+    const pageIsInvalid = (validIds && !validIds?.every(({ valid }) => valid)) || commonDataError || validatedIdsError;
+    const pageStatus = useMemo(() => {
+        if (!programId || !enrollmentId || !teiId) return EnrollmentAddEventPageStatuses.MISSING_REQUIRED_VALUES;
+        if (pageIsInvalid && validIds[0]?.valid && !validIds[1]?.valid) return EnrollmentAddEventPageStatuses.ORG_UNIT_INVALID;
+        if (pageIsInvalid && !validIds[0]?.valid) return EnrollmentAddEventPageStatuses.PROGRAM_INVALID;
+        if (pageIsInvalid) return EnrollmentAddEventPageStatuses.PAGE_INVALID;
+        if (!validIds?.length) return EnrollmentAddEventPageStatuses.LOADING;
```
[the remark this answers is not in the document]
eirikhaugstulen: Yep, did some rewriting on this. Feel free to have a look! 
Thank you!

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.js
simonadomnisoru: I agree with you, in this case, is best to keep it where it is to avoid the usage of `useProgramInfo` hook in the container. 
Thanks for looking into this!

src/core_modules/capture-core/utils/cachedDataHooks/useValidatedIDsFromCache.js
```diff
@@ -0,0 +1,75 @@
...
+            promises.push(getCachedSingleResourceFromKeyAsync(userStores.ORGANISATION_UNITS_BY_PROGRAM, programId));
+        }
+
+        Promise.all(promises)
+            .then((adapterResponses) => {
+                // $FlowFixMe - Missing declaration for U
+                const adapterResponseObjects = keys.reduce((acc, { id, type, convert }, index) => {
+                    acc[type] = {
+                        id,
+                        valid: convert(adapterResponses[index]),
```
simonadomnisoru, line 45: Hey, 
It looks a bit strange to me to call `adapterResponses[index]` with the `index` coming from the reducer of `keys` array. Can we match `keys` and `adapterResponses` using the `id` instead?

One way to do that is to add the `id` (`orgUnitId` and `programId`) as part of the promises response `adapterResponses`. It can then be used like this `adapterResponses.find(adapterResponse.id===id)`

Does that make sense? Thanks!

eirikhaugstulen: I agree that it's not exactly intuitive. For the programId in this case, it makes sense to perhaps add the key to the `contains`-method in `IndexedDBAdapter`, and then rewrite its usages. For the organization units, it would require a rewrite of the `get`-method. This get-method has different use cases depending on which userStore is passed in. In this case it's the `ORGANISATION_UNITS_BY_PROGRAM`. The method has no knowledge of which orgUnit we want to check for, so the only key passed in is the programId. I'm struggling a bit with finding a solution to how we would implement this for the orgUnitSearch. Do you have any thoughts on this? Do you still think it makes sense with a rewrite?
Thanks!

simonadomnisoru: Good morning,
You don't have to rewrite `contains` or `get` from `IndexedDBAdapter`. You can act one call after by modifying the promises responses in `MetaDataStoreUtils.js`. There you can add whatever extra information you need later on. 

I rise a PR https://github.com/dhis2/capture-app/pull/2270 to your branch to showcase one way you can do that. Let me know if it makes sense or you have doubts. 
Thank you!

simonadomnisoru (approved):

PR #2271, opened 2021-11-23
fix: [DHIS2-10651] Preserve sharing settings backport to v35

jasminenguyennn:
### Issue 
[DHIS2-10651](https://jira.dhis2.org/browse/DHIS2-10651): Preserve sharing settings Backport to v35
Original [PR](https://github.com/dhis2/capture-app/pull/1514)
Since v35 and v36 have too many differences, I can not simply cherry-pick the commit as it brings a lot of unexpected files.

sharing-settings

### DIFF
--- a/src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/templates.epics.js
+++ b/src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/templates.epics.js
@@ -80,27 +80,10 @@ export const updateTemplateEpic = (action$: InputObservable, store: ReduxStore) 
                 userAccesses,
             };
 
-            console.log({ eventFilterData });
             const api = getApi();
 
             const requestPromise = api
                 .update(`eventFilters/${id}`, eventFilterData)
-                // .then(() => api
-                //     .post(`sharing?type=eventFilter&id=${id}`, {
-                //         object: {
-                //             publicAccess: '--------',
-                //             externalAccess: false,
-                //         },
-                //     })
-                //     .catch((error) => {
-                //         log.error(
-                //             errorCreator('could not set sharing settings for template')({
-                //                 error,
-                //                 eventFilterData,
-                //                 templateId: id,
-                //             }),
-                //         );
-                //     }))
                 .then(() => {
                     const isActiveTemplate =
             store.value.workingListsTemplates[listId].selectedTemplateId === id;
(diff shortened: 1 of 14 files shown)


src/core_modules/capture-core/components/Pages/MainPage/WorkingLists/epics/templates.epics.js
```diff
@@ -60,38 +60,35 @@ export const updateTemplateEpic = (action$: InputObservable, store: ReduxStore)
...
-                    .catch((error) => {
-                        log.error(
-                            errorCreator('could not set sharing settings for template')({
-                                error,
-                                eventFilterData,
-                                templateId: template.id,
-                            }),
-                        );
-                    }))
+                .update(`eventFilters/${id}`, eventFilterData)
```
simonadomnisoru, line 86: Hey, 
Can `mutate` here be used here to avoid the usage of `getApi`? Just like it is done in the [original PR](https://github.com/dhis2/capture-app/pull/1514/files#diff-167426afa724a37993f1f3d242f2da27890060a7b6edc91320f907a27dcdb36aR84)
Thanks

jasminenguyennn: hey mutate is not available in v35 and v34. thanks!

simonadomnisoru (approved):

JoakimSM (commented):
Thanks for the PR @jasminenguyennn. I understand this is not straight forward (hence why I didn't back port in the first place). 

On this branch I get in trouble when I'm trying to save a new view (For example: set age in the malaria case registration program and then save current view). Can you look into it? (This is working in the v35 branch)

jasminenguyennn:
thanks @JoakimSM for flagging this issue, I pushed the fix for it

JoakimSM:
I'm still not able to save a new view. It's seems like we are doing a PUT when we should be doing a POST. Can you try digging into it? Thanks

https://user-images.githubusercontent.com/12745737/146742799-c3d3d2da-97c7-441e-b684-b8d7ab2bc472.mov

jasminenguyennn:
@JoakimSM good catch. 
there must be something changed in the v36 `onAddTemplate` which I took from the original [PR](https://github.com/dhis2/capture-app/pull/1514/files#diff-9448a6942d1ee5ace24181d56a4cba62db94ede51556697974bb60f38514c972R54) that cause this issue (POST/PUT). but I revert it, hope it resolve.

JoakimSM (approved):
Thanks! Good work 👍

PR #2285, opened 2021-11-26
fix: [DHIS2-12112] add missing programStage rules

JoakimSM:
Program rules restricted to a specific program stage were not included in the program rules execution.

This only fixes the new enrollment event page. Will fix the edit page in a different ticket.

### DIFF
--- a/src/core_modules/capture-core/metaData/Program/ProgramStage.js
+++ b/src/core_modules/capture-core/metaData/Program/ProgramStage.js
@@ -4,6 +4,7 @@
 /* eslint-disable no-restricted-syntax */
 
 import isFunction from 'd2-utilizr/lib/isFunction';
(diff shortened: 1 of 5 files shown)


simonadomnisoru (approved):
Hey @JoakimSM ,
The fix looks good to me! Out of curiosity, do you remember why the program rules were originally stored inside `RenderFoundation` and not inside `ProgramStage`?
Thanks!

JoakimSM:
> Hey @JoakimSM ,
> The fix looks good to me! Out of curiosity, do you remember why the program rules were originally stored inside `RenderFoundation` and not inside `ProgramStage`?
> Thanks!

Hey @simonadomnisoru,

In the very early days we didn't have a ProgramStage class at all. It seems like we forgot to move the class prototype methods at some point.

PR #2308, opened 2021-12-03
chore: [DHIS2-12142] fix profile tei

jasminenguyennn:
### Issue 
[DHIS2-12142](https://jira.dhis2.org/browse/DHIS2-12142):  Fix widget profile values
- Display TEI optionSet name instead of code
- Get subvalue in case the valueType is among special groups (ORG_UNI, IMG, FILES)

A screenshot example:

Screenshot 2021-12-03 at 07 31 52


src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -54,25 +56,46 @@ export const WidgetProfile = ({ teiId, programId }: Props) => {
...
-                    key: curr.trackedEntityAttribute.displayName,
-                    value: formatValue(foundAttribute?.value, foundAttribute?.valueType),
-                });
+                setListAttributes(
+                    await programTrackedEntityAttributes
+                        .filter(item => item.displayInList)
+                        .reduce(async (promisedAcc, { trackedEntityAttribute: { id, displayName, optionSet } }) => {
+                            const foundAttribute = attributes
+                                .find(item => item.attribute === id);
+                            let value;
```
simonadomnisoru, line 72: Hey, 
I feel that the complexity of this `reduce` is high. Can the whole `if-else` logic where you set the `value` be extracted to another function? 
Thanks!

src/core_modules/capture-core/components/WidgetProfile/getSubValueForTei.js
```diff
@@ -0,0 +1,20 @@
...
+import { dataElementTypes } from '../../metaData';
+
+const getImageOrFileResourceSubvalue = async (key: string) => {
+    const { displayName: name } = await getApi().get(`fileResources/${key}`);
+    return name;
+};
+
+
+const getOrganisationUnitSubvalue = async (key: string) => {
+    const { organisationUnits = [] } = await getApi().get(`organisationUnits?filter=id:in:[${key}]`);
```
simonadomnisoru, line 12: Hey, 
The `setListAttributes` and `getOrganisationUnitSubvalue` are used inside a React component. I think that in this case the DHIS2 hook [useDataQuery](https://runtime.dhis2.nu/#/hooks/useDataQuery) can be used instead of deprecated `getApi`. Do you see any reason why the hooks cannot be used? 
Thanks!

jasminenguyennn: hey, I prefer to separate this sub value logic in different file to make it easier to control the element type. `getOrganisationUnitSubvalue` is called conditionally which is not suitable for using hook (even when using `lazy: true`

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
jasminenguyennn: hey, I feel the complexity here is valid (eslint doesn't complain) and moving to other function may make the flow more difficult to follow

JoakimSM (commented):
Seems to be working well 👍 I would like us to change a couple of things before we merge this though. See comments.

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -54,25 +56,45 @@ export const WidgetProfile = ({ teiId, programId }: Props) => {
...
+                            key: displayName,
+                            value,
+                        }];
+                    }, Promise.resolve([]));
+
+                setListAttributes(computedAttributes);
+            }
+        };
+
+        getListAttributes();
```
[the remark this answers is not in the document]
jasminenguyennn: thanks @JoakimSM, I made the changes based on your suggestion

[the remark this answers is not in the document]
simonadomnisoru: Hey,
You're right @JoakimSM, the subvalues are needed in the edit mode and it makes sense to compute them only once. 

I suggest we aim to keep the `WidgetProfile.component.js` as clean as possible. The logic related to the subValues (`getListAttributes` callback together with its `useEffect` call) can be abstracted and moved in a hook. The hook can be used inside the `WidgetProfile.component.js` like this: 
`const listAttributes = useTrackedEntityAttributes(...)`;

WHYT @jasminenguyennn and @JoakimSM? Thank you!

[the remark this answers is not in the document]
jasminenguyennn: @simonadomnisoru absolutely, I made the changes to extract logic together in a hook. Thanks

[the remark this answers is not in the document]
simonadomnisoru: Hey @jasminenguyennn, 
I still see the `getListAttributes` logic inside the `WidgetProfile`. Did I miss the latest commits?  
Thank you!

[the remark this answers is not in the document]
jasminenguyennn: @simonadomnisoru that's right some commits were missing, should be updated now

simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -48,31 +49,22 @@ export const WidgetProfile = ({ teiId, programId }: Props) => {
     const loading = programsLoading || trackedEntityInstancesLoading;
     const error = programsError || trackedEntityInstancesError;
 
+    const listAttributes = useListAttributes(programsData, trackedEntityInstancesData);
```
JoakimSM, line 52: NIT: Maybe `useClientAttributesWithSubvalues` and `clientAttributesWithSubvalues` (or something that doesn't imply that this has something to do with displayInList)

src/core_modules/capture-core/components/WidgetProfile/hooks.js
```diff
@@ -1,7 +1,86 @@
...
+    const [listAttributes, setListAttributes] = useState([]);
+
+    const getListAttributes = useCallback(async () => {
+        if (programsData && trackedEntityInstancesData) {
+            const querySingleResource = makeQuerySingleResource(dataEngine.query.bind(dataEngine));
+
+            const { programs: { programTrackedEntityAttributes } } = programsData;
+            const { trackedEntityInstances: { attributes } } = trackedEntityInstancesData;
+            const computedAttributes = await programTrackedEntityAttributes
+                .filter(item => item.displayInList)
```
JoakimSM, line 49: We should remove the filtering here (if I'm not missing something)

JoakimSM (commented):
Minor comments. Looks good overall 👍

src/core_modules/capture-core/components/WidgetProfile/hooks.js
JoakimSM: You can absolutely make a case for keeping this filtering and letting the edit mode retrieve the values that are not displayInList. But I suggest we do it like this for now, we can improve later if needed.

JoakimSM (approved):

PR #2326, opened 2021-12-08
feat: [DHIS2-5345] implement data element description and url popover

jasminenguyennn:
### Feature

[DHIS2-5345](https://jira.dhis2.org/browse/DHIS2-5345): Display description and URL for data element
After pulling the data from API
1) add the description and url to the IndexedDB cache
2) add the description and url to metadata objects in memory
3) add support in the D2Form components for showing the description / url.

Screenshot 2021-11-30 at 15 35 54


simonadomnisoru (approved):
Great job! LGTM!

src/core_modules/capture-core/components/FormFields/New/HOC/withDataElementDescription.js
```diff
@@ -0,0 +1,84 @@
...
+                        { url ? <div className={classes.popOverContainer}>
+                            <div className={classes.label}>{i18n.t('URL')}</div>
+                            <a href={url} target={'_blank'} rel="noreferrer">{url}</a>
+                        </div> : null}
+                    </Popover>}
+                </div>);
+            }
+
+            render() {
+                const { dataElement, classes, ...passOnProps } = this.props;
```
JoakimSM, line 68: Might be of interest to others in the future, so writing my suggestion here (we started talking about this on Slack). We need to make some changes for this to work in every form. 

1. The data element should not be included in the input in this context (it is in the view event page currently, but shouldn't. In other forms it is missing). Therefore, we should include description and url directly like we do for the label (flat structure).
2. I think it makes sense to add the properties here: https://github.com/dhis2/capture-app/blob/e4a50693a337607969671e2599e4e95730b88d06/src/core_modules/capture-core/components/D2Form/field/configs/base/configBaseDefaultForm.js#L28. (Didn't check this thoroughly, just a hunch).

Great if you verify that this is working in the new single event form.

NIT: Maybe rename the file to 'withDescription'.

JoakimSM (commented):
This looks good @jasminenguyennn 👍 You followed the pattern that was already established in this part of the code, I think that makes sense.

One change request in comment.

src/core_modules/capture-core/components/FormFields/New/HOC/withDataElementDescription.js
jasminenguyennn: thanks @JoakimSM for the suggestion. You are indeed right, the data element is not defined in some cases so I move to use the base props as suggested.

JoakimSM (approved):
Thanks, LGTM 👍

PR #2327, opened 2021-12-08
refactor: showcase promises response 

simonadomnisoru:


eirikhaugstulen (approved):
Thanks for the great solution, @simonadomnisoru!

PR #2332, opened 2021-12-09
fix: [DHIS2-12139] add programId to navigation

jasminenguyennn:
### Issue
[DHIS2-12139]():
Navigate to capture app when clicking `View Dashboard`, I believe the programId was missing in the url so the `navigateToEnrollmentOverviewEpic` can not function properly


eirikhaugstulen (approved):
LGTM!

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.component.js
```diff
@@ -102,6 +104,7 @@ const RegistrationDataEntryPlain = ({
         <CardListButton
             teiId={item.id}
             orgUnitId={item.tei.orgUnit}
+            programId={item.tei.programOwners[0].program}
```
simonadomnisoru, line 107: Hey,
Why did you choose the first item from the `programOwners` array? The `TEI` can be owned by different programs therefore I think that this logic can fail in some cases.  

As far as I know, there are 2 possibilities:
1. The user selected a program, then the `programId` will be in the URL and it can be taken from there. 
2. The user didn't select a program, so I don't think any `programId` should be passed to the `onClick` callback because we don't know it.  

Thanks!

jasminenguyennn: hey @simonadomnisoru, since this is the modal opened on top on the page I don't know if taking the programId from url is good idea.  These are the results from the deduplication so that's why I'm taking the data information from `item` but I'm not sure about the first item from `programOwners`.
it would be great to have your thought here @JoakimSM

JoakimSM: Hey, yes we should use the program context from the URL as @simonadomnisoru suggested here. Great if you can update to use the suggested logic @jasminenguyennn 🙂

jasminenguyennn: thanks @JoakimSM @simonadomnisoru  for clarifying. I took programId from url and moved it to CardButton

simonadomnisoru (approved):

PR #2333, opened 2021-12-09
fix: [TECH-870] Stabilise and improve rules execution

JoakimSM:
Addressed the main changes through comments in the PR.


src/core_modules/capture-core-utils/rulesEngine/services/VariableService/variableService.types.js
```diff
@@ -0,0 +1,95 @@
...
+    programRuleVariableSourceType: string,
+    programId: string,
+    dataElementId?: ?string,
+    trackedEntityAttributeId?: ?string,
+    programStageId?: ?string,
+    useNameForOptionSet?: ?boolean,
+};
+
+type EventMain = {
+    +eventId?: string,
```
jasminenguyennn, line 17: is the `+` here not necessary ?

src/core_modules/capture-core/rules/converters/dateUtils.js
```diff
@@ -47,6 +47,7 @@ class DateUtils implements IDateUtils {
         return `'${newRulesDate}'`;
     }
     compareDates(firstRulesDate: string, secondRulesDate: string): number {
+        secondRulesDate = moment().toISOString();
```
simonadomnisoru, line 50: Why is `secondRulesDate` value overwritten here? Maybe can be removed from the functions parameters `compareDates(firstRulesDate: string, secondRulesDate: string)` 
Thanks!

src/core_modules/capture-core/components/WidgetEventEdit/DataEntry/epics/editEventDataEntry.epics.js
```diff
@@ -47,44 +50,51 @@ export const openEditEventInDataEntryEpic = (action$: InputObservable, store: Re
...
         }));
 
-
 const runRulesForEditSingleEvent = (store: ReduxStore, dataEntryId: string, itemId: string, uid: string, fieldData?: ?FieldData) => {
     const state = store.value;
     const formId = getDataEntryKey(dataEntryId, itemId);
     const eventId = state.dataEntries[dataEntryId].eventId;
-    const event = state.events[eventId];
-    const { programId } = state.currentSelections;
+    const event = state.events[eventId];  // TODO: Refactor
```
simonadomnisoru, line 57: There are a couple of `TODOs` in this file. I feel that it would be helpful if there is more context given in the comments or if a Jira ticket number is mentioned. Thanks!

src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/newEventDataEntry.epics.js
```diff
@@ -87,6 +87,11 @@ export const openNewEventInDataEntryEpic = (action$: InputObservable, store: Red
             return page === 'new';
         }),
         filter((action) => {
+            const { programId } = store.value.currentSelections;
+            const program = programCollection.get(programId);
```
simonadomnisoru, line 91: Hey, 
I think that this logic is already in `getEventProgramThrowIfNotFound` and the function can be called here. WDYT?
Thanks!

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/hooks/useRuleEffects.js
```diff
@@ -0,0 +1,93 @@
+// @flow
```
JoakimSM, line 1: Refactored the rules engine execution for the enrollment page and added missing parts and data conversions.

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/helpers/structureEvents.js
```diff
@@ -0,0 +1,44 @@
+// @flow
```
JoakimSM, line 1: Added some missing parts to the rules engine variable service. Also, some common logic that earlier resided outside of the rules engine, I moved into he engine itself.

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/variableService.types.js
JoakimSM: Thanks for the comment. The idea here is to signalise to flow that we will not make changes to the objects being passed in here (i.e, that the objects are readonly in the scope of the variable service). This helps us avoid flow errors when we pass in objects that are annotated slightly differently. For example, say that the eventId is mandatory for the objects that are passed in, but the definition in the variable service says that it is optional; in this instance flow would complain (without the +) because we can potentially remove the eventId from the object inside the variable service (and cause the annotation for the objects that are passed in to be invalid). 

Sorry for the long answer, hope it makes sense.

src/core_modules/capture-core/rules/converters/dateUtils.js
JoakimSM: Good catch! I was thinking about refactoring it, but didn't complete the job. Reverting the change here.

src/core_modules/capture-core/components/WidgetEventEdit/DataEntry/epics/editEventDataEntry.epics.js
JoakimSM: Improved the comments. Thanks for the input.

src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/newEventDataEntry.epics.js
JoakimSM: Thanks for the comment. In this code block the program can be either tracker or event and it doesn't feel right to throw if it is a tracker program (i.e., it's not an exception to get a tracker program here). I would reserve exception throwing for cases where we didn't get what we expected.

(and yes this code needs some refactoring also, but that's for another day)

simonadomnisoru (approved):

PR #2362, opened 2021-12-18
chore: [TECH-686] Tests for EnrollmentEventNewForm

eirikhaugstulen:
Added tests for checking form & rules functionality on the EnrollmentEventNew-page. 
Please comment if there is anything else we should test.


simonadomnisoru (approved):
Hey,
Great job here! I like seeing PRs that target our technical debt. 👏

Nit: The Scenarios `Complete` and `Save` add new data every time they run. Ideally, to avoid flaky tests, the Cypress scenarios should not leave permanent effects on the database. Unfortunately, without the "Delete event" feature, the data cannot be cleaned at the moment. Let's keep this in mind and maybe come back to these tests when the feature is added. 

Thank you!

PR #2364, opened 2021-12-18
fix: [DHIS2-12126] Capture crash when navigating back from event with CategoryOption

eirikhaugstulen:


jasminenguyennn (approved):
👍

simonadomnisoru (approved):
LGTM!

PR #2374, opened 2021-12-23
chore: [TECH-808] add ouMode backport to v36

jasminenguyennn:


jasminenguyennn:
Failing tests are related to BE changes in 36 (probably similar to 37), previously we have to disable the tests in v37 but I'm not sure what to do in this case

simonadomnisoru (approved):
LGTM! 
The 10 failures in Cypress are [known issues](https://github.com/dhis2/capture-app/wiki/Pull-Request-Guidelines#cypress-tests-to-follow-up-on-the-tests-are-commented-out-from-our-code-base) therefore I don't think we have to worry about them in this PR.

PR #2380, opened 2021-12-28
fix: [DHIS2-12247] fix render loop in enrollment page

jasminenguyennn:


simonadomnisoru (commented):
Hey @jasminenguyennn,

I am still seeing render loop warnings when the organization unit is selected/deselected from the top bar. Can you investigate if there is a fixed? 

Thank you!

jasminenguyennn:
hey @simonadomnisoru, I also noticed that warning but the issue is coming from `OrganisationUnitTree` in ui library. I'm talking with the frontend team to find a solution. Thank you

jasminenguyennn:
@simonadomnisoru according to FE team, this is unfortunately the way app-runtime works right now (more context in the thread [here](https://dhis2.slack.com/archives/C0BP0RABF/p1641545855004500?thread_ts=1641544953.003000&cid=C0BP0RABF)). So I'm afraid we have to deal with the warning from OrganisationUnitTree and can only solve the one from our Enrollment page

simonadomnisoru (approved):

JoakimSM (approved):

PR #2382, opened 2021-12-28
feat: [TECH-902] Use new tracker endpoints in new page

eirikhaugstulen:
Switched to new tracker endpoints in the new page. This includes events, TEIs and TEIs with enrollment. 

Note: the GET-`tracker/relationships` functionality needs to be revisited.


jasminenguyennn (approved):

simonadomnisoru:
Hey @eirikhaugstulen ,
Are the failed Cypress tests related to the `GET-tracker/relationships` functionality?
Thanks!

eirikhaugstulen:
Hey @simonadomnisoru!
Most of the tests are failing because they're using the classes which have been changed so far, and will therefore be fixed in upcoming PR's. There are also some tests that are failing because of missing functionality or bugs in the new tracker endpoints. I'm making a list of these breaking changes that we need to review with both FE & BE. I would also prefer to not comment out these test, but rather keep them in to remind us of missing functionality. (These would have to be fixed before merging to master, but I don't see that happening anytime soon)
Does this make sense? 
Thanks!

simonadomnisoru (approved):

PR #2383, opened 2021-12-31
feat: [TECH-904] Use new tracker endpoints in enrollment overview

eirikhaugstulen:


simonadomnisoru (commented):
Hey @eirikhaugstulen,

I am not able to mark the enrollment for a followUp. The `POST` request returns `200` but the enrollment status is not correctly updated afterward. Is this a regression bug introduced by the new endpoints?
Thank you!

eirikhaugstulen:
Hey @simonadomnisoru!
Yes, this is something that I've noted down. The request goes through and the API responds with 200 and that one enrollment is updated. Yet, when getting the supposedly updated enrollment, this is not the case. Even tried with different values in postman, so I don't think this is a frontend-issue. 
Thanks!

simonadomnisoru (approved):

jasminenguyennn (approved):

PR #2390, opened 2022-01-03
fix: cypress test year change

simonadomnisoru:
The test was based on a hardcoded year value. It started falling after the 1st of January.


jasminenguyennn (approved):
👍  good catch

PR #2408, opened 2022-01-07
fix: [DHIS2-12172] missing org unit code

simonadomnisoru:
fix the missing org unit code.
The UI organization-unit-tree needs a change in other to access the code when selecting an org unit. I started a discussion about it https://github.com/dhis2/notes/discussions/322


simonadomnisoru:
The code was moved to another branch https://github.com/dhis2/capture-app/pull/2432

PR #2416, opened 2022-01-11
feat: [DHIS2-12301] enable default enrollment dashboard

simonadomnisoru:
- Enable default navigation to the enrollment dashboard from the main, new, and search pages. 
- Fix some cypress failure from LockedSelector.feature, NewPage.feature, SearchPage.feature

Change the dataStore key from `useNewDashboard` to `useOldDashboard`. The user can reach the old Tracker if they set the `programId` true in the `useOldDashboard`. Steps to enable the old tracker navigation:

1. Make a POST-request to `/api/dataStore/capture/useOldDashboard`
2. Make API-body JSON with `programId`: `true/false`
3. The capture app should then navigate the user to the old tracker if the program is set to true.


src/core_modules/capture-core/reducers/descriptions/app.reducerDescriptionGetter.js
```diff
@@ -69,7 +69,7 @@ export const getAppReducerDesc = (appUpdaters: Updaters) => createReducerDescrip
     },
     [enrollmentPageActionTypes.PAGE_OPEN]: (state) => {
         const newState = { ...state };
-        newState.page = null;
+        newState.page = 'enrollment';
```
jasminenguyennn, line 72: hey @simonadomnisoru , is there a reason why this has changed to "enrollment", I don't see it's used anywhere but seeing other places they are all `null`

simonadomnisoru: Hey @jasminenguyennn , 
When I enabled the direct navigation to the enrollment dashboard some browser history bugs appeared that were not visible before. When the enrollment page opens (in this case on `enrollmentPageActionTypes.PAGE_OPEN`), the page should be set to `enrollment` not `null`.  This change fix the previous undetected errors. 
Thanks!

jasminenguyennn (approved):
👍

JoakimSM (approved):
LGTM

PR #2368, opened 2021-12-21
fix: [DHIS2-11837] Saving message for compulsory fields

jasminenguyennn:
### Issue 
[DHIS2-11837](): users can save form without filling compulsory fields
This happens when `typesToCheck === typeBase` and `fieldUI.errorType === null`, so adding check if errorType is present 

Clean up form errors including 
- Unnecessary onSave event
- Unnecessary `fieldsValidated` on input
- Fix typo

PR #2421, opened 2022-01-11
fix: [DHIS2-11837] Saving message for compulsory fields (#2368) backporting to v37

jasminenguyennn:
* fix: [DHIS2-11837] validate compulsory fields

* fix: [DHIS2-11837] unnecessary onSave

* fix: [DHIS2-11837] fix is completing

* chore: [DHIS2-11837] use saveTypes


simonadomnisoru (approved):
LGTM!

PR #2430, opened 2022-01-15
fix: [DHIS2-12324] Change label to enrollmentDate in ScopeSelector

eirikhaugstulen:
Change enrollment-label from created to enrollmentDate

### DIFF
--- a/cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageForm.feature
+++ b/cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageForm.feature
@@ -1,41 +1,41 @@
-Feature: User interacts with Enrollment event page view/edit form 
+Feature: User interacts with Enrollment event page view/edit form
 
 Scenario: The user can view the sections, the labels and the values in the form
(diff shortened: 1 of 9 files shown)


JoakimSM (approved):
LGTM

simonadomnisoru (commented):
Hey @eirikhaugstulen,

The changes look good to me. It's awsome that you also fixed the other date-related Cypress failures. 👏
I just have one suggestion. Does it make sense to also change the `buildEnrollmentsAsOptions` in the [LockedSelector](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/LockedSelector/QuickSelector/QuickSelector.container.js#L18). In theory, that function is never run because `LockedSelector` is not used in the enrollments pages. Therefore this change is more to maintain homogeneity between the `ScopeSelector` and `LockedSelector`. WDYT? 
Thank you!

eirikhaugstulen:
Hey @simonadomnisoru,
Yes, that makes sense! Should be updated now.

simonadomnisoru (approved):

PR #2435, opened 2022-01-17
fix: [DHIS2-11280] add default program value

jasminenguyennn:
### Issue

[DHIS2-11280](https://jira.dhis2.org/browse/DHIS2-12280): Program not preselected when registering TEI through creation of Relationship

- Add programId as fallback value in `ProgramSelector`
- Testing is not possible right now, we will need to set up the relationship type in DB and change the share settings.
Screenshot 2022-01-17 at 13 03 45

### How to test
In maintenance app
- Change relationship type share settings to **Can capture and view**
Screenshot 2022-01-17 at 13 24 03

- Assign a program in relationship type
Screenshot 2022-01-17 at 13 23 52


simonadomnisoru (approved):
LGTM!

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/RegistrationSection/ProgramSelector/ProgramSelector.container.js
```diff
@@ -6,7 +6,7 @@ import { changeProgram, clearProgramFilter } from '../registrationSection.action
 const mapStateToProps = (state: ReduxState) => ({
     orgUnitIds: state.newRelationshipRegisterTei.orgUnit ? [state.newRelationshipRegisterTei.orgUnit.id] : null,
     trackedEntityTypeId: state.newRelationship.selectedRelationshipType.to.trackedEntityTypeId,
-    value: state.newRelationshipRegisterTei.programId,
+    value: state.newRelationshipRegisterTei.programId ?? state.newRelationship?.selectedRelationshipType?.to?.programId,
```
JoakimSM, line 9: We should try to figure out why state.newRelationshipRegisterTei.programId is not updated instead, due to a couple of reasons: a) The programId we are looking for could be on the "from" side (this workflow is not working properly at the moment, created JIRA issue for this) b) There is some logic that relies on `state.newRelationshipRegisterTei.programId` being set.

JoakimSM (commented):
Hey @jasminenguyennn , See comment 🙂

src/core_modules/capture-core/components/Pages/NewRelationship/RegisterTei/RegistrationSection/ProgramSelector/ProgramSelector.container.js
jasminenguyennn: thanks @JoakimSM I agree. on the second look I found that the programId is set to null here https://github.com/dhis2/capture-app/pull/2435/files#diff-182458f2805a717c67f38531199e47402242a16c155da4562cd53071ec1b15beL75. 
So putting the programId there should fix it

JoakimSM (approved):
LGTM 👍

PR #2355, opened 2021-12-16
feat: [DHIS2] New tracker endpoints in capture

eirikhaugstulen:
Temporary PR to collect epic of replacing the tracker endpoints

PR #2444, opened 2022-01-20
feat: [DHIS2-10950] update tracked entity attributes 

simonadomnisoru:
- validate the data elements, the unique fields as well as the program rules.
- make `POST` requests with the TEI attributes values.
- Loading state while we wait for a response
- handle API errors
- After TEI attributes are successfully updated:
       1. Close the edit modal
       2. use callbacks to update the TEI displayName in the `ScopeSelector` 
       3. use callbacks to update the attributes in the `WidgetProfile`
       4. use callbacks to retrigger the enrollment rules with the new attributes


src/core_modules/capture-core/components/WidgetProfile/DataEntry/dataEntry.epics.js
```diff
@@ -0,0 +1,87 @@
...
+                trackedEntityTypeId,
+                trackedEntityInstanceId,
+                onSaveExternal,
+                onSaveSuccessActionType,
+                onSaveErrorActionType,
+            } = action.payload;
+
+            const values = formsValues[`${dataEntryId}-${itemId}`];
+            const serverData = {
+                attributes: deriveAttributesFromFormValues(values),
```
simonadomnisoru, line 60: This is the same code we use when we create a new TEI. The values are not properly converted for server format. There is [another PR opened](https://github.com/dhis2/capture-app/pull/2267) to fix this for new TEI. Once that PR is merged I will update the code here also.

src/core_modules/capture-core/components/WidgetProfile/DataEntry/dataEntry.actions.js
```diff
@@ -73,6 +98,68 @@ export const getUpdateFieldActions = (context: Context, innerAction: ReduxAction
...
+    onSaveSuccessActionType?: string,
+    onSaveErrorActionType?: string,
+    uid: string,
+}) =>
+    actionCreator(dataEntryActionTypes.TEI_UPDATE)(
+        {},
+        {
+            offline: {
+                effect: {
+                    url: 'trackedEntityInstances',
```
simonadomnisoru, line 154: Reminder to change the URL to the [New tracker endpoints](https://github.com/dhis2/capture-app/pull/2355/files).

JoakimSM (commented):
Been doing some thinking on this one and it would be great if we can approach it a bit differently.

I would like us to implement a more generic solution for using the updated attribute values elsewhere in the app. It would also be great to only use the Redux Store when we «have to». The current reasons we «have to»:
- The current offline framework requires us to dispatch actions. Therefore, we might need to utilise the Redux store in some capacity when saving.
- The old dataEntry / form component uses the Redux store.

A third point is that we currently wait for the saving to complete before closing the modal/update the values elsewhere; i.e, no optimistic UI. I’m fine with this for now, we can improve on this point later (there is some added complexity for rollbacks and so on)

So, my suggestion is to use callbacks for updating the attributes elsewhere (elsewhere = the view part of the profile Widget, the enrollmentDomain of the Redux store, The TEI selector at the top). Since we are using the offline framework, this becomes a bit tricky, but should be doable. I would also prefer to update the values without refetching.

We might need a chat about this one.

JoakimSM:
The functionality for updating tracked entity attributes seems to work well 👍 One minor problem: A different set of attributes are shown at the TEI selector at the top after updating (mine changed from "Sharon Larson" to "5668 Sharon"). Can you investigate?

One last thing: Can we try to get rid of the TEI modal state in the Redux store? Try to keep what we need in the Components instead.

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -54,6 +58,17 @@ const WidgetProfilePlain = ({ teiId, programId, showEdit = false, orgUnitId = ''
...
 
+    useEffect(() => {
+        hasError && setTeiModalState(TEI_MODAL_STATE.OPEN_ERROR);
+    }, [hasError]);
+
+    useEffect(() => {
+        if (storedAttributeValues?.length > 0) {
+            setTeiModalState(TEI_MODAL_STATE.CLOSE);
+            onUpdateTeiAttributeValues && onUpdateTeiAttributeValues(storedAttributeValues, teiDisplayName);
+        }
```
JoakimSM, line 69: We will need to do cleanup here. If not, values can/will be overwritten when we open another tei.

JoakimSM (commented):
Getting very close now 🙂

JoakimSM (approved):
🎉

PR #2446, opened 2022-01-21
fix: [DHIS2-12134] fix pending schedule event

jasminenguyennn:
### Issue 

[DHIS2-12134](https://jira.dhis2.org/browse/DHIS2-12134): New and scheduled events don't show up in the Stages and Events Widget


simonadomnisoru (commented):
Hey @jasminenguyennn,

Sometimes the event still appears as pending. After I refresh the page the event is displayed correctly. 

Screenshot 2022-01-24 at 09 23 01

Screenshot 2022-01-24 at 09 23 19
 
Thank you!

simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.epics.js
```diff
@@ -37,7 +37,7 @@ export const scheduleNewEnrollmentEventEpic = (action$: InputObservable) =>
             }] };
 
             onSaveExternal && onSaveExternal(serverData, uid, onSaveSuccessActionType, onSaveErrorActionType);
```
JoakimSM, line 39: I know you didn't add this here, but we can safely remove the last two arguments here.

JoakimSM (commented):
What you changed makes sense for Scheduled events 👍

I did some testing though, and from time to time, my new event gets stuck as pending in the list. Seems like I can only reproduce it when my debugger is closed, but even with the debugger closed it only happens once in a while. My hunch is that the first render of the list works fine, but there is trouble when we re-render the list?

Can you investigate?

jasminenguyennn:
thanks @JoakimSM. I changed the approach to compute data in the event list.

JoakimSM (approved):
Seems to work well now 👍

PR #2449, opened 2022-01-23
feat: [TECH] Cypress component testing

eirikhaugstulen:
This is a suggestion to how we could solve the component-testing that we have talked about the last couple of months. The cypress component testing-feature is still in Alpha, so it might change as it is developed, but with my limited testing it seems rigid enough for our use case. 

Please let me know if I should rename `__cytests__`. This was just so it didn't interfere with Jest.


simonadomnisoru (commented):
Hey @eirikhaugstulen, 
Does it make sense to add a script in `package.json` to open/run the component tests? Maybe some like this: 
`cy:openComponentTesting": "concurrently --kill-others \"yarn start:forCypress\" \"wait-on 'http-get://localhost:3000' && cypress open-ct\`

If can also be added as a step in the [Github checks flow](https://github.com/dhis2/capture-app/blob/master/.github/workflows/verify-app.yml#L92) 😄

eirikhaugstulen:
Will be superseded by new PR if picked up again.

PR #2456, opened 2022-01-25
feat: [TECH-893] Ensure org unit updated in redux

jasminenguyennn:
### Issue 
[TECH-893](https://jira.dhis2.org/browse/TECH-893): Ensure organisation units in Redux store are updated

Changing the approach to wrap the components with `OrgUnitFetcher` and wait until the redux updated


src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/NewEventDataEntryWrapper.component.js
```diff
@@ -62,45 +66,47 @@ const NewEventDataEntryWrapperPlain = ({
...
 
     return (
-        <Paper className={classes.paper}>
-            <div className={classes.title} >
-                {i18n.t('New {{titleText}}', {
-                    titleText,
-                    interpolation: { escapeValue: false },
-                })}
-            </div>
+        <OrgUnitFetcher orgUnitId={orgUnitId}>
```
simonadomnisoru, line 69: Hey @jasminenguyennn,

In `EnrollmentEditEventPage` and `MainPage` the `OrgUnitFetcher` is used at page level. I like this approach because it ensures we have the data updated in the whole page context. 
Why is the logic for `NewPage` different and you use the `OrgUnitFetcher` inside the `NewEventDataEntryWrapper` and not at the page level? 🤔

Does it make sense to use the `OrgUnitFetcher` inside `LockerSelector` and `ScopeSelector`? All the pages use one of these selectors and I think is easier to maintain if is just in only 2 places. The `ScopeSelector` [already has a hook to fetch the orgUnit](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/ScopeSelector/ScopeSelector.container.js#L46) if its name is missing or different. I think you can replace it with your logic from `OrgUnitFetcher`. 

Another suggestion is to dig into `AppPages` and `AppPagesLoader` and use `OrgUnitFetcher` there. If the strategy for the future is to get rid of `withAppUrlSync` this may be the most suitable approach. 

Let me know what you think. Thank you!

jasminenguyennn: hey @simonadomnisoru, I put the `OrgUnitFetcher` here mainly because it's used [here](resetDataEntryForNewEventEpic) in the epic. If the `LockerSelector` and `ScopeSelector` already did the job then the check in the DataEntry is redundant but I'm afraid it's not the case since we have a comment [here](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetEventEdit/DataEntry/epics/editEventDataEntry.epics.js#L35) in `WidgetEventEdit` that it may fail. 

The goal is the move away completely from redux but right now it's not possible because of the `LockedSelector`/`ScopeSelector` so that's why I have to make an intermediate solution.

I'm not sure fetching in `AppPages` and `AppPagesLoader` is a good solution since depending on each use case we want to fetch orgUnit from currentSelection or the url, and it's not evident in the top level to do so. 

Let me know if that makes sense. Thank you

simonadomnisoru (approved):

JoakimSM (commented):
This looks good @jasminenguyennn. One question: Do you think it makes sense to also wrap the "New Single Event Page" and the "Edit Single Event Page"? I see that you wrapped the file NewEventDataEntryWrapper (part of the new single event page), but I would be more confident of this fix if we also wrapped the mentioned pages closer to the top / entry level.

JoakimSM (approved):
LGTM 👍

PR #2478, opened 2022-01-31
feat: [DHIS2-12455] Filter & list applicable relationship types in new TEI relationship handler

eirikhaugstulen:


src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/NewTrackedEntityRelationship.component.js
```diff
@@ -0,0 +1,51 @@
...
+import { withStyles } from '@material-ui/core';
+import { NewTEIRelationshipStatuses } from '../WidgetTrackedEntityRelationship.const';
+import { RelationshipTypeSelector } from './RelationshipTypeSelector/RelationshipTypeSelector';
+
+const styles = {
+    container: {
+        padding: spacers.dp16,
+        paddingTop: 0,
+    },
+    creationselector: {
```
jasminenguyennn, line 13: nit: camelCase

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/RelationshipTypeSelector/RelationshipTypeSelector.js
```diff
@@ -0,0 +1,132 @@
...
+import { Button, spacers } from '@dhis2/ui';
+import React, { useMemo } from 'react';
+import { withStyles } from '@material-ui/core';
+
+const styles = {
+    container: {
+        padding: spacers.dp16,
+        paddingTop: 0,
+    },
+    typeselector: {
```
jasminenguyennn, line 11: nit: camelCase

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/NewTrackedEntityRelationship.container.js
```diff
@@ -0,0 +1,103 @@
...
+    if (!showDialog || !renderRef.current) {
+        return null;
+    }
+
+    return ReactDOM.createPortal((
+        <>
+            <div className={classes.container}>
+                <div className={classes.bar}>
+                    <LinkButton onClick={onCancel} className={classes.linkText}>
+                        Go back without saving relationship
```
simonadomnisoru, line 82: Hey, can you add the `i18n.t`? Thanks!

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/hooks.js
```diff
@@ -0,0 +1,66 @@
...
+    const [loading, setLoading] = useState(true);
+    const error = useRef();
+    useMemo(() => {
+        getCachedResourceAsync(userStores.RELATIONSHIP_TYPES)
+            .then((relationshipTypes) => {
+                setCachedTypes(relationshipTypes?.response);
+                setLoading(false);
+            })
+            .catch((e) => {
+                console.error(e);
```
simonadomnisoru, line 20: Hey, 
Can you `log.error(...)` a meaningful error here? 
Thanks!

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/NewTrackedEntityRelationship.component.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import i18n from '@dhis2/d2-i18n';
+import { Button, IconSearch16, IconAdd16, spacers } from '@dhis2/ui';
+import { withStyles } from '@material-ui/core';
+import { NewTEIRelationshipStatuses } from '../WidgetTrackedEntityRelationship.const';
+import { RelationshipTypeSelector } from './RelationshipTypeSelector/RelationshipTypeSelector';
+
+const styles = {
```
simonadomnisoru, line 8: Hey, 
Can you have a look at the `WidgetTrackedEntityRelationship` width? On a laptop screen, you can see other widgets below. It looks fine on bigger screens. 

Screenshot 2022-02-07 at 15 45 18

Thanks!

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/WidgetTrackedEntityRelationship.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+import React, { useCallback, useState } from 'react';
+import { Button } from '@dhis2/ui';
+import i18n from '@dhis2/d2-i18n';
+import { NewTrackedEntityRelationship } from './NewTrackedEntityRelationship/NewTrackedEntityRelationship.container';
+import type { Props } from './WidgetTrackedEntityRelationship.types';
+
+export const WidgetTrackedEntityRelationship = ({ relationshipTypes, renderRef, trackedEntityType }: Props) => {
```
simonadomnisoru, line 8: Hey, 
Can you have a look at the `WidgetTrackedEntityRelationship` contents width? On a laptop screen, you can see other widgets below. It looks fine on bigger screens. 

Screenshot 2022-02-07 at 15 45 18

Thanks!

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/NewTrackedEntityRelationship.component.js
```diff
@@ -0,0 +1,51 @@
...
+        );
+    }
+
+    if (pageStatus === NewTEIRelationshipStatuses.MISSING_CREATION_MODE) {
+        return (
+            <div className={classes.container}>
+                <div className={classes.creationselector}>
+                    <Button className={classes.creationselector}>
+                        <IconSearch16 />
+                        <p>{i18n.t('Link to an existing person')}</p>
```
simonadomnisoru, line 37: Hey, 
Do you plan to handle the onClick callbacks in another ticket?
Thanks!

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/hooks.js
```diff
@@ -0,0 +1,66 @@
...
+} from '../../MetaDataStoreUtils/MetaDataStoreUtils';
+import { userStores } from '../../storageControllers/stores';
+import type { RelationshipType } from './WidgetTrackedEntityRelationship.types';
+
+export const useRelationshipTypes = () => {
+    const [cachedTypes, setCachedTypes] = useState();
+    const [loading, setLoading] = useState(true);
+    const error = useRef();
+    useMemo(() => {
+        getCachedResourceAsync(userStores.RELATIONSHIP_TYPES)
```
simonadomnisoru, line 14: Hey, 
Why do you use the data from the Indexed DB here? To make the widget self-contained you can directly call the API using a hook. Something like this: 
```
const { error, loading, data } = useDataQuery(
    useMemo(
        () => ({
            relationshipTypes: {
                resource: 'relationshipTypes',
                params: () => ({
                    fields: 'id,displayName,bidirectional,fromToName,toFromName,fromConstraint[*],toConstraint[*],access[*]',
                }),
            },
        }),
        [],
    ),
);

```
Thank you!

JoakimSM: I see that I used the term self-contained in the JIRA-issue; not really the term I was looking for, so I've updated the JIRA description. The idea is to pass in the relationship types to the Widget (See more info in this ticket: https://jira.dhis2.org/browse/DHIS2-12362 that @jasminenguyennn is working on). The relationship types passed in should also be used in the "add" part that @eirikhaugstulen is working on here. I.e, we shouldn't get the relationshipTypes in the Widget itself (So at some point we need to coordinate)

Concerning metadata retrieval in general: For the time being I think it makes sense to grab it from our own indexedDB cache, because:
1) We have an established pattern that a) ensures we have updated cache values b) that won't make unnecessary requests to the api.
2) Currently, we will often need to use refetch() from useDataQuery to get the data. refetch() will skip the app-runtime cache and redo the api request. In this particular example we won't have that problem, but often we will need to get some new data if one of the props changes (programId, teiId etc.).

When the app-runtime adds offline-caching we can experiment with it to see if it meets our requirements. If/when it meets our requirements when can lay out a plan to change our approach. I would really like not to be alpha testers for this since we don't need the extra work and I think it would be beneficial for us to see how this works in another app first (I have some concerns; especially how will they ensure the cache is up to date)

We should probably have a conversion about this, so it is clear to the whole team. I think it would be great if @Bekkalizer also joined. I will set something up when I get back.

JoakimSM (commented):
Looks like a good start @eirikhaugstulen 👍 I have some initial comments, will probably dive into more details later.

src/core_modules/capture-core/components/WidgetTrackedEntityRelationship/NewTrackedEntityRelationship/NewTrackedEntityRelationship.container.js
```diff
@@ -0,0 +1,103 @@
...
+    hideDialog,
+    relationshipTypes,
+    trackedEntityType,
+    classes,
+    ...passOnProps
+}: Props) => {
+    const [selectedRelationshipType, setSelectedRelationshipType] = useState();
+    const [creationMode, setCreationMode] = useState();
+    const { programId } = useLocationQuery();
+    const filteredRelationshipTypes = useFilteredRelationshipTypes(relationshipTypes, trackedEntityType, programId);
```
JoakimSM, line 51: Some structural suggestions that we can dive into more tomorrow/Monday:

- Move the filtering logic to the RelationshipTypeSelector component (will make the type selection live in it's own bucket instead of spreading it out)
- Build your own data structure in a hook (in the filtering hook or an additional hook) and use that data structure to display information in the selection component directly. We can reuse the display component when we are doing Event -> Tei relationships (and even for other types of relationships if we expand the model a bit). When selecting a type: save the constraint information that is needed for the link/register TEI part and final saving of the relationship (no need to re-find the relationship type here I feel)

But let's talk about this tomorrow/Monday since I didn't really elaborate here.

eirikhaugstulen:
Superseded by new PR.

PR #2498, opened 2022-02-08
fix: [DHIS2-12557] Unique TEAs

JoakimSM:
Fixes problem described in JIRA [DHIS2-12557](https://jira.dhis2.org/browse/DHIS2-12557) and also the app crashing in the registration page (when trying to display a unique tracked entity attribute error).


simonadomnisoru (approved):
Hey @JoakimSM, 

LGTM! The TEI is correctly linked now. 👏
I was curious if we can test this functionality somehow. In the database used by Cypress, I could not find any program that has assigned unique tracked entity attributes that are not "Automatically generated". Therefore this may be an edge case where end-to-end testing is not possible. 🤔

Thanks!

JoakimSM:
Hey @simonadomnisoru, absolutely makes sense to do some kind of testing here. Let me think about it 🙂

jasminenguyennn (approved):

JoakimSM:
Will merge this without the tests for now since we don't have the required data in DB.

PR #2499, opened 2022-02-08
fix: [DHIS2-12472] List all messages related to field - V36

eirikhaugstulen:


JoakimSM (commented):
I'm sorry for adding this comment after you've already merged this to other branches. But I think it would be good idea to always return arrays here because:
1) It would simplify a bit and no need to do typeof checks
2) We would earlier catch bugs that surfaces because the contents is an array. Something that comes to mind is the postProcessErrorMessage in the Formbuilder that expects a string (I have PR with some changes here so maybe merge that one first).

Maybe we should also rename some of the variables/properties to reflect that we can have multiple messages (errorMessage -> errorMessages)

eirikhaugstulen:
Superseded by https://github.com/dhis2/capture-app/pull/2513

PR #2508, opened 2022-02-11
feat: [DHIS2-9281] share view

simonadomnisoru:


src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/TemplateMaintenance/SharingDialog.component.js
```diff
@@ -32,7 +37,7 @@ const SharingDialogPlain = ({ onClose, open, templateId, classes }: Props) => {
             open={open}
             id={templateId}
             onRequestClose={handleClose}
-            type={'eventFilter'}
+            type={SHARING_DIALOG_TYPE[storeId] || ''}
```
jasminenguyennn, line 40: hey @simonadomnisoru  `storeId` is not optional so I think we don't need the fall back in this case?

JoakimSM (commented):
A couple of points I hope you can look into 🙂 

Currently the `WorkingListsBase` is very generic, it doesn't use Redux and doesn't care where it gets its data from. Therefore I would like to propose we do not send in storeId (implies Redux), but instead try to send it something more generic. For now, my suggestion would be to send in the sharing type directly, something like `TemplateSharingType`.

One bug I noticed that I didn't dig into (works for the event filters so you can compare): Change the sharing settings (add a user for example), close the dialog, make a change to the filters and then save the the updated view ("Update view" in the menu). If you open the sharing settings again you will see that the sharing settings are reset to the default. I would guess that the sharing settings are not posted as part of the update (if they are, we might have a backend bug on our hands). Great if you can investigate. Like i said, it's working for event filters, so that can be helpful when debugging.

JoakimSM (approved):
Good work! 👍

PR #2541, opened 2022-02-25
feat: [DHIS2-12141] TEI Re enrollment 

simonadomnisoru:


JoakimSM (approved):
Overall good work, Let's merge this 👍  Something we will need to improve up on later is how and when the form initialises. Made a ticket: https://jira.dhis2.org/browse/DHIS2-12733

PR #2544, opened 2022-02-25
feat: [DHIS2-12615] Link TEI-relationship with existing TEI

eirikhaugstulen:
Search-functionality for retrieving existing TEIs and linking these in a relationship.

Missing functionality:
- This search form does not treat a program constraint as a hard constraint, and is therefore changeable. This will need some work before merging in.
- Was not sure of what to add as commit & rollback when posting relationship to api - we should improve the error handling on this


eirikhaugstulen:
Superseded by new PR.

PR #2564, opened 2022-03-02
fix: [DHIS2-12119] fix cypress test

jasminenguyennn:
### Issue:

The behavior is expected so I only adapt the tests


simonadomnisoru (approved):
LGTM! Thanks @jasminenguyennn!

PR #2565, opened 2022-03-02
[DHIS2-12524] Allow no result to trigger fallback search

jasminenguyennn:
### Issue
[DHIS2-12524](): No option for "Search in all programs" if no results in current program

Allow no result to trigger fallback search, then display modal in case no result found on all search

Screenshot 2022-03-02 at 09 41 58

Screenshot 2022-03-02 at 09 42 10


simonadomnisoru (approved):
Looks good 🎉!

JoakimSM (approved):
LGTM

PR #2571, opened 2022-03-03
refactor: [TECH-920][TECH-660] use ScopeSelector in the New page

simonadomnisoru:


src/core_modules/capture-core/components/Pages/New/TopBar.js
```diff
@@ -0,0 +1,90 @@
...
+            customActionsOnProgramIdReset={[
+                cleanUpDataEntry(NEW_TEI_DATA_ENTRY_ID),
+                cleanUpDataEntry(NEW_SINGLE_EVENT_DATA_ENTRY_ID),
+                cleanUpDataEntry(NEW_RELATIONSHIP_EVENT_DATA_ENTRY_ID),
+            ]}
+            customActionsOnOrgUnitIdReset={[
+                cleanUpDataEntry(NEW_TEI_DATA_ENTRY_ID),
+                cleanUpDataEntry(NEW_SINGLE_EVENT_DATA_ENTRY_ID),
+                cleanUpDataEntry(NEW_RELATIONSHIP_EVENT_DATA_ENTRY_ID),
+            ]}
```
JoakimSM, line 78: We should strive to get rid of these properties and the functionality for them inside the scope selector (if we need some actions to be dispatched we can do it in `onResetOrgUnitId` or `onResetProgramId`). Been doing some testing today and It seems these clean ups are also being run from elsewhere (RegistrationDataEntry.container), so I don't think we need to dispatch these here at all. Am I mistaken?

JoakimSM (commented):
Looks like a pretty good first step to get out of the selector / syncing problem. I will push a PR that deals with a problem I found (deselecting org. unit also deselects category options) and does some minor clean up in an epic (an epic we should prioritise to do a proper refactor of soon, I will create a ticket). 

Great if you have a look at the other comment I made.

src/core_modules/capture-core/components/Pages/New/TopBar.js
simonadomnisoru: Hey, 
You were right those cleanups were already run from inside `RegistrationDataEntry.container`. I clean the code of them inside ScopeSelector in the latest commit. 
Thank you!

JoakimSM (approved):
🎉

PR #2590, opened 2022-03-07
[DHIS2-12783] hide refer tab

jasminenguyennn:


JoakimSM (approved):
LGTM

I changed the JIRA ticket to a bug, therefore we got a new ticket number. So if you could merge using DHIS2-12783 in the merge commit message it would be great.

simonadomnisoru (approved):

PR #2600, opened 2022-03-09
fix: [DHIS2-12778] tooltip display

jasminenguyennn:
### Issue 
[DHIS2-12778]: Fix tooltip in create new button

Screenshot 2022-03-09 at 07 56 31


simonadomnisoru (approved):
Looks good!

PR #2653, opened 2022-03-16
fix: [DHIS2-12869] publicAccess when creating a TEI filter

simonadomnisoru:


JoakimSM:
We will do this change on the backend instead. https://jira.dhis2.org/browse/DHIS2-12886

PR #2654, opened 2022-03-17
fix: [DHIS2-12870] use api fields name for TEIFilters

simonadomnisoru:
Remove the conversion and use API fields name 
`regDate` to `createdAt`
`regUnit` to `orgUnit`

### DIFF
--- a/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/getTeiListData.js
+++ b/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/getTeiListData.js
@@ -37,17 +37,10 @@ const getMainApiFilterQueryArgs = (filters?: RawFilterQueryArgs = {}, filtersOnl
             };
         }, {});
 
-const getApiOrderById = (sortById: string, columnsMetaForDataFetching: TeiColumnsMetaForDataFetching) => {
-    const { id } = columnsMetaForDataFetching.get(sortById) || {};
-    return id;
-};
-
 const getApiOrderByQueryArgument = (sortById: string, sortByDirection: string, columnsMetaForDataFetching: TeiColumnsMetaForDataFetching) => {
-    const apiId = getApiOrderById(sortById, columnsMetaForDataFetching);
-    return `${apiId}:${sortByDirection}`;
+    const { id } = columnsMetaForDataFetching.get(sortById) || {};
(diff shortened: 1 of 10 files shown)


src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/getTeiListData.js
```diff
@@ -38,8 +38,8 @@ const getMainApiFilterQueryArgs = (filters?: RawFilterQueryArgs = {}, filtersOnl
         }, {});
 
 const getApiOrderById = (sortById: string, columnsMetaForDataFetching: TeiColumnsMetaForDataFetching) => {
-    const { id, apiName } = columnsMetaForDataFetching.get(sortById) || {};
-    return apiName || id;
+    const { id } = columnsMetaForDataFetching.get(sortById) || {};
+    return id;
```
JoakimSM, line 42: I think we can simplify and remove this function altogether?

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/getTeiListData.js
```diff
@@ -37,17 +37,10 @@ const getMainApiFilterQueryArgs = (filters?: RawFilterQueryArgs = {}, filtersOnl
...
-const getApiOrderById = (sortById: string, columnsMetaForDataFetching: TeiColumnsMetaForDataFetching) => {
-    const { id, apiName } = columnsMetaForDataFetching.get(sortById) || {};
-    return apiName || id;
-};
-
 const getApiOrderByQueryArgument = (sortById: string, sortByDirection: string, columnsMetaForDataFetching: TeiColumnsMetaForDataFetching) => {
-    const apiId = getApiOrderById(sortById, columnsMetaForDataFetching);
-    return `${apiId}:${sortByDirection}`;
+    const { id } = columnsMetaForDataFetching.get(sortById) || {};
+    return id ? `${id}:${sortByDirection}` : 'createdAt:desc';
```
JoakimSM, line 42: Do you think it makes sense to change the logic so that the only time that we validate is when we load the view? I.e., we assume that sortById and sortByDirection is valid here, but when we actually load a view we check that the sort by column is actually one of the columns.

JoakimSM (approved):
Looks great.

PR #2656, opened 2022-03-18
fix: [DHIS2-12878] fix filter undefined value

jasminenguyennn:
### Issue 
[DHIS2-12878](https://jira.dhis2.org/browse/DHIS2-12878): Capture app adds a wrong filter when TEI attribute filter contains unsupported operand

### DIFF
--- a/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/helpers/TEIFilters/apiTEIFilterToClientConfigConverter/convertToClientFilters.js
+++ b/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/helpers/TEIFilters/apiTEIFilterToClientConfigConverter/convertToClientFilters.js
@@ -22,7 +22,7 @@ import type {
 } from '../../../types';
 import { DATE_TYPES, ASSIGNEE_MODES } from '../../../constants';
 
-const getTextFilter = (filter: ApiDataFilterText): TextFilterData => {
+const getTextFilter = (filter: ApiDataFilterText): ?TextFilterData => {
     const value = filter.like;
     return value ? { value } : undefined;
 };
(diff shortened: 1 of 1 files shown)


simonadomnisoru (approved):
Looks good!

JoakimSM (commented):
For text filters this looks good. can we also the same for the other supported value types? You can skip optionSets for now since there is some complexity there.

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/helpers/TEIFilters/apiTEIFilterToClientConfigConverter/convertToClientFilters.js
```diff
@@ -22,15 +22,20 @@ import type {
...
-    return { value };
+    return value ? { value } : undefined;
 };
 
-const getNumericFilter = (filter: ApiDataFilterNumeric): NumericFilterData => ({
-    ge: filter.ge ? Number(filter.ge) : undefined,
-    le: filter.le ? Number(filter.le) : undefined,
-});
+const getNumericFilter = (filter: ApiDataFilterNumeric): ?NumericFilterData => {
+    if (filter.ge && filter.le) {
```
JoakimSM, line 31: Either ge or le should be enough to activate the filter.

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/helpers/TEIFilters/apiTEIFilterToClientConfigConverter/convertToClientFilters.js
```diff
@@ -22,15 +22,20 @@ import type {
...
+        return {
+            ge: Number(filter.ge),
+            le: Number(filter.le),
+        };
+    }
+    return undefined;
+};
 
 const getBooleanFilter = (filter: ApiDataFilterBoolean): BooleanFilterData => ({
     values: filter.in.map(value => value === 'true'),
```
JoakimSM, line 41: We should probably do something with the Boolean filter too

PR #2661, opened 2022-03-21
fix: [DHIS2-12880] refetch tei after updating

jasminenguyennn:
### Issue 
[DHIS2-12880](https://jira.dhis2.org/browse/DHIS2-12880): refetch TEI after updating widget enrollment


src/core_modules/capture-core/components/WidgetEnrollment/Actions/actions.types.js
```diff
@@ -4,6 +4,7 @@ import type { QueryRefetchFunction } from '@dhis2/app-runtime';
 export type Props = {|
     enrollment: Object,
     refetch: QueryRefetchFunction,
```
simonadomnisoru, line 6: NIT: rename `refetch` to `referchEnrollment`

simonadomnisoru (approved):

PR #2665, opened 2022-03-22
fix: [DHIS2-12930] add correct fields-filtering on TEI-search

eirikhaugstulen:
Fix the errors that occurred after implementing fields-filtering in BE.


i18n/en.pot
```diff
@@ -1035,6 +1035,9 @@ msgstr "Write a comment about this enrollment"
 msgid "This enrollment doesn't have any comments"
 msgstr "This enrollment doesn't have any comments"
 
+msgid "organisation unit could not be retrieved. Please try again later."
```
simonadomnisoru, line 1038: Hey, 
It looks like this `msgid` was deleted on master yesterday by this commit 81d2e4c456e1eae0fee12406f515173a797f7e3b.  Can you pull the lastest version? 
Thank you!

eirikhaugstulen: Hey,

Tried to update my local branch from master, and I also compared it to remote/master, but this is still present. (Even though the comparison says that the files are identical)
Should I manually delete this msgid?

Thanks!

simonadomnisoru: Hey,
Thank you for looking into it. I dig more into the commit I mentioned 81d2e4c456e1eae0fee12406f515173a797f7e3b. This `msgid` exists in the codebase, the previous commit shouldn't have deleted it and your PR is correctly adding it back.  
Thank you!

simonadomnisoru (approved):

JoakimSM (approved):

PR #2672, opened 2022-03-23
fix: [DHIS2-12929] use interface language for localeCompare

JoakimSM:
By default localeCompare uses the locale from your browser and this can lead to some unexpected behaviour. Changed localeCompare to use the interface language from DHIS.

### DIFF
--- a/src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/sortIndicatorsFn.js
+++ b/src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/sortIndicatorsFn.js
@@ -1,13 +1,18 @@
-// eslint-disable-next-line complexity
+/* eslint-disable complexity */
+import { localeCompareStrings } from '../../../utils/localeCompareStrings';
+
 export const sortIndicatorsFn = (a, b) => {
     if (typeof b === 'string') {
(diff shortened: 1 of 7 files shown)


src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/sortIndicatorsFn.js
```diff
@@ -1,13 +1,18 @@
...
+            (a.key && localeCompareStrings(a.key, b.message)) ||
+            (a.message && localeCompareStrings(a.message, b.message))
+        );
     }
     if (b.key) {
-        return (((a.key && a.key.localeCompare(b.key)) || (a.message && a.message.localeCompare(b.key))));
+        return ((a.key && localeCompareStrings(a.key, b.key)) || (a.message && localeCompareStrings(a.message, b.key)));
     }
     return 1;
 };
```
JoakimSM, line 18: @eirikhaugstulen: Can you have a second look at this function when you find the time (in another PR). It seems like we are missing some logic for typeof a === 'string' and the result could depend on whether an item is on the a or b side.

simonadomnisoru (approved):

jasminenguyennn (approved):

PR #2677, opened 2022-03-24
fix: [DHIS2-12928] handle missing program rules outputConverters

simonadomnisoru:


src/core_modules/capture-core-utils/rulesEngine/constants/typeToInterfaceFnName.const.js
```diff
@@ -20,4 +20,9 @@ export const mapTypeToInterfaceFnName = {
     [typeKeys.PERCENTAGE]: 'convertPercentage',
     [typeKeys.URL]: 'convertUrl',
     [typeKeys.AGE]: 'convertAge',
+    [typeKeys.FILE_RESOURCE]: 'convertFile',
+    [typeKeys.ORGANISATION_UNIT]: 'convertOrganisationUnit',
+    [typeKeys.IMAGE]: 'convertImage',
+    [typeKeys.USERNAME]: 'convertUserName',
+    [typeKeys.COORDINATE]: 'convertCoordinate',
```
JoakimSM, line 27: 👍

src/core_modules/capture-core-utils/rulesEngine/constants/typeKeys.const.js
```diff
@@ -23,6 +23,6 @@ export const typeKeys = {
     ORGANISATION_UNIT: 'ORGANISATION_UNIT',
     AGE: 'AGE',
     URL: 'URL',
-    FILE: 'FILE',
+    FILE_RESOURCE: 'FILE_RESOURCE',
```
JoakimSM, line 26: 👍

src/core_modules/capture-core-utils/rulesEngine/processors/rulesEffectsProcessor/rulesEffectsProcessor.js
```diff
@@ -97,8 +99,17 @@ export function getRulesEffectsProcessor(
...
+                return valueType;
+            }
             // $FlowExpectedError
-            outputValue = outputConverters[converterName](normalizedValue);
+            const outputConverter = outputConverters[converterName];
+            if (!outputConverter) {
+                log.warn(errorCreator('outputConverter for valueType is missing')({ valueType }));
+                return valueType;
+            }
+            outputValue = outputConverter(normalizedValue);
```
JoakimSM, line 112: I think it makes sense log something here if we don't find the converterName or outputConverter, so that's good. But I think returning `''` would be preferable (instead of valueType). Also, maybe we can combine the logging?

Anyway, from what I could gather the main problem here is that we're missing valueType for the dataElements that are passed into the rules engine (i.e, valueType here will always be undefined for all dataElements). Great if you can look into this.

JoakimSM (approved):

PR #2679, opened 2022-03-24
fix: [DHIS2-12950] Use old Tracker Capture dashboard

simonadomnisoru:
- use the old Tracker Capture dashboard. 
- it should NOT be possible to switch to the new dashboard using the datastore config.


jasminenguyennn (approved):

JoakimSM (approved):
LGTM

PR #2681, opened 2022-03-25
fix: [DHIS2-12951] default old Tracker Capture dashboard

simonadomnisoru:
- should use the old Tracker Capture dashboard as the default, 
- allow opt-in to the new dashboard by setting the "useNewDashboard" dataStore config

The user can reach the new enrollment Dashboard if they set the `programId` true in the useNewDashboard. Steps to enable the new enrollment dashboard navigation:

1. Make a POST-request to` /api/dataStore/capture/useNewDashboard`
2. Make API-body JSON with `programId: true/false`
3. The capture app should then navigate the user to the old tracker if the program is set to true.


JoakimSM (approved):
LGTM 👍

PR #2712, opened 2022-04-05
chore: [DHIS2-13029] attempt to fix cypress test

jasminenguyennn:
### Issue 
Test scenario: New event page > Url with invalid program id randomly failed on PRs

I look into the capture of the run, the page url has not changed when `cy.visit`, this PR tried to fix that by reassign the url, make the page reload after that.


JoakimSM (approved):
As a temporary workaround, I'm fine with this. When we have replaced the LockedSelector everywhere we can revisit.

PR #2751, opened 2022-04-26
feat: [DHIS2-12140] add the search components into the Main page

simonadomnisoru:


src/core_modules/capture-core/components/TemplateSelector/TemplateSelector.container.js
```diff
@@ -0,0 +1,25 @@
...
+import { buildUrlQueryString, useLocationQuery } from '../../utils/routing';
+import { useTEIFilters } from './useTemplates';
+
+export const TemplateSelector = () => {
+    const history = useHistory();
+    const { programId, orgUnitId } = useLocationQuery();
+    const { TEIFilters } = useTEIFilters(programId);
+
+    const onSelectTemplate = template =>
+        history.push(`/?${buildUrlQueryString({ orgUnitId, programId, selectedTemplateId: template.id })}`);
```
simonadomnisoru, line 14: Hey, 
I couldn't find a way to replace `history.push` for the functional approach. The `history` object is mutable and the react-router-dom library gives only this example for the `useHistory()` hook. Let me know if you have any ideas on how to make this functional.   
Thank you!

PR #2757, opened 2022-04-30
feat: [DHIS2-12678] Stay in the same page when clicking search fallback

jasminenguyennn:
### Feature

[DHIS2-12678](https://jira.dhis2.org/browse/DHIS2-12678): Stay in the same page when clicking search fallback

When you execute the manual fallback, you should stay on the same page and the "search in all programs" results should just be appended below the search in current program results. 
_Note_: deduplication is not in the scope of this ticket and will need to be discussed with the BE.

Screenshot 2022-04-30 at 08 12 32

Documentation PR: https://github.com/dhis2/capture-app/pull/2784


simonadomnisoru:
Hey @jasminenguyennn, 

Overall, it looks good! 👏 I found a couple of small bugs, maybe you can have a look at them. 

1. The "Results found in all programs" components remain visible when changing the program or when clicking again on "Search by attributes" button. The steps to reproduce it are in the video below. 

https://user-images.githubusercontent.com/6817073/166879034-9bab690a-ea36-47a3-ad20-9a074ca1bd6a.mov


2. When the user clicks the "View dashboard" button the program context is changed and the enrollment is selected. According to Jira's description, the option "View dashboard" should take the user to the enrollment dashboard, with no enrollment selected, but keeping the program context. The steps to reproduce it are in the video below.

https://user-images.githubusercontent.com/6817073/166879079-5b4fb8d6-3a98-4464-b53d-4b80602f70b4.mov


3. Lastly the Jira ticket mentions that the two result modules should be collapsible to save space. Is this behavior possible to implement?  


Let me know if these suggestions make sense and if you have any questions. 
Thank you!

jasminenguyennn:
thanks @simonadomnisoru. I updated the code accordingly for point 1 and 3. 
Regarding point 2, I think the behaviour to select first active enrollment was done across the app and not sure if we want to change that.

simonadomnisoru (approved):
Thanks for looking into my suggestions! 😄

geethaalwan (commented):
Tested successfully on 2.39 here https://deploy-preview-2757--dhis2-capture.netlify.app/

geethaalwan (approved):

JoakimSM (commented):
Some initial thoughts (that we hopefully can improve): 

1. We should disable fallback search if one of the conditions defined in the ticket are met. I.e, if no searchable attributes for the TET are filled in or if there aren't enough TET searchable attributes filled in (based on on the minimum number of attributes required to search) we should show a modal instead of executing the search.
2. We should keep all search criteria the user added in the form (even if it's not part of the TET searchable attributes). For example, in the "Child programme", try to set "Gender" to "Male" and search in all programs. You will see that "Male" is wiped from the form.
3. Can we skip refresh of the rest of the page when searching in all programs? 
4. This one is up for debate and not specified in the ticket: I would prefer if we collapsed the first search section when we search in all programs. I found it quite confusing to have to scroll through the first list before finding the new list (when there are more results then what fits on the screen). @Bekkalizer and @karolinelien opinions welcome.

src/core_modules/capture-core/components/Pages/Search/SearchResults/SearchResults.component.js
```diff
@@ -16,6 +17,9 @@ import { useScopeInfo } from '../../../../hooks/useScopeInfo';
 import {
     navigateToEnrollmentOverview,
 } from '../../../../actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.actions';
+import { buildUrlQueryString } from '../../../../utils/routing';
+import { getUrlQueries } from '../../../../utils/url';
```
JoakimSM, line 21:
```suggestion
import { useLocationQuery } from '../../../../utils/routing';
```

src/core_modules/capture-core/components/Pages/Search/SearchResults/SearchResults.component.js
```diff
@@ -171,40 +189,95 @@ export const SearchResultsIndex = ({
...
+        >
+            <CardList
+                noItemsText={i18n.t('No results found')}
+                currentSearchScopeName={currentSearchScopeName}
+                items={otherResults}
+                dataElements={dataElements}
+                renderCustomCardActions={({ item, enrollmentType, programName }) => (<CardListButtons
                     programName={programName}
-                    currentSearchScopeId={currentSearchScopeId}
+                    currentSearchScopeId={undefined}
```
JoakimSM, line 253: can we remove the prop if undefined is the correct value?

JoakimSM (approved):

PR #2762, opened 2022-05-02
fix: [DHIS2-10332] confirm metadata loading for program at the end of the sequence

JoakimSM:
Earlier, if you aborted the loading sequence at certain points, you could experience missing metadata on subsequent app loads. This fix ensures that the loading of programs and related metadata is complete before confirming the load sequence.


jasminenguyennn (approved):

src/core_modules/capture-core/metaDataStoreLoaders/baseLoader/confirmLoadingSequence.js
```diff
@@ -0,0 +1,17 @@
...
+export const confirmLoadingSequence = async (stalePrograms: StalePrograms) => {
+    const { storageController, storeNames } = getContext();
+    await Promise.all(
+        stalePrograms.map(async ({ id, version }) => {
+            const program = await storageController.get(storeNames.PROGRAMS, id);
+            const updatedProgram = {
+                ...program,
+                version,
+            };
+            await storageController.set(storeNames.PROGRAMS, updatedProgram);
```
superskip, line 14: I have one question: do you know at which point in the process of initialising the cache the problem occurred? My feeling after reading this PR is that it happens during the transaction of javascript objects to entries in indexed DB. This line initiates such a transaction, and the shape of the statement suggests there's a chance the whole entry might be reinitialised.

simonadomnisoru (approved):

src/core_modules/capture-core/metaDataStoreLoaders/baseLoader/confirmLoadingSequence.js
JoakimSM: Basically there are multiple transactions taking place if the program version has changed and this fix ensures that the confirmation happens after all the transactions instead of confirming in the first transaction (by updating the version in the cache). I think what you describe here should be handled by the IndexedDB. We can have a chat if it is unclear.

superskip (approved):

geethaalwan (approved):
Tested successfully on 2.39 and 2.38.1  here https://deploy-preview-2762--dhis2-capture.netlify.app/

PR #2764, opened 2022-05-04
refactor: [DHIS2-13049] use scopeselector in main page

simonadomnisoru:


src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
```diff
@@ -45,7 +41,10 @@ const MainPageContainer = () => {
         .push(`/?${buildUrlQueryString({ programId })}&all`);
 
     const MainPageStatus = useMemo(() => {
-        const selectedProgram = programId && programCollection.get(programId);
+        if (!programId) return MainPageStatuses.DEFAULT;
+        const withoutOrgUnit = () => !orgUnitId && !showAllAccessible;
```
jasminenguyennn, line 45: I wonder what is the benefit of declaring a function here when it can be just a variable?

simonadomnisoru: Hey, 
The benefit, in this case, is less complexity in the parent function. By moving some of the logic into other functions, the complexity was reduced from 12 to 8 and `eslint` no longer complains and highlights the code.
Thanks!

jasminenguyennn: but it can just be `const withoutOrgUnit = !orgUnitId && !showAllAccessible` with a useMemo?

simonadomnisoru: I updated the code with your suggestion. Thanks!

jasminenguyennn (approved):

geethaalwan (approved):
Tested successfully on 2.39 version here https://deploy-preview-2764--dhis2-capture.netlify.app

PR #2769, opened 2022-05-06
fix: [DHIS2-12188] do not download program indicators unless displayInForm=true

superskip:
The app does not use program indicators with displayInForm=false, so we ask the backend to filter out these to cut down on the loading time.

### DIFF
--- a/src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storeProgramIndicators.js
+++ b/src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storeProgramIndicators.js
@@ -21,7 +21,7 @@ export const storeProgramIndicators = async (programIds: Array<string>) => {
         resource: 'programIndicators',
         params: {
             fields: fieldsParam,
-            filter: [`displayInForm:eq:true`, `program.id:in:[${programIds.join(',')}]`],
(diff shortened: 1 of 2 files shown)


simonadomnisoru (approved):

src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storeProgramIndicators.js
```diff
@@ -13,15 +13,15 @@ const convert = (response) => {
...
+const fieldsParam = 'id,displayName,code,shortName,style,expression,' +
 'displayDescription,description,filter,program[id]';
 
 export const storeProgramIndicators = async (programIds: Array<string>) => {
     const query = {
         resource: 'programIndicators',
         params: {
             fields: fieldsParam,
-            filter: `program.id:in:[${programIds.join(',')}]`,
+            filter: `displayInForm:eq:true&filter=program.id:in:[${programIds.join(',')}]`,
```
Bekkalizer, line 24: To make two separate filter parameters, we need to split the filter into an array with two items, like this:
filter: [ `programid filter`, `display in form filter`],

superskip: Yes, that worked! :tada:

Bekkalizer (approved):

PR #2772, opened 2022-05-06
fix: [DHIS2-10332] confirm metadata loading for program at the end of the sequence

JoakimSM:


superskip (commented):
Shouldn't there be a backport for v38 rather than v35?

jasminenguyennn (approved):

JoakimSM:
> Shouldn't there be a backport for v38 rather than v35?

@superskip We actually don't need to backport to v38 anymore since the next 2.38 release will include the app from the master branch (the v38 branch will be deleted shortly). I opted to do an backport for v35 because this was one the versions the reporter had problems with and it was an easy backport. But you are right that we usually only back port to the 3 pervious versions.

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.35.14 here https://deploy-preview-2772--dhis2-capture.netlify.app/

PR #2775, opened 2022-05-08
fix: [DHIS2-13241] URL out of sync after adding event

eirikhaugstulen:
After adding an event, the user is redirected to the main-page with the programId and orgUnitId is stored in the URL. 
The URL is missing a "?", and the Ids are therefore interpreted as a part of the pathname, and not as a query parameter.
A refresh would cause the app to lose the program- and orgUnit-context.


jasminenguyennn (approved):

geethaalwan (approved):
Tested successfully on 2.39 and 2.38.1 here https://deploy-preview-2775--dhis2-capture.netlify.app/

PR #2778, opened 2022-05-10
feat: [DHIS2-12140] working list based on displayfrontpagelist

simonadomnisoru:


JoakimSM (commented):
Thanks @simonadomnisoru,

Can we always use the URL as the template/view/working list source? And update the template id in the Redux store based on what we have in the URL? I'm available for a chat if you want.

JoakimSM (commented):
I dived a bit more into this and it seems to work well 👍 I have some structural suggestions though. The working lists (event and tracker) are a good candidates for independent Widgets in the future. The `WorkingListsBase` "module" is "ready" be used in such a Widget and it would be great to not add code here that would break this (and require a refactor). The `TeiWorkingLists` would definitely require some refactoring, but I think it will be worth the effort if we hand off the responsibility for updating the URL to the outer components by exposing an event: "onChangeTemplate" (or something similar). When we want to update the URL from within the `TeiWorkingLists` we would call `onChangeTemplate` and the callback handler would do the URL-updating logic.

Let me know if you want a chat about this 🙂

jasminenguyennn (approved):

JoakimSM (approved):
LGTM 👍

geethaalwan (approved):

PR #2781, opened 2022-05-11
fix: [DHIS2-12120] conversion of program indicators to program rules

superskip:
- Renamed `eventDate` -> `occurredAt` to fix behaviour of source type `DATAELEMENT_PREVIOUS_EVENT`.
- Changed parsing of variables appearing as `programStageId.dataElementId` in program indicator expressions.

### DIFF
--- a/src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/WidgetIndicatorContent.js
+++ b/src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/WidgetIndicatorContent.js
@@ -86,7 +86,7 @@ const WidgetIndicatorContentComponent = ({ widgetData, emptyText, classes }: Con
             className={cx(classes.indicatorRow, { isLastItem })}
         >
             <div>{indicator.key}</div>
-            {(String(indicator.value) || indicator.color) && ( // convert to string to make sure 0 gets displayed
+            {(indicator.value || indicator.color) && (
                 <div className={classes.indicatorValue}>
                     {indicator.color ? renderLegend(indicator.color) : null}
                     {indicator.value}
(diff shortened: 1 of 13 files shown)


jasminenguyennn (approved):

simonadomnisoru:
Hey @superskip, 

I am not able to see the Indicator widget in capture-app. The indicator appears in the old tracker. You can see the steps on how I configured the program rules and the rule variable in the video below. 

https://user-images.githubusercontent.com/6817073/168031135-4576ec03-5dbd-4fe4-b0f9-350efab354bc.mov

Thank you!

src/core_modules/capture-core/components/WidgetIndicator/WidgetIndicatorContent/WidgetIndicatorContent.js
```diff
@@ -86,7 +86,7 @@ const WidgetIndicatorContentComponent = ({ widgetData, emptyText, classes }: Con
             className={cx(classes.indicatorRow, { isLastItem })}
         >
             <div>{indicator.key}</div>
-            {(indicator.value || indicator.color) && (
+            {(String(indicator.value) || indicator.color) && ( // convert to string to make sure 0 gets displayed
```
JoakimSM, line 89: The flow type definition says that the indicator value should be a string when passed on to the Widget. I feel that makes sense and think it would be better to do this conversion in the rules engine itself. The resulting value from the following program rule actions should be converted to string in the rules engine: Display key/value pair, display text, error on complete, show error, show warning, warning on complete.

@Bekkalizer: Concerns?

Bekkalizer: It seems sensible that items to display should always be converted to text in the engine.

JoakimSM (approved):

vilkg (approved):

PR #2782, opened 2022-05-11
fix: [DHIS2-8613] disable future dates in the "occurredAt" field

superskip:


src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js
```diff
@@ -119,6 +119,7 @@ const buildReportDateSettingsFn = () => {
             calendarWidth: 350,
             label: props.formFoundation.getLabel('occurredAt'),
             required: true,
+            latestSelectable: new Date(),
```
simonadomnisoru, line 122: Hey @superskip, 

I think that from the `EditEventDataEntry.component.js` file you only restrict future dates for the case when the user is editing an event. Does it make sense to add `latestSelectable: new Date()` inside [SingleEventRegistrationEntry](https://github.com/dhis2/capture-app/blob/812d16523c552adb0aa4038754ccfdfbfacef7d5/src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/DataEntry.component.js#L159) so that future dates are restricted when the user adds a new event also?

Thank you!

superskip: That is very sensible! :+1: 
Restricting future dates on edit event, like I have done here - not so much. It actually makes more sense to have no date restriction here, and only have it on the registration page, I think..

simonadomnisoru (approved):

superskip:
I observed that this feature has already been implemented for enrollment date. I made another PR using the same solution as enrollment dates. For the record, this PR does not in its present state restrict the report date for events in tracker programs.

PR #2783, opened 2022-05-12
fix: [DHIS2-12138] remove duplicate count call

simonadomnisoru:


JoakimSM (commented):
Looks good @simonadomnisoru 👍 The only potential problem scenario I can see is: If you exit the page while the api request for the duplicate check is running, the duplicate check will overwrite the reset. Implementing a `takeUntil` in the `loadSearchGroupDuplicatesForReviewEpic` should fix this.

vilkg (approved):
lgtm :)

PR #2793, opened 2022-05-16
fix: [DHIS2-13289] disable non-admin user cypress scenario

simonadomnisoru:
[DHIS2-13289](https://jira.dhis2.org/browse/DHIS2-13289)


superskip (approved):
Ok, it should be fine to delete that file. We can easily recover it by reverting this commit.

jasminenguyennn (approved):

PR #2795, opened 2022-05-17
feat: [DHIS2-13279] navigate to register new user

jasminenguyennn:
### Feature

[DHIS2-12372](https://jira.dhis2.org/browse/DHIS2-12372)

In case the user searches for an entity and the system returns zero results in all programs then the user sees a modal. In the modal they can either
a) choose to go back and edit their search terms or 
b) choose to create a new user.

In case of b) the user will be prompted to the registration page. The registration form will be prefiled with the search terms that the user has filled before. 

 
This ticket is to implement the Register an user button as b option.


simonadomnisoru:
Hey @jasminenguyennn, 

I noticed that the registration form is not prefilled with the search terms that the user has filled in the search page. You can see the steps I followed in the video below.

https://user-images.githubusercontent.com/6817073/169808072-90832f64-0d3c-47f9-ae0d-bd6d45c093fe.mov

 Can you have a look at it? Thank you!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.38.2 and 2.39 versions

PR #2798, opened 2022-05-18
feat: [DHIS2-13285] Preserve view settings when changing org unit

JoakimSM:
[DHIS2-13285](https://jira.dhis2.org/browse/DHIS2-13285)

This PR ensure that changing the org. unit doesn't reset the view settings. Data will be fetched based on the previous view settings (but reset to the first page).

### DIFF
--- a/src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
+++ b/src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
@@ -62,6 +62,9 @@ const getStyles = (theme: Theme) => ({
     loadingCell: {
         textAlign: 'center',
     },
+    loader: {
+        display: 'inline-block',
+    },
 });
 
 export type Column = {
@@ -96,6 +99,7 @@ type Props = {
         loadingRow: string,
         row: string,
         dataRow: string,
+        loader: string,
     }
 }
 
@@ -216,7 +220,7 @@ class Index extends React.Component<Props> {
                         colSpan={columnsCount}
                         className={classNames(classes.cell, classes.bodyCell, classes.loadingCell)}
                     >
-                        <CircularLoader />
+                        <CircularLoader className={classes.loader} />
                     </Cell>
                 </Row>
             ) : this.renderRows(visibleColumns, columnsCount);
--- a/src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/ListViewLoader/ListViewLoader.component.js
+++ b/src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/ListViewLoader/ListViewLoader.component.js
@@ -7,14 +7,13 @@ import { useIsContextInSync } from './useIsContextInSync';
 import type { Props } from './listViewLoader.types';
 
 const EventListUpdaterWithLoadingIndicator = withErrorMessageHandler()(
-    withLoadingIndicator(() => ({ margin: 10 }))(ListViewUpdater));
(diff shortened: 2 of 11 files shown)


src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.js
```diff
@@ -62,6 +62,9 @@ const getStyles = (theme: Theme) => ({
     loadingCell: {
         textAlign: 'center',
     },
+    loader: {
+        display: 'inline-block',
```
JoakimSM, line 66: A bit out of scope, but this is just ensuring that the spinner is aligned in the center (horizontally)

src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/ListViewLoader/ListViewLoader.component.js
```diff
@@ -61,28 +59,26 @@ const useLoadView = ({
...
         prevTemplateRef.current = currentTemplate;
         firstRunRef.current = false;
-
         if (triggerLoad) {
             onLoadView(currentTemplate,
                 { programId, programStageId, orgUnitId, categories },
             );
         }
         return () => cancelLoadViewIfApplicable();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
simonadomnisoru, line 75: Hey, 
What is the scenario where the hook behaves inconsistently due to `orgUnitId` being part of the dependency array? 
From the cases I tested, it looked like `eslint-disable-next-line` can be avoided if you check that `orgUnitId` has a value before calling the `onLoadView` function. 
Thanks!

simonadomnisoru:
Hey,  
I found a bug when the user wants to see all the events accessible in a program. The working list doesn’t load and it just shows the spinner. Is it related to your changes? 

all org unit

Thanks!

src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/ListViewLoader/ListViewLoader.component.js
JoakimSM: I think there is some advanced scenarios where we can get in trouble that might/might not be possible in the UI right now (for example: still in loading, change orgUnit comes to mind). This could probably have been refactored, but there is some complexity here that I don't want to dive into right now. I think it's fine to disable this check at times to keep the complexity down and make it easier to read.

JoakimSM:
> Hey, I found a bug when the user wants to see all the events accessible in a program. The working list doesn’t load and it just shows the spinner. Is it related to your changes?
> 
> all org unit
> 
> Thanks!

Likely, but I'm not able to reproduce it 🤔 Can you give me a step by step guide?

simonadomnisoru:
> @simonadomnisoru Likely, but I'm not able to reproduce it 🤔 Can you give me a step by step guide?

Hey @JoakimSM,
I am not able to reproduce it anymore today. Probably there was something with the cache. These are the steps I did a few days ago when I could reproduce it: 
1. Go to the main page
2. Select the "Child Programme" program
3. Click the "Or see all records accessible to you in Child Programme " button
4. The working list doesn’t load and it just shows the spinner.

Thank you!

simonadomnisoru (approved):

JoakimSM:
Ok @simonadomnisoru, I'll move this on to testing since we can't reproduce.

geethaalwan (approved):

PR #2799, opened 2022-05-18
feat: [DHIS2-12611] open enrollment event

simonadomnisoru:


src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storePrograms.js
```diff
@@ -6,9 +6,9 @@ import type { apiProgramsResponse } from './types';
 const convert = (() => {
     const sort = (arr: Array<any>, sortBy: string = 'sortOrder') => {
         arr.sort((a, b) => {
-            if (a[sortBy] == null) {
+            if (a[sortBy] === null) {
```
JoakimSM, line 9: I believe we would get in trouble here if we have programs where some of the stages have sort order while others don't (and I think we have to treat undefined the same way as null). Suggestion for sort order implementation: 
```
if (a[sortBy] == null) {
   return b[sortBy] == null ? 0 : 1;
}

if ((b[sortBy] == null)) {
   return -1;
}

return a[sortBy] - b[sortBy];
```

jasminenguyennn (approved):

JoakimSM (approved):

geethaalwan (approved):

PR #2811, opened 2022-05-23
fix: [DHIS2-13062] Fix profile tet name

jasminenguyennn:


src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -36,6 +35,7 @@ const WidgetProfilePlain = ({ teiId, programId, showEdit = false, orgUnitId = ''
         loading: trackedEntityInstancesLoading,
         error: trackedEntityInstancesError,
         trackedEntityInstanceAttributes,
+        trackedEntityTypeName,
```
simonadomnisoru, line 38: Hey @jasminenguyennn , 

I believe that all the information related to the `trackedEntityType` is already inside the `program` variable. 
It can be accesed like this: `const trackedEntityTypeName = program?.trackedEntityType?.displayName;`
Does it make sense to get from the `program` and avoid doing another call to the API? 

Thank you!

simonadomnisoru: Hey, 

The `WidgetProfile` doesn't use any cached data from IndexDB. Each time the component is mounted there is a call to the API to get the whole `program` metadata based on the `programId`. In my opinion this logic ensures that all the `program` information (attributes, programRules, form elements, sections, TET etc.) fetched on [line 30](https://github.com/dhis2/capture-app/blob/e037f24f9f5ee8bac7fb754919fde8f3cf0f91b9/src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js#L30 ) will reflect the latest changes from Maintenance every time the compoent is mounted. 

Does it make sense or we can have a call about it to discuss it further? Thanks!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.39 and 2.38.1 versions

PR #2842, opened 2022-05-31
fix: [DHIS2-13183] navigation to register new single event page

superskip:


src/core_modules/capture-core/components/Pages/New/newPage.epics.js
```diff
@@ -6,11 +6,13 @@ import { topBarActionsActionTypes } from '../../TopBarActions/TopBarActions.acti
 import { buildUrlQueryString, deriveURLParamsFromLocation } from '../../../utils/routing';
 import { resetLocationChange } from '../../LockedSelector/QuickSelector/actions/QuickSelector.actions';
 
-export const openNewRegistrationPageFromLockedSelectorEpic = (action$: InputObservable, _: ReduxStore, { history }: ApiUtils) =>
+export const openNewRegistrationPageFromLockedSelectorEpic = (action$: InputObservable, store: ReduxStore, { history }: ApiUtils) =>
     action$.pipe(
         ofType(lockedSelectorActionTypes.NEW_REGISTRATION_PAGE_OPEN, topBarActionsActionTypes.NEW_REGISTRATION_PAGE_OPEN),
         switchMap(() => {
-            const { orgUnitId, programId } = deriveURLParamsFromLocation();
+            const params = deriveURLParamsFromLocation();
```
simonadomnisoru, line 13: Hey @superskip,

Does it make sense to send `orgUnitId` and `programId` when the action `openNewRegistrationPageFromScopeSelector` is [dispatched](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/TopBarActions/TopBarActions.container.js#L49) and then use it in the epic like this?
`const { orgUnitId = action.payload.orgUnitId, programId = action.payload.programId } =  deriveURLParamsFromLocation(); `

In the near future, we want to remove the `currentSelections` logic from Redux and that's why I'm suggesting this alternative approach.

Thank you!

simonadomnisoru, line 13: Hey @superskip,

Does it make sense to send `orgUnitId` and `programId` when the action `openNewRegistrationPageFromScopeSelector` is [dispatched](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/TopBarActions/TopBarActions.container.js#L49) and then use it in the epic like this?
`const { orgUnitId = action.payload.orgUnitId, programId = action.payload.programId } =  deriveURLParamsFromLocation(); `

In the near future, we want to remove the `currentSelections` logic from Redux and that's why I'm suggesting this alternative approach.

Thank you!

JoakimSM: I agree with the approach in general @simonadomnisoru. Looking into the code I realised that this PR is using the old LockedSelector. Surely we can merge the new selector in, but the entire workflow for opening the new page is troublesome and we should prioritise a refactor. I've made a new task for this here: https://dhis2.atlassian.net/browse/TECH-1202. I would say we keep the changes done here by @superskip and do some refactoring in this new task.

JoakimSM: I agree with the approach in general @simonadomnisoru. Looking into the code I realised that this PR is using the old LockedSelector. Surely we can merge the new selector in, but the entire workflow for opening the new page is troublesome and we should prioritise a refactor. I've made a new task for this here: https://dhis2.atlassian.net/browse/TECH-1202. I would say we keep the changes done here by @superskip and do some refactoring in this new task.

simonadomnisoru (approved):

vilkg (approved):

PR #2864, opened 2022-06-14
docs: [DHIS2-12611] open enrollment event

simonadomnisoru:


docs/user/using-the-capture-app.md
```diff
@@ -200,14 +200,13 @@ To configure the auto generation of an event you need to take the following step
...
 ![](resources/images/auto-generated-05.png)
 
-Now, for every new enrollment in this program one event will be auto generated. One program can also have multiple stages marked as auto generated.
-For all the auto generated events 
+### Enrollment with open data entry form
 
- a) the organisation unit will be the same as the user is reporting for, during the enrollment and 
- 
- b) all the events will be part of the current enrollment. 
+A program can be configured to have zero or more program stages that automatically open upon a new enrollment. The user will be taken to the new enrollment event page with the option to add an event in the first program stage that has "Open data entry form after enrollment" option set to true.
```
eirikhaugstulen, line 205: Hey!
I think the text `option to add an event` threw me a bit of the track here. The user is navigated directly to the dataentry-form here, right? Perhaps it would be simpler to just have 
`The user will be taken to the new enrollment event page in the first program stage that has "Open data entry form after enrollment" option set to true.`? Just something that I did not understand, but it might only be me. How do you feel about this @simonadomnisoru?

simonadomnisoru: Hey,
I agree, the sentence was too long and hard to understand. I simplify it and apply your suggestion. 
Thank you for the review! 😄

superskip: This is how I interpret the docs: (also a potential rephrase suggestion)
`A program can be configured to automatically open zero or more program stages upon creating a new enrollment. The program stage option "Open data entry form after enrollment" enables this behavior. The user will then be taken to a new event registration page immediately after creating a new enrollment. If more than one program stage has this option selected, the event registration page of the first such program stage is used.`

PR #2882, opened 2022-06-24
fix: [DHIS2-12303] enrollment dropdownbutton open state

simonadomnisoru:


src/core_modules/capture-core/components/WidgetEnrollment/Actions/Actions.component.js
```diff
@@ -32,57 +32,74 @@ export const ActionsPlain = ({
...
+                onClick={() => setOpen(prev => !prev)}
+                component={
+                    loading ? null : (
                         <FlyoutMenu dense maxWidth="250px">
                             <AddNew
                                 onlyEnrollOnce={onlyEnrollOnce}
                                 tetName={tetName}
                                 canAddNew={canAddNew}
-                                onAddNew={onAddNew}
+                                onAddNew={(arg) => {
```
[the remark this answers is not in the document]
simonadomnisoru: Good catch! I change the code to reflect your suggestion. Thanks!

src/core_modules/capture-core/components/WidgetEnrollment/Actions/Actions.component.js
```diff
@@ -32,17 +32,29 @@ export const ActionsPlain = ({
...
-                <span>
-                    {loading ? null : (
+}: PlainProps) => {
+    const [open, setOpen] = useState(false);
+    const handleOnUpdate = (arg) => {
+        setOpen(prev => !prev);
+        onUpdate(arg);
+    };
+    const handleOnDelete = (arg) => {
+        setOpen(prev => !prev);
```
jasminenguyennn, line 42: since delete action will navigate to dashboard page, do we need to close the menu?

simonadomnisoru: Hey,
Yes, the menu needs to be closed to handle the case when the Delete enrollment request fails. In this scenario, the user stays on the same page and the menu should be closed.
Thanks!

jasminenguyennn (approved):

PR #2890, opened 2022-06-30
chore: [TECH-1233] Refactor OrgUnitField and OrgUnitTree

jasminenguyennn:


src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/OrgUnitField.component.js
```diff
@@ -51,61 +47,104 @@ type Props = {
...
+
+    const ready = searchText?.length ? !searchLoading : !loading;
+
+    React.useEffect(() => {
+        if (searchText?.length) {
+            refetchOrg({ variables: { searchText } });
+        }
+    }, [refetchOrg, searchText]);
+
+    const getRoots = React.useMemo(() => {
```
simonadomnisoru, line 107: Hey, 

The error `rootNode is undefined` is thrown when I try to search for an org unit (write or delete some letters in the field). This error occurs in both the ScopeSelector and the forms. 
I think it may be related to the logic of the `getRoots` function. Can you have a look?



https://user-images.githubusercontent.com/6817073/176856318-b1c742b8-1f13-4344-8cfb-e20ed49c4c36.mov

Thank you!

simonadomnisoru (approved):
Looks like the `rootNode` issue was fixed by the latest commit. Everything else looks good also!

geethaalwan (approved):
Tested successfully o n 2.39 version

PR #2893, opened 2022-07-04
fix: [DHIS2-11514] use org unit code

simonadomnisoru:


eirikhaugstulen:
Hey, @simonadomnisoru!

Was testing this PR and noticed that you checked if orgUnit.code was set before initializing the registration form. Had to try the expected behavior on Play as well, and it seems like the registration form does not load if a orgUnitCode is not present. These codes should be optional I think, although I think they are often used. Is there something I have missed, or should this be treated as a bug? 
Thanks!

simonadomnisoru:
> Hey, @simonadomnisoru!
> 
> Was testing this PR and noticed that you checked if orgUnit.code was set before initializing the registration form. Had to try the expected behavior on Play as well, and it seems like the registration form does not load if a orgUnitCode is not present. These codes should be optional I think, although I think they are often used. Is there something I have missed, or should this be treated as a bug? Thanks!

Hey @eirikhaugstulen,

You are right the orgUnit code is optional. Right now, it can be used by the program rules or to generate unique data elements. It is up to admins to set it up correctly if they want to use it in these scenarios. I remove the check as you suggested.

Thank you! 😄

jasminenguyennn (approved):

JoakimSM (approved):
Seems good to me 👍

geethaalwan (approved):
Tested successfully on 2.36.13 version

PR #2897, opened 2022-07-07
fix: [DHIS2-13230] reimplement code for evaluating program rule expressions

superskip:
The previous implementation could not handle d2-functions with arbitrary expressions as parameters.


simonadomnisoru (approved):

JoakimSM (approved):
Really solid logic @superskip! Seems to work very well.

Since you are on vacation and I need to build on this code in another PR, I did refactor it a bit to make it more in line with our current functional approach and I also tried to make it easier to comprehend what is actually going on (i.e., refactored for readability and maintainability)

geethaalwan (approved):

PR #2898, opened 2022-07-07
fix: [DHIS2-11514] use org unit code

simonadomnisoru:


yarn.lock
```diff
@@ -2023,6 +2023,13 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
+"@babel/runtime@^7.15.4", "@babel/runtime@^7.9.2":
```
jasminenguyennn, line 2026: any reason why we modify yarn.lock file?

simonadomnisoru: Hey, 

The yarn.lock file is modified because I bump the [react-redux version](https://github.com/dhis2/capture-app/pull/2898/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R47) from `7.2.2` to `7.2.4`. In the`7.2.2` version, `shallowEqual` has some issues and it doesn't compare correctly the nested objects. Because of this, it doesn't properly trigger the [useSelector](https://github.com/dhis2/capture-app/pull/2898/files#diff-3410ffd10a40475b5d0b9c28f1b65eccfbd286a5143ba4757bfed0fabb2e3a80R12) hook. Updating react-redux and yarn solved these problems. 
Let me know if that makes sense.  

Thank you!

JoakimSM (approved):
LGTM

geethaalwan (approved):
Tested successfully on 2.36.13 version

PR #2900, opened 2022-07-07
fix: [DHIS2-13476] percentage value type

JoakimSM:
- Only allow integer values between 0 and 100


simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.37.8 version

PR #2906, opened 2022-07-12
chore: [TECH-1257] Clean LockedSelector

simonadomnisoru:
Clean as much as possible of the `LockedSelector` deprecated component. 

The only left logic is the epics chain that starts when the action `updateSelectionsFromUrl` is dispatched from `withAppUrlSync`. These epics and reducers check the validity of the URL parameters and hydrate `currentSelection` with data. After the pages are decoupled from `withAppUrlSync` and `currentSelection` the remaining `LockedSleector` logic can be safely deleted.


jasminenguyennn (approved):

geethaalwan (approved):

PR #2909, opened 2022-07-14
chore: jest config update

JoakimSM:
Fixed an issue where jest was using stale code for internal core modules. When importing modules from capture-core, capture-core-utils or capture-ui using the absolute path, jest would get the modules from the ".d2" folder (and ".d2" is only updated when running yarn start or yarn build).


simonadomnisoru (approved):
Good catch!

jasminenguyennn (approved):

PR #2915, opened 2022-07-19
feat: [DHIS2-13472] Deselecting reg. unit should expand tree to previous reg. unit

jasminenguyennn:
## Feature
[DHIS2-13472](): Deselecting org unit and go back should expand the org unit to high light previous selected org unit

### In the scope selector
Screenshot 2022-07-29 at 07 22 57

### In the form
Screenshot 2022-07-29 at 07 23 18

Demo: https://deploy-preview-2915--dhis2-capture.netlify.app/


src/core_modules/capture-core/reducers/descriptions/app.reducerDescriptionGetter.js
```diff
@@ -131,8 +131,8 @@ export const getAppReducerDesc = (appUpdaters: Updaters) => createReducerDescrip
         ...state,
         locationSwitchInProgress: true,
     }),
-    [lockedSelectorActionTypes.FROM_URL_UPDATE]: (state, action) => ({
-        ...state,
+    [lockedSelectorActionTypes.FROM_URL_UPDATE]: (state, action) => ({ ...state,
```
simonadomnisoru, line 134: Hey,

IMHO we should avoid adding more logic to the deprecated `LockedSelector` epics and reducers. I know most pages still use this logic to validate the URL parameters and hydrate `currentSelection` and `app` with data. As far as I know, we want to get rid of it soon and let each page handle its own initializations. There are pages where all the `LockedSelector` logic was already removed like the `/enrollment` and `/enrollmentEventEdit`. In these pages, the current feature implementation is not working. 

My proposal is to handle this new feature inside `OrgUnitField` and its parent components (`SingleOrgUnitSelectField` and `ScopeSelector/QuickSelector/OrgUnitSelector`). When the org unit is deselected from within the parents they can store the `previousOrgUnitId` in their state. After the deselection, the stored `previousOrgUnitId` can be passed down to `OrgUnitField` which can use the `usePreviousOrganizationUnit` hook to get the path.

Let me know if this makes sense and if the design gets complicated maybe we can discuss it more in our FE check-in meetings.

Thank you!

jasminenguyennn: hey @simonadomnisoru, thank you for your suggestion, I thought about setting the previous id when resetting the orgUnitId but the component is rerendered so the state is empty, so I have to based on the state in redux, currently there are still a lot of logic in `currentSelections` and `LockedSelector`, I know it's not the good idea to add the new logic into that but currently have no idea how that could work without relying on redux. 
Please let me know if that's clear or we can set up separate meeting. Thanks

src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/SingleOrgUnitSelectField.component.js
```diff
@@ -67,6 +77,7 @@ class SingleOrgUnitSelectFieldPlain extends React.Component<Props> {
 
     onDeselectOrgUnit = () => {
         this.props.onBlur(null);
+        this.props.value && this.setState({ previousOrgUnitId: this.props.value.id });
```
simonadomnisoru, line 80: Hey, 

Maybe the [lgtm alert](https://lgtm.com/projects/g/dhis2/capture-app/rev/pr-d281e681b7d6f46cf589008700fbd38045a79a5f) is fixed if `this.setState` is called first? Like this:
```
onDeselectOrgUnit = () => {
        this.props.value && this.setState({ previousOrgUnitId: this.props.value.id });
        this.props.onBlur(null);
}
```
Can you have a look? Thanks!

simonadomnisoru (approved):

geethaalwan (approved):

PR #2922, opened 2022-07-21
fix: [DHIS2-13486] encapsulate strings in double quotes

JoakimSM:


src/core_modules/capture-core-utils/rulesEngine/constants/typeKeys.const.js
```diff
@@ -11,13 +11,11 @@ export const typeKeys = {
     DATETIME: 'DATETIME',
     TIME: 'TIME',
     NUMBER: 'NUMBER',
-    UNIT_INTERVAL: 'UNIT_INTERVAL',
```
simonadomnisoru, line 14: It is great that you cleaned these unsupported types!

src/core_modules/capture-core-utils/rulesEngine/services/expressionService/injectVariableValues.js
```diff
@@ -0,0 +1,76 @@
...
+    const variableMatches = expression.match(/[A#CV]\{[\w -_]+}/g) || [];
+    // Inject the value for each matched variable
+    return variableMatches.reduce((accExpression, variableMatch) => {
+        // Remove pre/suf-fix to get the variable name for the variableHash
+        const variableName = variableMatch // lgtm [js/incomplete-sanitization]
+            .replace('#{', '')
+            .replace('A{', '')
+            .replace('C{', '')
+            .replace('V{', '')
+            .replace('}', '');
```
simonadomnisoru, line 66: NIT: Why not use `.replace(/}/g, '');` instead and get rid of the CodeQL error and the `// lgtm [js/incomplete-sanitization]` comment? 
The suf-fix logic will not be affected and in my opinion, the code is cleaner. Thanks!

simonadomnisoru (approved):
Looks good to me! 
I like that you've cleaned up and moved some of the code. The file structure and general logic of the RulesEngine are now easier to follow.

src/core_modules/capture-core-utils/rulesEngine/services/expressionService/injectVariableValues.js
JoakimSM: Hey @simonadomnisoru , Thanks for the review. I think my reasoning was that the variableMatch would always contain only 1 occurrence of the closing curly brace so there shouldn't be any reasons to look for more. It also follows the pattern from the other replacements here (replacing only the first occurrence). I agree it is not great to stick the lgtm suppression in there, but I think we might find it confusing reading the code later and one of the replacements differs from the others (and might wonder why that is).

geethaalwan (approved):

PR #2943, opened 2022-08-01
fix: [DHIS2-13534] edit single event and hide delete button

simonadomnisoru:


src/core_modules/capture-core/components/WidgetEventEdit/DataEntry/withDeleteButton.js
```diff
@@ -70,14 +71,14 @@ const getDeleteButton = (InnerComponent: React.ComponentType<any>) =>
...
-            const { onDelete, ...passOnProps } = this.props;
+            const { onDelete, hasDeleteButton, ...passOnProps } = this.props;
             const hasWriteAccess = this.props.formFoundation.access.data.write;
             return (
                 // $FlowFixMe[cannot-spread-inexact] automated comment
                 <InnerComponent
                     innerRef={(innerInstance) => { this.innerInstance = innerInstance; }}
                     // $FlowFixMe[extra-arg] automated comment
-                    deleteButton={this.renderDeleteButton(hasWriteAccess)}
+                    deleteButton={hasDeleteButton ? this.renderDeleteButton(hasWriteAccess) : null}
```
jasminenguyennn, line 81: I think we can clean up the `hasWriteAccess` here because it's not used in the method, maybe pass the `hasDeleteButton` as parameter?

jasminenguyennn (approved):

src/core_modules/capture-core/components/Pages/ViewEvent/EventDetailsSection/EventDetailsSection.container.js
```diff
@@ -8,6 +8,7 @@ import {
 
 const mapStateToProps = (state: ReduxState) => ({
     showEditEvent: state.viewEventPage.eventDetailsSection && state.viewEventPage.eventDetailsSection.showEditEvent,
+    programId: state.currentSelections.programId,
```
eirikhaugstulen, line 11: Hey @simonadomnisoru,
Can't really remember, but I have a vague memory that we wanted to move away from `currentSelections` as this might not be fully trusted to be update. This was the reason we removed the `connected-redux-router`-dependency.
Is this a concern that we're still having or is it okay to use in this case?

simonadomnisoru: Hey @eirikhaugstulen,

Indeed, we try to move away from `currentSelections`. But for this page and use case, I believe is safe to use it. This is because in a [previous commit](https://github.com/dhis2/capture-app/commit/0f23429c9a2a9e171e6e3ae1003d1008abb793bd#diff-3716f21e7b006a1c4cd91722cec7a4ae7ad50395c60ef78a2f9fb0dc30dcd560L32), I remove it from the epics chain. Now by adding it again I fix the bug that was introduced by the above commit. 

Thanks!

geethaalwan (approved):

src/core_modules/capture-core/components/Pages/ViewEvent/EventDetailsSection/EventDetailsSection.container.js
eirikhaugstulen: All fine by me, just wanted to check in! Looks good 👍

eirikhaugstulen (approved):

PR #2944, opened 2022-08-02
fix: [DHIS2-13181] change delete event endpoint to new tracker api

simonadomnisoru:


src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/eventList.epics.js
```diff
@@ -90,13 +89,22 @@ export const updateEventListEpic = (action$: InputObservable) =>
...
+) =>
     action$.pipe(
         ofType(actionTypes.EVENT_REQUEST_DELETE),
         concatMap((action) => {
             const { eventId, storeId } = action.payload;
-            const deletePromise = getApi()
-                .delete(`events/${eventId}`)
+            const deletePromise = mutate({
+                resource: 'tracker?async=false&importStrategy=DELETE',
+                type: 'create',
```
jasminenguyennn, line 103: should this be 'delete' instead of create?

simonadomnisoru: Hey @jasminenguyennn , 

Yes, it should be `delete`. Unfortunately, when the correct type is used, the `mutate` method does not support the property `data`, which is needed by the new tracker endpoint. This is the error thrown by the app platform and I reported it in [LIBS-340](https://dhis2.atlassian.net/browse/LIBS-340).

Screenshot 2022-08-02 at 16 10 41

Until this bug is fixed, the workaround we already apply to [delete an enrollment](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetEnrollment/Actions/Actions.container.js#L16) is to use the type `create`. 

Let me know if this makes sense and if you have another suggestion for a workaround. 
Thank you!

eirikhaugstulen (approved):
LGTM!

geethaalwan (approved):

PR #2945, opened 2022-08-04
fix: [DHIS2-13551] fallback missing trackedEntity field in the events api

simonadomnisoru:
Add fallback to maintain compatibility with 2.38 API

### DIFF
--- a/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
+++ b/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
@@ -2,13 +2,15 @@
 import { useMemo, useEffect, useState } from 'react';
 import { useDataQuery } from '@dhis2/app-runtime';
 
+const defaultState = {
+    program: undefined,
+    programStage: undefined,
+    enrollment: undefined,
+    trackedEntity: undefined,
+};
+
 export const useEvent = (eventId: string) => {
-    const [event, setEvent] = useState({
-        program: undefined,
-        programStage: undefined,
-        enrollment: undefined,
-        trackedEntity: undefined,
-    });
+    const [event, setEvent] = useState(defaultState);
     const { data, error, loading, refetch } = useDataQuery(
         useMemo(
             () => ({
@@ -41,7 +43,10 @@ export const useEvent = (eventId: string) => {
     );
 
     useEffect(() => {
-        eventId && refetch({ variables: { id: eventId } });
+        if (eventId) {
+            setEvent(defaultState);
+            refetch({ variables: { id: eventId } });
+        }
     }, [eventId, refetch]);
 
     useEffect(() => {
--- a/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
+++ b/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
@@ -1,8 +1,14 @@
 // @flow
-import { useMemo } from 'react';
+import { useMemo, useEffect, useState } from 'react';
 import { useDataQuery } from '@dhis2/app-runtime';
 
 export const useEvent = (eventId: string) => {
+    const [event, setEvent] = useState({
+        program: undefined,
+        programStage: undefined,
+        enrollment: undefined,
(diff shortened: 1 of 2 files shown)


jasminenguyennn (approved):

superskip (commented):
Hey! The interaction between destructured useMemo return values and useEffect's dependency array seems legit.

I think there might be a case missing: If the first provided eventId does not yield a trackedEntity object, the fallback trackedEntity is inserted. If another eventId is provided later, also not yielding a trackedEntity object, the fallback does not get inserted.

I think it might also be nice to include a flag in the return statement indicating whether or not fallback handling is currently in progress.

simonadomnisoru:
> Hey! The interaction between destructured useMemo return values and useEffect's dependency array seems legit.
> 
> I think there is a case missing: If the first provided eventId does not yield a trackedEntity object, the fallback trackedEntity is inserted. If another eventId is provided later, also not yielding a trackedEntity object, the fallback does not get inserted.
> 
> I think it might also be nice to include a flag for the return statement indicating whether or not fallback handling is currently in progress.

Hey @superskip,

Good catch. I update the code to handle the case of changing the `eventId`. Let me know if it looks good now.

Thanks!

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
```diff
@@ -9,26 +9,22 @@ export const useEvent = (eventId: string) => {
         enrollment: undefined,
         trackedEntity: undefined,
     });
-    const { data, error, loading } = useDataQuery(
+    const { data, error, loading, refetch } = useDataQuery(
         useMemo(
```
superskip, line 13: Nitpick: Might as well drop useMemo and just move the query object up above useEvent.

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/hooks/useEvent.js
```diff
@@ -9,26 +9,22 @@ export const useEvent = (eventId: string) => {
...
         ),
     );
 
-    const {
-        data: dataFallback,
-        refetch,
-        called,
-    } = useDataQuery(
+    const { data: dataFallback, refetch: refetchFallback } = useDataQuery(
         useMemo(
```
superskip, line 28: Nitpick: Same here.

superskip (commented):
Hey @simonadomnisoru !
Looks better, but you haven't addressed the point about returning undefined until fallback handling has completed. If you would like an idea on how it might be done, how about saying `let fallbackHandlingInProgress = false` at the beginning of the hook, set it equal `loading` from the fallback query, and also set it to true right after calling `refetchFallback`?

simonadomnisoru:
> Hey @simonadomnisoru ! Looks better, but you haven't addressed the point about returning undefined until fallback handling has completed. If you would like an idea on how it might be done, how about saying `let fallbackHandlingInProgress = false` at the beginning of the hook, set it equal `loading` from the fallback query, and also set it to true right after calling `refetchFallback`?
> 
> EDIT: The initial let statement can probably be left out, come to think of it.

Hey,
I push another commit that resets the event state to default each time the `eventId` changes. IMHO checking that all the props of `event` have a value is enough to know if the fallback has finished or not. This approach will work in both cases (when the fallback is needed and when is not needed) and this check is already done in the  [component page](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.js#L53). 

Let me know if that makes sense or if you prefer to have a call about it and discuss more.
Thank you!

superskip:
Hello @simonadomnisoru,
I suppose the enrollment endpoint is guaranteed to return a `trackedEntity`, which means you can indeed use this property as completion condition. :+1:
My gut tells me the right place to check this condition is in the return statement though, but that might admittedly not be the most reliable source of truth. Feel free to keep it this way if you think it is just as well.
(Apparently the nitpicks were sufficiently microscopic to be completely overlooked. Oh well :shrug:)

superskip (approved):

geethaalwan (approved):
Tested successfully on 2.38.2 version

PR #2947, opened 2022-08-11
fix: [DHIS2-13545] format updatedAt with the serverTimeZoneId

simonadomnisoru:


jasminenguyennn (approved):

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.component.js
```diff
@@ -41,6 +41,12 @@ const getGeometryType = geometryType =>
     (geometryType === 'Point' ? dataElementTypes.COORDINATE : dataElementTypes.POLYGON);
 const getEnrollmentDateLabel = program => program.enrollmentDateLabel || i18n.t('Enrollment date');
 const getIncidentDateLabel = program => program.incidentDateLabel || i18n.t('Incident date');
+const getLastUpdatedAt = (serverTimeZoneId, enrollment) => (
+    i18n.t('Last updated {{date}}', {
+        date: serverTimeZoneId
+            ? moment.parseZone(enrollment.updatedAt).utcOffset(serverTimeZoneId).fromNow()
```
JoakimSM, line 47: Looked into this a bit and as far as I can understand this won't work as expected when we actually change the timezone of the server to something else than utc. You can change the timezone of the server by running: `dpkg-reconfigure tzdata` (I'll help you out, let's slack)

I suggest we continue to use moment for now (moment is not recommended for new projects, read more about this in the moment docs). To get timezone compatibility with moment we will have to use the moment-timezone package (`https://momentjs.com/timezone/`). Something like `moment.tz(updatedAt, serverTimeZoneId).fromNow()` should do it I think (for importing I think `import moment from 'moment-timezone'`). Have a look and let me know if you have any question.

JoakimSM (approved):
Looks good

geethaalwan (approved):

PR #2948, opened 2022-08-16
fix: [DHIS2-13185] disable Edit event button when user doesn't have write access to the program stage

jasminenguyennn:


simonadomnisoru (commented):
Hey @jasminenguyennn,

The fix looks good. 
When the button is disabled do you think is useful for the user to have a `Tooltip` explaining the reason? Something like this: 

Screenshot 2022-08-17 at 09 31 25

Thanks!

simonadomnisoru (approved):

geethaalwan (approved):

PR #2976, opened 2022-08-24
fix: [DHIS2-13548] fix search with one attribute contains only space

jasminenguyennn:


simonadomnisoru (approved):
LGTM! It was a great idea to add a cypress scenario.

geethaalwan (approved):

PR #2981, opened 2022-08-26
fix: [DHIS2-13544] form validation for scheduled event

simonadomnisoru:
- load the `editEvent` data entry key in the redux store for the scheduled event. no need to load `viewEvent` in this case.
- split the edit form data entry ids in `singleEvent` and `enrollmentEvent` depending on the page.
- improve the cancel event callback logic


src/core_modules/capture-core/components/WidgetEventEdit/ViewEventDataEntry/viewEventDataEntry.actions.js
```diff
@@ -42,20 +42,23 @@ export const loadViewEventDataEntry =
...
         enrollment?: EnrollmentData,
         attributeValues?: Array<AttributeValue>,
     }) => {
-        const dataEntryId = viewEventIds.dataEntryId;
-        const itemId = viewEventIds.itemId;
         const dataEntryPropsToInclude = [
             {
                 id: 'occurredAt',
                 type: 'DATE',
+                validatorContainers: getEventDateValidatorContainers(),
```
superskip, line 61: The main point in this fix? No idea how it works, but I suppose it does :)

superskip (commented):
The new structure seems to have made some of the logic simpler, so that's quite all right.
The changes were mostly straight forward refactors, the only concern I have is whether it was done thoroughly (but I believe that's what it's always like when creating/reviewing a refactor).
However, before proceeding you should look into the naming convention issue I'm pointing out in the comment below.

superskip (approved):
LGTM!

src/core_modules/capture-core/components/WidgetEventEdit/ViewEventDataEntry/viewEventDataEntry.actions.js
simonadomnisoru: Hey @superskip, 

This line adds validation checks to the Report Date field. Now if the date doesn't have a valid value, an error message will appear below the field and the user will not be able to save the event until the error is fixed.  

Thanks for the review!

geethaalwan (approved):
Tested successfully on 2.39.0 and 2.38.2 versions

PR #2983, opened 2022-08-30
fix: temporary disable failing test related to DHIS2-13676

jasminenguyennn:


cypress/integration/ScopeSelector.feature
```diff
@@ -62,10 +62,11 @@ Feature: User uses the ScopeSelector to navigate
     And there should be visible a title with Child Program
     And there should be Child Programme domain forms visible to search with
 
-  Scenario: Main page > Having a program preselected, select an org unit which does not contain the program
-    Given you land on a main event page with preselected program
-    When you select org unit that is incompatible with the already selected program
-    Then you can see message on the scope selector
+  # TECH-1345
+  # Scenario: Main page > Having a program preselected, select an org unit which does not contain the program
```
simonadomnisoru, line 66: Hey, 
I already did a [fix](https://github.com/dhis2/capture-app/pull/2778/files#diff-9a9433ed9a2822d6876653c594b49bd8d2b4e5508c440c0fc35c0b758471ec07R379) for this test in one of my PRs. I believe is better to add the fix than to disable the test. WDYT?
Thanks!

simonadomnisoru (changes requested):

cypress/integration/ScopeSelector.feature
JoakimSM: I think we should fix the code at least. Might be better to comment out the test instead of doing a workaround in it (and re-enable when we fix the code)

simonadomnisoru (approved):

PR #2984, opened 2022-08-31
chore: [TECH-1345] fix cypress test

JoakimSM:
Follow up on https://github.com/dhis2/capture-app/pull/2983#discussion_r958359191

My interpretation was wrong here @simonadomnisoru. The problem was that the org unit was already visible of course. Your solution would definitely work, but to simplify it, I changed the test to select an org. unit that is not initially visible.


simonadomnisoru (approved):
Looks good!

jasminenguyennn (approved):

PR #2987, opened 2022-09-05
fix: [DHIS2-8210] use display order given in TEA registration form

superskip:
Programs may organise their associated tracked entity attributes into sections.
The order of the attributes however, is not taken into account by the app.
This fix aims to make the app start taking this aspect more seriously, i.e. arrange the attributes in the same order as seen in the section-definitions.


simonadomnisoru (approved):
LGTM!

JoakimSM (approved):
👍 Nice

geethaalwan (approved):
Tested successfully on 2.40,2.39.1,2.38.2 versions

PR #2994, opened 2022-09-06
feat: [DSHI2-13309] add schedule date in view/edit event page

jasminenguyennn:


simonadomnisoru (commented):
Hey @jasminenguyennn, 

Good job, overall it looks good 😄 . I have 2 comments:

1) Firstly, the `Schedule date` is displayed in the view and edit mode on the single event page. 




Screenshot 2022-09-09 at 10 47 37

2) Secondly, the `Due date` label and the tooltip `Go to “Schedule” tab to reschedule this event` appear for enrollment events that have the status `active`.
Screenshot 2022-09-09 at 10 57 59


As far as I understand this feature is only for the enrollment events with the status `schedule` or `overdue`. Can you have a look? 
Thanks!

jasminenguyennn:
Thanks @simonadomnisoru good catch! I updated the branch to hide the schedule date on condition.

simonadomnisoru (commented):
Hey,
I have a few comments about the flow of this feature:
1) When I click the `Scheadule` button, it looks like nothing happens. I expected to be redirected back to the Enrollment dashboard in the same way it happens when a new event is scheduled. 
2) I go back to the Enrollment dashboard using the `Back to all stages and event` button and I cannot see any change in the program stage table. When I refresh the page I see a new event on the table. The Redux store and `enrollmentDomain` can be used to be able to see the data without having to refresh the page. The same way is done when adding/editing an event. 
3) Currently a new event is scheduled each time the `Scheadule` button is clicked. I may be wrong on this one, but as far as I understood from the design docs there shouldn't be a new event added. Instead, the event `Due date` value should be edited and the same event is rescheduled with the new date. 

In the video below you can see the steps I took. Can you have a look?

https://user-images.githubusercontent.com/6817073/190341351-7f6519ff-b355-4228-9601-8f6803f1c822.mov

Thank you!

jasminenguyennn:
thanks @simonadomnisoru. I initially thought the handling of the schedule event will be in a different ticket but it makes sense to do so in the same ticket. So I updated the branch with the handling event accordingly. 
Please have a look. Thanks

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPage.epics.js
```diff
@@ -35,3 +38,20 @@ export const saveNewEventFailedEpic = (action$: InputObservable) =>
             return batchActions([saveFailed(), rollbackEnrollmentEventWithoutId(meta.uid)]);
         }),
     );
+
+export const updateEventSucceededEpic = (action$: InputObservable) =>
```
simonadomnisoru, line 42: Hey, 
Is there any reason why `updateEventSucceededEpic` and `updateEventFailedEpic` epics are inside `EnrollmentAddEventPage.epics.js`? Because the re-schedule feature is on the `EnrollmentEditEvent` page I believe they should be moved to the edit page folder. 
WDYT?

src/core_modules/capture-core/components/WidgetEventEdit/DataEntry/editEventDataEntry.actions.js
```diff
@@ -101,6 +101,10 @@ export const openEventForEditInDataEntry = ({
             type: 'DATE',
             validatorContainers: getEventDateValidatorContainers(),
         },
+        {
+            id: 'scheduledAt',
+            type: 'DATE',
```
simonadomnisoru, line 106: Does it make sense to also add here the same validator you added in the [getOpenDataEntryActions](https://github.com/dhis2/capture-app/pull/2994/files#diff-a2f9ba1fd93edf38289833e1bddf0591b4f7acde3eae1d3a90e3d68769946034R19)?

src/core_modules/capture-core/components/WidgetEventEdit/WidgetEventEdit.container.js
```diff
@@ -65,6 +72,10 @@ export const WidgetEventEditPlain = ({
     if (error) {
         return error.errorComponent;
     }
+    const onHandleScheduleSave = (eventId: string, eventData: Object) => {
```
simonadomnisoru, line 75: Hey,
I suggest we avoid navigating from inside the widget and let the page deal with all the redirections. The same way it is currently done by other props of the `WidgetEventEdit` widget like [onGoBack](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.js#L77) and [onCancelEditEvent](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.js#L73).
Thanks!

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.epics.js
```diff
@@ -2,10 +2,14 @@
...
+import {
+    scheduleEventWidgetActionTypes,
+    scheduleEvent,
+    updateScheduledDateForEvent,
+} from './WidgetEventSchedule.actions';
+import { statusTypes } from '../../events/statusTypes';
 
-
-export const scheduleNewEnrollmentEventEpic = (action$: InputObservable) =>
+export const scheduleNewEnrollmentEventEpic = (action$: InputObservable, store: ReduxStore) =>
```
simonadomnisoru, line 12: NIT: Both schedule and re-schedule will use this function so maybe `scheduleEnrollmentEventEpic` is a more suitable name for this epic.

src/core_modules/capture-core/components/WidgetEventSchedule/WidgetEventSchedule.epics.js
```diff
@@ -18,16 +22,21 @@ export const scheduleNewEnrollmentEventEpic = (action$: InputObservable) =>
...
                 onSaveSuccessActionType,
                 onSaveErrorActionType,
             } = action.payload;
 
+            const { events } = store.value;
+            const existingEnrollment = events[eventId]
+            && [statusTypes.SCHEDULE, statusTypes.OVERDUE].includes(events[eventId].status);
 
             const serverData = { events: [{
                 scheduledAt: scheduleDate,
```
simonadomnisoru, line 36: Hey, 
I am having troubles re-scheduling the event. I get a 500 error that mentions the invalid format of the date.

Screenshot 2022-09-21 at 10 22 05


I tried both types of calendars that we currently support. Also when I use the `dd-mm-yyyy` calendar I get an invalid value. What am I doing wrong?

Screenshot 2022-09-21 at 10 21 29


Screenshot 2022-09-21 at 10 35 31

Thank you!

simonadomnisoru (approved):
Looks great! Thanks for looking at my suggestions!

geethaalwan (approved):
Tested successfully on 2.39.0 ,2.39.1,2.40 versions

PR #2995, opened 2022-09-09
feat: [DHIS2-13661] program stage working list default filters

simonadomnisoru:


jasminenguyennn:
Hi @simonadomnisoru, great work in this PR 🎉
I only have few comments
- When program stage is selected, both Report data and Event status stay disabled 
Screenshot 2022-09-12 at 09 45 19
- nitpick:  the label should be consistent `Event status` instead of `Event Status`

src/core_modules/capture-core-utils/featuresSupport/support.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+export const FEATURES = {
```
jasminenguyennn, line 2: I think we should freeze the object

simonadomnisoru:
> Hi @simonadomnisoru, great work in this PR 🎉 I only have few comments
> 
> * When program stage is selected, both Report data and Event status stay disabled
> 
> Screenshot 2022-09-12 at 09 45 19
> 
> - nitpick: the label should be consistent `Event status` instead of `Event Status`

Hey @jasminenguyennn, 

Thanks for the review! I implemented the small suggestions. 
Regarding the comment about what happens when a program stage is selected, that part of the feature has not been implemented yet. I am currently working to analyze it [DHIS2-13662](https://dhis2.atlassian.net/browse/DHIS2-13662) and I will create more tickets and PRs in the next few days. 
Thanks!

jasminenguyennn (approved):
👍  LGTM

PR #3001, opened 2022-09-13
fix: [DHIS2-13723] send program id from event

jasminenguyennn:


src/core_modules/capture-core/components/CardList/CardListItem.component.js
```diff
@@ -94,9 +94,15 @@ const deriveEnrollmentType =
   };
 
 const deriveEnrollmentOrgUnitAndDate =
-  (enrollments, enrollmentType, currentProgramId): {orgUnitName?: string, enrolledAt?: string} => {
-      if (!currentProgramId) {
-          return {};
+  (enrollments, enrollmentType, currentProgramId): {orgUnitName?: string, enrolledAt?: string, programFromEvent?: string} => {
+      if (!enrollments?.length) { return {}; }
+      if (!currentProgramId && enrollments.length) {
+          const { orgUnitName, enrolledAt, program: programFromEvent } = enrollments[0];
```
simonadomnisoru, line 100: NIT: Because the variable is obtained from `enrollments[0]`, I believe it should be called`programFromEnrollment` rather than `programFromEvent`.

simonadomnisoru (approved):

geethaalwan (commented):
Tested successfully on 2.39.0 and 2.38.2 versions

geethaalwan (approved):

PR #3005, opened 2022-09-15
chore: [TECH-862] [TECH-1409] fix cypress tests

JoakimSM:
Fixes Cypress tests listed in [TECH-862](https://dhis2.atlassian.net/browse/TECH-862) and other failing tests previously listed in our wiki.


cypress/integration/SearchPage.feature
```diff
@@ -1,248 +1,217 @@
...
-#    Given you are in the search page with the Child Programme being preselected from the url
-#    And you expand the attributes search area
-#    When you fill in the first and last name with values that will return results
-#    And you click search
-#    And you can see the first page of the results
-#    When you click the fallback search button
-#    Then you see the attributes search area being expanded
-#    And that first and last name are prefilled
-
-  Scenario: Fallback search excludes attributes that are not relevant in the tetype scope
```
simonadomnisoru, line 222: Hey,
Any reason why the scenarios `Fallback search excludes attributes that are not relevant in the tetype scope` and `Fallback search with results and navigation back to the program show` were deleted? They were not disabled beforehand which means that in theory, they are valid tests. 
Thank you!

cypress/integration/SearchPage.feature
```diff
@@ -1,248 +1,217 @@
...
-  Scenario: Fallback search with results
-    Given you are in the search page with the Child Programme being preselected from the url
-    And you expand the attributes search area
-    When you fill in the first and last name with values that will return results
-    And you click search
-    And you can see the first page of the results
-    When you click the fallback search button
-    Then you stay in the same page with all program results
-
-# TODO - Commenting out this test and creating a new tech-ticket (TECH-866)
```
simonadomnisoru, line 210: The ticket `TECH-866` is still open. Should we either close that ticket or try to find a solution to it?

cypress/integration/AllAccessibleRecordsPage.feature
```diff
@@ -24,11 +24,3 @@ Feature: the user interacts with the "view all records regardless of OrgUnit"-pa
   Scenario: the working list will hydrate when updated by the URL
     When the user navigates to /#/?programId=VBqh0ynB2wv&all
     Then the working list should be displayed
-
-#  Scenario: The user can edit an event without selecting OrgUnit
```
simonadomnisoru, line 28: Hey,
Any reason why this scenario was removed? 
Thanks!

JoakimSM: Good catch. Re-added, fixed and enabled.

cypress/integration/SearchPage.feature
JoakimSM: Thanks, ticket closed. The test scenario is no longer valid (the fallback workflow has changed)

simonadomnisoru (approved):

PR #3008, opened 2022-09-21
feat: [DHIS2-13739] navigate to new tracker based on opt-in func

jasminenguyennn:


src/core_modules/capture-core/utils/routing/newDashboard.js
```diff
@@ -0,0 +1,2 @@
+export const shouldUseNewDashboard = (userDataStore, dataStore, programId) =>
```
simonadomnisoru, line 1: Hey,
Great idea to move this helper function to the `/utils` folder. Does it make sense to import and use it everywhere, like in [navigateToEnrollmentOverview](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/actions/navigateToEnrollmentOverview/navigateToEnrollmentOverview.epics.js#L37) and [RegistrationDataEntry](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js#L42)?
Thanks!

JoakimSM: Makes sense! I refactored the code since @jasminenguyennn is off today (and we will make an effort to get this included in the 2.39.0 bundled version)

geethaalwan (commented):
Tested successfully on 2.39.0,2.38.2,2.40 versions

geethaalwan (approved):

PR #3020, opened 2022-09-29
feat: [DHIS2-13681] optional decimal places argument for d2:round

superskip:


src/core_modules/capture-core-utils/rulesEngine/d2Functions/getD2Functions.js
```diff
@@ -21,8 +21,14 @@ export const getD2Functions = ({
         execute: (params: any) => Math.floor(params[0]),
     },
     round: {
-        parameters: 1,
-        execute: (params: any) => Math.round(params[0]),
+        parameters: { min: 1, max: 2 },
+        execute: (params: any) => {
+            if (params[1]) {
+                const decimalCorrection = 10 ** Math.round(params[1]);
+                return Math.round((params[0] + Number.EPSILON) * decimalCorrection) / decimalCorrection;
```
simonadomnisoru, line 28: NIT: probably `return +Math.round(params[0]).toFixed(params[1])` also works

simonadomnisoru (approved):
LGTM!

jasminenguyennn (approved):

src/core_modules/capture-core-utils/rulesEngine/d2Functions/getD2Functions.js
superskip: Thanks for the review!
I think your solution is what we actually want here :+1:

superskip: Oops, no, it doesn't work.. `params[1] == 1` means 9.28 should be rounded to 9.3, and that can't be done with `Math.round` unless one first scales the number.

geethaalwan (approved):

PR #3032, opened 2022-10-05
feat: [DHIS2-13672] add referral actions to add event widget

jasminenguyennn:


simonadomnisoru (approved):
LGTM! 😄 

ps: I couldn't find in the demo database a relationship with `referral` set to `true`. Therefore is not possible to add a cypress test for now. Let's keep this in mind and go over it again once the whole Referral feature is developed. Thanks!

geethaalwan (approved):

src/core_modules/capture-core/components/WidgetReferral/useReferral.js
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }, [relationshipTypeData]);
+
+    useEffect(() => {
+        if (selectedRelationshipTypes.length === 1) {
+            const { fromConstraint, toConstraint } = selectedRelationshipTypes[0];
+
+            if ((fromConstraint?.programStage?.id === programStageId
+                || toConstraint?.programStage?.id === programStageId)
+                && programStageData?.programStageReferral?.referral
```
JoakimSM, line 45: We shouldn't check for this flag (programStageData.programStageReferral.referral). Its use has not been determined yet. Checking the relationship type should be enough for now.

src/core_modules/capture-core/components/WidgetReferral/useReferral.js
```diff
@@ -0,0 +1,65 @@
...
+    }), [programStageId]));
+
+    const { data: relationshipTypeData } = useDataQuery({
+        relationshipTypes: {
+            resource: 'relationshipTypes',
+            params: {
+                fields: 'id,access[*],referral,fromConstraint[*],toConstraint[*]',
+            },
+        },
+    });
```
JoakimSM, line 28: Can we grab the relationship metadata from the indexedDB cache instead? The useDataQuery will currently re-download the data every time you open the new event page. The new event and edit event page are not self-contained at all so I think it's fine if you just use the indexedDB cache from here.

JoakimSM (approved):
Looks good 👍

PR #3054, opened 2022-10-11
feat: [DHIS2-13897] respect Hide due date flag

jasminenguyennn:


src/core_modules/capture-core/components/WidgetEventSchedule/hooks/useDetermineSuggestedScheduleDate.js
```diff
@@ -51,17 +51,36 @@ type Props = {
...
 }
+
+const calculateSuggestedDateFromStart = ({
+    generatedByEnrollmentDate,
+    displayIncidentDate,
+    enrolledAt,
+    occurredAt,
+    minDaysFromStart,
+}: Object) => {
+    let suggestedScheduleDate;
```
simonadomnisoru, line 65: NIT: Use `const` instead of `let`.
```
    const suggestedStartDate = (generatedByEnrollmentDate || !displayIncidentDate)
        ? enrolledAt : occurredAt;
    const suggestedScheduleDate = moment(suggestedStartDate).add(minDaysFromStart, 'days').format();
    return suggestedScheduleDate;
```

simonadomnisoru (commented):
Hey @jasminenguyennn, 

Overall it looks really good! I only have 2 comments.
Firstly, does it make sense to respect the `hideDueDate` flag when a new event is scheduled from the `enrollmentEventNew` page? 

Secondly, can you add a cypress scenario for the feature? The program Malaria focus investigation with programId `M3xtLkYBlKI` has stages with the flag 'Hide due date' set to true and can be used for the test.

Thank you!

src/core_modules/capture-core/components/WidgetEventSchedule/hooks/useDetermineSuggestedScheduleDate.js
jasminenguyennn: thanks @simonadomnisoru but I think I'd keep the `let` here to emphasise on the logic of calculating the suggested date.

simonadomnisoru (approved):

geethaalwan (approved):

PR #3055, opened 2022-10-11
fix: [DHIS2-13906] variable names with spaces in rules engine

JoakimSM:
Fixes a problem where program rule variable names with spaces don't work when used in some d2 functions.


simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.38.2,2.39.1,2.39.0,2.40 versions

PR #3058, opened 2022-10-12
feat: [DHIS2-13895] show scheduled date for active/complete events

jasminenguyennn:


src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js
```diff
@@ -160,17 +168,14 @@ const buildScheduleDateSettingsFn = () => {
...
             width: '100%',
             calendarWidth: 350,
             label: props.formFoundation.getLabel('scheduledAt'),
             disabled: true,
         }),
-        getIsHidden: (props: Object) => {
-            const isScheduleableStatus = [statusTypes.SCHEDULE, statusTypes.OVERDUE].includes(props.eventStatus);
-            return props.hideDueDate || !isScheduleableStatus;
-        },
+        getIsHidden: (props: Object) => props.hideDueDate,
```
simonadomnisoru, line 178: Hey, 

Should the Scheduled date also be shown in Event programs? If not one way to fix it is to check that the event belongs to a Tracker program by looking at the `props.id`. Something like this: `getIsHidden: (props: Object) => props.id === dataEntryIds.ENROLLMENT_EVENT && props.hideDueDate`

Screenshot 2022-10-18 at 09 21 53

Thanks!

jasminenguyennn: hey good catch @simonadomnisoru, I updated the branch to reflex the changes now. thanks for looking into it

simonadomnisoru: Hey,
I re-tested and I can see still the scheduled date in the Event programs. The logic I previously suggested was wrong 😓
I believe something like this should work better:
`getIsHidden: (props: Object) => props.id !== dataEntryIds.ENROLLMENT_EVENT || props.hideDueDate`. 

WDYT? Thanks! 🙏

jasminenguyennn: thanks @simonadomnisoru! I verified it should be hidden now on the view event page
Screenshot 2022-10-19 at 09 09 06

simonadomnisoru (approved):

geethaalwan (approved):

PR #3062, opened 2022-10-13
feat: [DHIS2-13661][DHIS2-13747][DHIS2-13748][DHIS2-13749][DHIS2-13750] program stages working list

simonadomnisoru:
In this PR, I've incorporated the tickets needed to release the first iteration of the program stage's worklist feature. Main logic:  
- feature toggling based on the API version
- add program stage working list default filters  
- extend TEIWorkingList to support tracker/events results  
- filter by both TEA and data elements  
- sort by both TEA and data elements  
- remove program stage filter button


src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/Setup/useDefaultColumnConfig.js
```diff
@@ -35,7 +35,37 @@ const mainConfig: Array<MainColumnConfig> = [{
...
+    visible: false,
+    type: dataElementTypes.TEXT,
+    header: i18n.t('Event status'),
+    filterHidden: true,
+},
+{
+    id: 'occurredAt',
+    visible: false,
+    type: dataElementTypes.DATE,
+    header: i18n.t('Report date'),
```
jasminenguyennn, line 49: if the report date label has been configured in the maintenance app, we should display it here.

jasminenguyennn (commented):
Great works 🎉  everything works well, I just have a concern why we can not combine the report date filter with other filters (except the program stage)? 

https://user-images.githubusercontent.com/89806888/196112777-5a680963-6223-40b9-b49a-1b7a6e131ce1.mov

simonadomnisoru:
> Great works 🎉 everything works well, I just have a concern why we can not combine the report date filter with other filters (except the program stage)?

Hey @jasminenguyennn 
Good catch! The report date filter should work in combination with all other filters and my last commit fixed this issue. 
Thanks for the review! 😄

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/Setup/TeiWorkingListsSetup.component.js
```diff
@@ -136,6 +138,95 @@ const useFiltersOnly = ({
         ];
     }, [enrollmentDateLabel, incidentDateLabel, showIncidentDate, stages]);
 
+
+const useProgramStageFilters = ({ stages }: TrackerProgram, programStageId?: string) => {
+    const supportsProgramStageWorkingLists = useFeature(FEATURES.programStageWorkingList);
+    const programStage = programStageId && stages.get(programStageId);
+    const occurredAtLabel = programStage
+        // eslint-disable-next-line no-underscore-dangle
+        ? programStage.stageForm._labels.occurredAt
```
jasminenguyennn, line 147: hey @simonadomnisoru I think we can use `stageForm.getLabel('occurredAt')` here instead of calling the private variables. WDYT?

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/constants/additionalFilters.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+export const ADDITIONAL_FILTERS = {
+    programStage: 'programStage',
+    occurredAt: 'eventOccurredAt',
```
jasminenguyennn, line 4: Hey @simonadomnisoru, I noticed that the `report date` is not showing up in the list, this is due to unmatched term here, I believe changing this to `occurredAt` and add the field to the query should fix the problem
Screenshot 2022-10-19 at 09 31 25

simonadomnisoru: Hey @jasminenguyennn 

The `occurredAt` id is already used by the TEI filters. The list of filters doesn't work with duplicate filter ids. I need to use a different id for the program stage filters and that is why I choose to use `eventOccurredAt`. 

There is a converter to the API form when [filtering by report date](https://github.com/dhis2/capture-app/pull/3062/files#diff-f5af0891c6740a4d8327fd8aadb1c63a8fc5e99b84fb715312835f2bf180a38dR169). In my last commit, I also added a converter for ordering by report date and one for the report date values in the list. 

Let me know if it works fine now or if you find any other issues. Thank you!

jasminenguyennn (approved):
LTGM 👍  thanks for looking.

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/Setup/useDefaultColumnConfig.js
```diff
@@ -35,7 +36,43 @@ const mainConfig: Array<MainColumnConfig> = [{
...
+        {
+            id: ADDITIONAL_FILTERS.occurredAt,
+            visible: false,
+            type: dataElementTypes.DATE,
+            header: stages.get(programStageId)?.stageForm.getLabel('occurredAt')
+                || i18n.t(ADDITIONAL_FILTERS_LABELS.occurredAt),
+            filterHidden: true,
+            additionalColumn: true,
+        }]
+            .map(field => ({
```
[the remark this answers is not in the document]
simonadomnisoru: Hey,
Yes, the idea is to keep a single source of truth. This mapping for `programStageMainConfig` is following the same pattern as the [mainConfig](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/Setup/useDefaultColumnConfig.js#L33) one. 
In my latest commit, I also added the `filterHidden` and `additionalColumn` as you suggested.
Thanks for the review!

geethaalwan (approved):

PR #3065, opened 2022-10-17
feat: [DHIS2-13801] ask to create new after completing the event

jasminenguyennn:


src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -190,3 +193,23 @@ export const requestDeleteEventDataEntryEpic = (action$: InputObservable, store:
...
+    action$.pipe(
+        ofType(
+            actionTypes.START_CREATE_NEW_AFTER_COMPLETING,
+            newEventWidgetActionTypes.START_CREATE_NEW_AFTER_COMPLETING,
+        ),
+        flatMap((action) => {
+            const { isCreateNew, enrollmentId } = action.payload;
+            return from(querySingleResource({ resource: 'tracker/enrollments', id: enrollmentId })).pipe(
+                switchMap(({ trackedEntity, program, orgUnit: orgUnitId }) => {
+                    const params = { enrollmentId, orgUnitId, programId: program, teiId: trackedEntity };
```
simonadomnisoru, line 207: Hey,
According to the Jira description, if there is only one possible stage available, the user should be saved a click and taken directly to the New event workspace for that stage. For this use case, the `stageId` can be added to the URL parameters to achieve the functionality. WDYT?
Thank you!

src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -190,3 +193,23 @@ export const requestDeleteEventDataEntryEpic = (action$: InputObservable, store:
...
+export const startCreateNewAfterCompletingEpic = (
+    action$: InputObservable, store: ReduxStore, { querySingleResource, history }: ApiUtils) =>
+    action$.pipe(
+        ofType(
+            actionTypes.START_CREATE_NEW_AFTER_COMPLETING,
+            newEventWidgetActionTypes.START_CREATE_NEW_AFTER_COMPLETING,
+        ),
+        flatMap((action) => {
+            const { isCreateNew, enrollmentId } = action.payload;
+            return from(querySingleResource({ resource: 'tracker/enrollments', id: enrollmentId })).pipe(
```
simonadomnisoru, line 205: Hey,
Does it make sense to remove the call to the `tracker/enrollments` API endpoint? All the ids needed to build the page URL (orgUnitId, programId, teiId) are already in the components and can be passed to the `START_CREATE_NEW_AFTER_COMPLETING` action and then to the `startCreateNewAfterCompletingEpic` epic. 

I created a [patch](https://github.com/dhis2/capture-app/compare/DHIS2-13801...example-remove-endpoint-call?expand=1) to showcase one way how this can be done. WDYT?

src/core_modules/capture-core/components/DataEntry/withAskToCreateNew.js
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        getWrappedInstance() {
+            return this.innerInstance;
+        }
+
+        handleOnSave(eventId: string, dataEntryId: string, formFoundation: RenderFoundation, saveType?: string) {
+            if (this.props.allowGenerateNextVisit &&
+                (this.props.isCompleted || saveType === addEventSaveTypes.COMPLETE)) {
+                this.setState({ isOpen: true });
```
simonadomnisoru, line 40: Hey,
Should the modal be shown if there are no available stages where a new event can be added? This edge case occurs when the following conditions are met:
- all stages have `repeatable` false 
- all stages already have an event
- the stage where the event is completed has `allowGenerateNextVisit` true.  

https://user-images.githubusercontent.com/6817073/199425964-274a65a1-4c17-4ba0-8367-0ac16afd2d4f.mov


Does it make sense to check that there is at least one available stage? 
Thanks!

src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.actions.js
```diff
@@ -68,3 +69,6 @@ export const startDeleteEventDataEntry = (eventId: string, params: Object) =>
         },
     });
 
+
+export const startCreateNewAfterCompleting = ({ enrollmentId, isCreateNew, orgUnitId, programId, teiId }: Object) =>
```
simonadomnisoru, line 73: Hey, 
Any reason why `availableProgramStages` is missing from both the action parameters and the `actionCreator` function below? 
Thanks!

src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/validated.actions.js
```diff
@@ -66,3 +67,5 @@ export const saveEvent = (serverData: Object, onSaveSuccessActionType?: string,
             rollback: onSaveErrorActionType && { type: onSaveErrorActionType, meta: { serverData, uid } },
         },
     });
+export const startCreateNewAfterCompleting = ({ enrollmentId, isCreateNew, orgUnitId, programId, teiId, availableProgramStage }: Object) =>
```
simonadomnisoru, line 70: Hey,
Based on the [parameters](https://github.com/dhis2/capture-app/pull/3065/files#diff-219843c9418d6aab189726c11f8623961fa99a504e1a74a0cae067f004169d1cR98) used to call the action, I believe `availableProgramStage` has a small typo and is missing `s` at the end. 
Should it be `availableProgramStages` instead?
Thanks!

src/core_modules/capture-core/hooks/useAvailableProgramStages.js
```diff
@@ -0,0 +1,41 @@
...
+    if (enrollmentsError || programError) {
+        log.error(errorCreator('Widget Event Edit could not be loaded')({
+            enrollmentsError,
+            programError,
+        }));
+    }
+    const availableProgramStages = useMemo(() => programStage.allowGenerateNextVisit
+    && !programLoading && program?.programStages
+        ?.reduce((accStage, currentStage) => {
+            if (currentStage.id !== programStage.id) {
```
simonadomnisoru, line 26: Why do you exclude the currently selected `programStage.id` from the `availableProgramStages`? 
The modal is not shown when the only available stage is the one where an event is completed.

https://user-images.githubusercontent.com/6817073/199670448-e4333fc4-f166-4709-8955-46c05bd10996.mov

Thanks!

simonadomnisoru (commented):
Hey,
Can you add a Cypress scenario for this feature? The stage Foci investigation & classification from the program Malaria focus investigation has both `repeatable` and `allowGenerateNextVisit` set to `true` and it can be used to test against.  
Thank you!

src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -190,3 +193,23 @@ export const requestDeleteEventDataEntryEpic = (action$: InputObservable, store:
...
+            const { isCreateNew, enrollmentId, orgUnitId, programId, teiId, availableProgramStages } = action.payload;
+            const params = { enrollmentId, orgUnitId, programId, teiId };
+
+            if (isCreateNew) {
+                const finalParams = availableProgramStages.length === 1 ?
+                    { ...params, stageId: availableProgramStages[0].id } : params;
+                history.push(`/enrollmentEventNew?${buildUrlQueryString(finalParams)}`);
+                return EMPTY;
+            }
+            return of(navigateToEnrollmentOverview(params));
```
simonadomnisoru, line 213: NIT: 
As far as I am aware, the user has to go through the enrollment dashboard to reach the pages where it can add/edit an enrollment event. Therefore we can assume that the admin has already opted-in to use the enrollment dashboard for that program.
Based on this, I believe there is no need to check where to navigate, and is safe to go directly to the enrollment page like this: 
```
    history.push(`/enrollment?${buildUrlQueryString(params)}`);
    return EMPTY;
```

simonadomnisoru (approved):

geethaalwan (approved):

PR #3084, opened 2022-10-31
fix: [DHIS2-13941] adjust notice box for schedule events

jasminenguyennn:


simonadomnisoru (approved):
LGTM!

karolinelien (changes requested):
Seems like there is still a line break between the two sentences (ref Joe's uggestions in the issue)

karolinelien (approved):

PR #3087, opened 2022-11-01
fix: [DHIS2-13346] assigned value

jasminenguyennn:


JoakimSM (commented):
I had no success with this solution when I tested it (used a rule to set a compulsory field when another field was filled in). Nevertheless, looking into it, we should approach this a bit differently.

Currently we save the result of a field validation in the Redux store when you exit a field ("onblur" for most of the field types). If you use the "assign" action in the rules engine, we presume that you fill in a valid value and that value is not validated (The better approach here would have been to validate on render. This is a big change however, one we should carefully consider the implications of before doing it)

My suggestion is to still presume that any value you set using the "assign" action in the rules engine is valid. There is a missing implementation detail here however. Any validation errors that existed prior to the assignment is not cleared. So, we will have to reset the relevant parts of the `formsSectionsFieldsUI` in the Redux store for that field. You should be able to add this logic in the reducer.

The `formsSectionsFieldsUI` is grouped by sections, so we will have to look in multiple groups. The relevant properties are: `valid`, `errorMessage`, `errorData` and `validatingMessage` (I think)

Let me know if you have questions.

src/core_modules/capture-core/components/DataEntries/Enrollment/actions/enrollment.actionBatchs.js
```diff
@@ -38,7 +39,7 @@ export const runRulesOnUpdateFieldBatch = (
     });
 
     return batchActions([
-        updateRulesEffects(effects, formId),
+        updateRulesEffects(effects, formId, formBuilderId),
```
simonadomnisoru, line 42: Hey,
Looks like the bug is fixed for the newEnrollment form. In order to fix the same error in the rest of the forms, does it make sense to use `formBuilderId` every time `updateRulesEffects` is called?
Thanks!

jasminenguyennn: hey @simonadomnisoru , I'm not entirely sure what do you mean why the rest of the form, can you give an example?

simonadomnisoru: Sure, some examples that come to mind from the rest of the forms are: 

**1) Add an event in a tracker program** 
add new event in tracker programs

**2) Edit an event in a tracker program** 
edit event in tracker programs

**3) Edit the TEI attributes**
edit tei attributes 

**4) Add an event in an event program** 
add new event in event program

It seems that the assigned value is also not treated as an entered value in these scenarios. I believe if `formBuilderId` is used every time `updateRulesEffects` is called, then the error would be fixed in all forms.

Let me know if there are any further doubts. Thanks!

jasminenguyennn: hey @simonadomnisoru 
I can not reproduce the problem on the stated scenarios, can you share how you set up the program rules? thanks!



Screenshot 2023-01-24 at 15 50 22
Screenshot 2023-01-24 at 15 50 02

JoakimSM (commented):
We can't use `formBuilderId` I believe. That is because the assignment can refer to a field in another section. We will have to look for the data element / tracked entity attribute in all sections that is starting with the `formId`.

simonadomnisoru (approved):

JoakimSM (approved):
Looks good 👍

geethaalwan (approved):

PR #3089, opened 2022-11-03
refactor: [TECH-1283] default values for program rule variables

superskip:
This is a refactor of `VariableService`, the module which takes care of populating program rule variables inside the rules engine.
The goal in this patch is to introduce a standardized way of representing "empty" variables, and subsequently make sure such variables gets assigned a default value.

The new population process goes like this:
When creating a new program rule variable, the source value first gets processed by `inputConverter` (which in `VariableService` goes by the name `onProcessValue`).
The input converters have been modified so that they return `null` whenever the source value is undefined.
This enables `VariableService` to treat a variable as not having a value whenever the processed value equals `null`.
These processed values are then passed as arguments to `buildVariable`.
The variable built by `buildVariable` gets assigned a default value if the value given as input equals `null`.
Also the property `hasValue` is set based on the same condition.


src/core_modules/capture-core/rules/converters/inputConverter.js
```diff
@@ -6,117 +6,53 @@ import type { IConvertInputRulesValue } from 'capture-core-utils/rulesEngine/rul
...
-    convertCoordinate(value: any): string {
-        log.warn('convertCoordinate not implemented', value);
-        return '';
-    }
-
-    convertOrganisationUnit(value: any): string {
+        return null;
+    },
+    convertCoordinate: (value: any): ?string => (
+        (value && value.latitude && value.longitude) ? `[${value.latitude},${value.longitude}]` : null),
```
simonadomnisoru, line 43: Good job adding the missing `convertCoordinate` logic 😄 !

src/core_modules/capture-core/rules/converters/inputConverter.js
```diff
@@ -6,117 +6,53 @@ import type { IConvertInputRulesValue } from 'capture-core-utils/rulesEngine/rul
...
-
-    convertDate(value: any): string {
+export const inputConverter: IConvertInputRulesValue = {
+    convertText: (value: ?string): ?string => (value || null),
+    convertLongText: (value: ?string): ?string => (value || null),
+    convertLetter: (value: ?string): ?string => (value || null),
+    convertPhoneNumber: (value: ?string): ?string => (value || null),
+    convertEmail: (value: ?string): ?string => (value || null),
+    convertBoolean: (value: ?boolean): ?boolean => ((value || value === false) ? value : null),
+    convertTrueOnly: (value: ?boolean): ?boolean => ((value || value === false) ? value : null),
```
simonadomnisoru, line 16: Hey,
Is the comparison with `false` necessary for the `TrueOnly` converter? Maybe the logic can be simplified into this: 

`convertTrueOnly: (value: ?boolean): ?boolean => (value || null)`

WDYT?

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/defaultValues.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import { typeKeys } from '../../constants';
+
+export const defaultValues = {
+    [typeKeys.TEXT]: '',
+    [typeKeys.LONG_TEXT]: '',
+    [typeKeys.PHONE_NUMBER]: '',
+    [typeKeys.EMAIL]: '',
+    [typeKeys.BOOLEAN]: false,
+    [typeKeys.TRUE_ONLY]: false,
```
simonadomnisoru, line 10: Shouldn't the `TRUE_ONLY` default value should be `true` instead of `false`?

src/core_modules/capture-core-utils/rulesEngine/commonUtils/normalizeRuleVariable.js
```diff
@@ -3,30 +3,33 @@ import log from 'loglevel';
...
+    }
+    return numberRepresentation;
+};
 
-    const convertString = (stringRepresentation: number | string): string => stringRepresentation.toString();
+const convertBoolean = (value) => {
+    if (isString(value)) {
+        return value === 'true';
+    }
+    return value;
```
simonadomnisoru, line 21: Hey, 
Does it make sense to add to the `convertBoolean` function the use case for the `TRUE_ONLY` ruleEffectConverter which should [always return true](https://github.com/dhis2/capture-app/pull/3089/files#diff-831e3acf07b2db7e576e336385b36e9a35318c468d6824c6f680d14fccf01b76L29)?
Thanks!

src/core_modules/capture-core/rules/converters/inputConverter.js
superskip: I think you are right :)

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/defaultValues.js
superskip: "True only"-field really means checkbox, and those are considered unchecked by default, hence `false`.

src/core_modules/capture-core-utils/rulesEngine/commonUtils/normalizeRuleVariable.js
superskip: I was the one writing `[typeKeys.TRUE_ONLY]: () => true` once upon a time, but realised that this was utterly wrong while working on this ticket.

superskip:
I made a thorough comparison with Tracker Capture, and found that blanking out a field should make an associated program rule variable count as empty. Hence the input converters should transform empty strings into `null`. However, I generalised a bit too far in my last comment when I said that source value === default value implies empty variable, because a numeric source field containing a zero should not be treated as empty. In the end only the string value converters turned out to need adjustment. Or in a sense not really those either, as apparently empty strings get converted to `null` before arriving the rules engine [in this function](https://github.com/dhis2/capture-app/blob/f608019fd2e9979d4fdf0ddb3b463b4de016bef1/src/core_modules/capture-core/converters/formToClient.js#L77).

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/VariableService.js
```diff
@@ -182,6 +171,28 @@ export class VariableService {
...
+            // If a variable is mentioned in the content of the rule, but does not exist in the variables hash, show a warning:
+            log.warn(`Variable ${variableHashKey} was not defined.`);
+        } else {
+            const { variableType } = variableHash;
+            const variableValue = data === null ? this.defaultValues[variableType] : normalizeRuleVariable(data, variableType);
+
+            variablesHash[variableHashKey] = {
+                ...variableHash,
+                variableValue,
+                hasValue: data !== null,
```
JoakimSM, line 188: The `IConvertInputRulesValue` type definition also allows `undefined` to be the output from the converter. I don't think we are converting undefined to null anywhere (but this might be the solution). Can you look into what happens if `undefined` is returned from the converter and potentially how we can best deal with it?

JoakimSM:
> I made a thorough comparison with Tracker Capture, and found that blanking out a field should make an associated program rule variable count as empty. Hence the input converters should transform empty strings to `null`. However, I generalised a bit too far in my last comment when I said that source value === default value implies empty variable, because a numeric source field containing a zero should not be treated as empty. In the end only the string value converters turned out to need adjustment. Or in a sense not really those either, as apparently empty strings get converted to `null` [in this function](https://github.com/dhis2/capture-app/blob/f608019fd2e9979d4fdf0ddb3b463b4de016bef1/src/core_modules/capture-core/converters/formToClient.js#L77) before arriving the rules engine. I modified them anyway, as a safety measure.

Good conclusion!

JoakimSM (commented):
Solid work @superskip!

Couple of comments.

src/core_modules/capture-core/rules/converters/inputConverter.js
```diff
@@ -6,117 +6,56 @@ import type { IConvertInputRulesValue } from 'capture-core-utils/rulesEngine/rul
...
-    convertBoolean(value: any): boolean | string {
-        return (value || value === false) ? value : '';
-    }
-
-    convertTrueOnly(value: any): boolean | string {
-        return (value || value === false) ? value : '';
-    }
-
-    convertDate(value: any): string {
+const convertStringValue = (value: ?string): ?string => (value !== '' && value || null);
```
JoakimSM, line 9: `value !== '' && value || null` can be replaced by `value || null` if I'm not missing something

superskip: Ahaha.. :sweat_smile:  You're right

src/core_modules/capture-core-utils/rulesEngine/services/VariableService/defaultValues.js
JoakimSM: we agreed to keep `false` for now

JoakimSM (approved):

simonadomnisoru (approved):

geethaalwan (commented):

PR #3092, opened 2022-11-04
docs: [DHIS2-13654] program stage working list docs

simonadomnisoru:


docs/user/using-the-capture-app.md
```diff
@@ -789,6 +789,16 @@ You will find the predefined list views above the filters for the list. Click to
 
 ![](resources/images/tei_list_predefined_views.png)
 
+## Tracker program stage working list
+
+Visualize data elements from one single stage in the same working list. To do this, you choose “Program stage” under “More filters" and then a program stage option.
```
cooper-joe, line 794: I think we might want to avoid using the word "visualize", as we have so many apps that handle data visualization.
Suggestion:
`You can show data elements from a single stage in a working list. Select the "Program stage" option from the "More filters" dropdown, then choose a program stage.`

docs/user/using-the-capture-app.md
```diff
@@ -789,6 +789,16 @@ You will find the predefined list views above the filters for the list. Click to
...
 
+## Tracker program stage working list
+
+Visualize data elements from one single stage in the same working list. To do this, you choose “Program stage” under “More filters" and then a program stage option. 
+
+![](resources/images/program_stage_button.png)
+
+![](resources/images/program_stage_working_list.png)
+
+The list can be [filtered](#filter-the-list), [sorted](#sort-the-list) and [modified](#modify-the-list-layout) in the same way as the TEI working list.
```
cooper-joe, line 800: I'd avoid using the term "TEI working list" here, and instead use "as other working lists".

karolinelien (approved):
This looks good to me, after Joe's edits. Thanks Simona!

cooper-joe (approved):
thanks 👍

PR #3093, opened 2022-11-07
refactor: [TECH-1456] extract rules engine into its own package with yarn workspaces

simonadomnisoru:


simonadomnisoru:
> Some talking points / questions:
> 
> 1. Is there a technical reason why we have the flow-typed folder within the rules-engine package? I would prefer if we put everything in the root (npm style)
> 2. Can we use d2-app-scripts to build instead of using babel directly? Like here: https://github.com/dhis2/ui/tree/master/components/alert
> 3. We should be able to use a single `yarn.lock` file for the entire codebase. Hopefully it should be enough to add a dependency to the rules-engine in package.json in the root folder.
> 4. Folder structure: use `packages` subfolder? e.g. `packages/rules-engine`

Hey @JoakimSM,
I implemented the changes you suggested in the last commits. 
Thanks for the feedback! 😄

JoakimSM (approved):

geethaalwan (approved):

PR #3096, opened 2022-11-09
fix: [DHIS2-14081] use static style for accessory row in events table

cooper-joe:
Implements [DHIS2-14081](https://dhis2.atlassian.net/browse/DHIS2-14081).

- Uses the `DataTableCell` `staticStyle` prop on the "accessory row" in the Stages -> Events table on Enrollment dashboard to prevent hover styles. This row has no `onClick` and is not a data row, so should not have hover styles.


simonadomnisoru (approved):
LGTM!

geethaalwan (approved):

PR #3100, opened 2022-11-14
fix: [DHIS2-7913] apply hide fields program rule in form view mode

simonadomnisoru:


geethaalwan (approved):
Tested successfully on 2.37.9 version

PR #3115, opened 2022-11-21
fix: [DHIS2-14156] program stage WL event status convertor

simonadomnisoru:


src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getEventListData/convertToClientEvents.js
```diff
@@ -9,7 +11,18 @@ import type {
     TeiColumnsMetaForDataFetchingArray,
     ClientEvents,
 } from './types';
-import { getFilterClientName } from '../../../../helpers';
+import { getFilterClientName, ADDITIONAL_FILTERS } from '../../../../helpers';
+
+const isEventOverdue = (status: string, scheduledAt: string) =>
```
jasminenguyennn, line 16: hey we also have identical function [here](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/hooks/helpers.js#L11). I think we can refactor to use only 1. WDYT?

simonadomnisoru: Great suggestion! I refactor it with my last commit. Thanks!

jasminenguyennn (approved):

geethaalwan (approved):

PR #3117, opened 2022-11-21
docs: [DHIS2-13801] add doc for ask to create new event after completing

jasminenguyennn:


simonadomnisoru (approved):

karolinelien (approved):
I made a commit with some small spelling/clarification changes,didn't know it would be automatically approved, haha. But if those changes are fine then PR looks good to me :)

PR #3120, opened 2022-11-23
feat: [DHIS2-12525] enroll in current program

jasminenguyennn:


simonadomnisoru (approved):

jasminenguyennn:
there is some complexity here related to protected programs and decided that we will handle this in the enrollment page instead

PR #3122, opened 2022-11-24
fix: [DHIS2-13646] temporarily generate the updatedAt on the client

simonadomnisoru:


src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.js
```diff
@@ -27,17 +29,27 @@ export const EnrollmentAddEventPageDefault = ({
...
     const handleCancel = useCallback(() => {
         history.push(`enrollment?${buildUrlQueryString({ programId, orgUnitId, teiId, enrollmentId })}`);
     }, [history, programId, orgUnitId, teiId, enrollmentId]);
 
     const handleSave = useCallback(
         (data, uid) => {
-            dispatch(updateEnrollmentEventsWithoutId(uid, data.events[0]));
+            const clientGeneratedUpdatedAt = systemSettings?.serverTimeZoneId
+                ? moment.tz(new Date(), systemSettings.serverTimeZoneId).toISOString()
+                : new Date().toISOString();
```
JoakimSM, line 42: 1) These two lines will give the same ISO-string 2) I think we should create a string in the format the `updatedAt` has when it is retrieved from the server (without the timezone specified at the end of the string). I see that the current solution works because the timezone converter in the stageOverview is smart enough to recognise that it should not convert based on the timezone identifier if the string itself contains the timezone. 3) It would probably be best practice to wait for the `serverTimeZoneId` to be retrieved before rendering the form/show the save button.

After doing 3), we can replace these lines with: `moment().tz(systemSettings.serverTimeZoneId).format('YYYY-MM-DDTHH:mm:ss')`

(We should consider changing our entire timezone approach. More on that in a coming Tracker frontend meeting)

jasminenguyennn (approved):

simonadomnisoru:
Hey @dhis2/team-tracker, 
I changed the approach for this bug fix and took advantage of the new [useTimeZoneConversion](https://github.com/dhis2/app-runtime/pull/1308) hook introduced by the `@dhis2/app-runtime`.  The hook makes it easier to convert between server and client date and we don't have to get and use `serverTimeZoneId` directly.
Let me know what your thoughts are. Thanks!

JoakimSM:
> Hey @dhis2/team-tracker,
> I changed the approach for this bug fix and took advantage of the new [useTimeZoneConversion](https://github.com/dhis2/app-runtime/pull/1308) hook introduced by the `@dhis2/app-runtime`. The hook makes it easier to convert between server and client date and we don't have to get and use `serverTimeZoneId` directly.
> Let me know what your thoughts are. Thanks!

Neat! Good work.

JoakimSM (approved):

geethaalwan (approved):

PR #3125, opened 2022-11-28
feat: [DHIS2-13803] change label and path of view dashboard in search fallback

jasminenguyennn:


src/core_modules/capture-core/components/CardList/CardListItem.component.js
```diff
@@ -138,48 +146,73 @@ const CardListItemIndex = ({
...
-                                    programNameFromEnrollment &&
-                                    <ListEntry name={i18n.t('Program name')} value={programNameFromEnrollment} />
-                                }
-
+                                <ListEntry
+                                    name={i18n.t('Organisation unit')}
+                                    value={orgUnitName}
+                                />
+                                <ListEntry
+                                    name={i18n.t('Date of enrollment')}
```
simonadomnisoru, line 212: Does it make sense for this label to take the value of the metadata “Date of enrollment” configured from Maintenance? Thanks!

jasminenguyennn: thanks @simonadomnisoru good catch!

src/core_modules/capture-core/components/CardList/CardListItem.component.js
```diff
@@ -138,48 +146,75 @@ const CardListItemIndex = ({
...
+                    }
+                >
+                    {i18n.t('Previously enrolled')}
+                </Tag>
+            );
+        default:
+            return null;
+        }
+    };
+    console.log(program);
```
simonadomnisoru, line 187: Should remove these 2 `console` lines before merging the PR. Besides this everything looks good!

simonadomnisoru (approved):

JoakimSM:
1) We should always link to the program that is currently selected. In the image below every button under "results in all programs" should have the text: "View Child Programme dashboard"

2) A bit out of context for this PR, but great if we do it here anyway: Let's remove the "organisation unit" and "Date of enrollment" under "results found in all programs" . They are from an enrollment and doesn't make sense when looking at the TEI out of program/enrollment-context.

Skjermbilde 2023-02-01 kl  13 22 18

jasminenguyennn:
@JoakimSM thanks for pointing it out. I updated the branch accordingly

Screenshot 2023-02-03 at 09 16 41

JoakimSM (approved):

PR #3141, opened 2022-12-07
feat: [DHIS2-13237] Enrollment coordinates in enrollment widget

jasminenguyennn:
[DHIS2-13237](https://dhis2.atlassian.net/browse/DHIS2-13237)

**Tech summary**
`AddLocation.component.js`
1. toggle the `MapModal` 
2. show either `Add Coordinates` or `Add Area` based on the `geometryType`

`MiniMap.component`
1. displays the minimap
2. converts data to client format using  convertToClientCoordinates and pass it as `defaultValues` to `MapModal` 

`MapModal.container.js`
1. `onSetCoordinates` uses the `useUpdateEnrollment` hook to update the geometry 

`Coordinates.component.js`
1. displays the map for a point
2. `renderLatitude` and `renderLongitude` display the coordinates of the point under the map
4. manage the validation of the coordinates using `isValidCoordinate`
5. convert the data to API format using`convertCoordinatesToServer`

`Polygon.component.js`
1. displays the map for a polygon
2. convert the data to API format using`convertPolygonToServer`

`DeleteControl.component` 
1. refactor the logic of the [capture-ui component](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-ui/PolygonField/DeleteControl.component.js) into a functional component
2. builds and adds to the map a custom delete control button

[DHIS2-13237]: https://dhis2.atlassian.net/browse/DHIS2-13237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JoakimSM:
I see a map, that's good 🙂  Tried to register a new TEI and enroll, but the marker is not placed in the right spot. Maybe a mixup of Latitude and Longitude? Also, we should be able to update or set a new coordinate/polygon somehow, started a thread on slack for this since the design is incomplete it seems.

src/core_modules/capture-core/components/MapCoordinates/MapCoordinatesModal.js
```diff
@@ -0,0 +1,153 @@
...
+            }}
+            secondary
+        >
+            {i18n.t('Cancel')}
+        </Button>
+        <Button
+            onClick={() => {
+                if (position ?? coordinates) {
+                    const convertedCoordinates = convertToServerCoordinates(position ?? coordinates, type);
+                    // $FlowFixMe
```
simonadomnisoru, line 125: Hey, 
How about adding return types to the `convertToServerCoordinates` function and getting rid of the `// $FlowFixMe` comment? This worked when I tested it: 
```
const convertToServerCoordinates =
(coordinates?: Array<[number, number]> | null, type: string): ?[number, number] | ?Array<[number, number]> => {
    if (!coordinates) { return null; }
    switch (type) {
    case dataElementTypes.COORDINATE:
        return Array<[number, number]>([coordinates[1], coordinates[0]]);
    case dataElementTypes.POLYGON:
        return [coordinates[0]];
    default:
        return coordinates;
    }
};
```
WDYT?

src/core_modules/capture-core/components/MapCoordinates/MapCoordinatesModal.js
```diff
@@ -0,0 +1,153 @@
...
+            {`${i18n.t('Set')} ${getTitle()}`}
+        </Button>
+    </ButtonStrip>);
+
+    return (
+        <Modal
+            hide={!isOpen}
+            large
+        >
+            <ModalTitle>
```
simonadomnisoru, line 141: For enrollment without geometry, when I click the "Enrollment actions" button, an error appears in the console.
`Failed prop type: Invalid prop children of type object supplied to ModalTitle, expected string. in ModalTitle (created by MapCoordinatesModalPlain)`
Can you have a look?

Screenshot 2023-01-19 at 11 14 04

Thank you!

src/core_modules/capture-core/components/MapCoordinates/mapCoordinates.types.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+export type MiniMapProps = {
+    coordinates: any,
+    type: string,
+    classes: Object,
```
simonadomnisoru, line 6: NIT: `...CssClasses` can be used instead of `classes: Object`

src/core_modules/capture-core/components/WidgetEnrollment/Actions/AddLocation/AddLocation.component.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+import { IconLocation16, MenuItem } from '@dhis2/ui';
+import React, { useState, useMemo } from 'react';
+import i18n from '@dhis2/d2-i18n';
+import { dataElementTypes } from '../../../../metaData';
+import { MapCoordinatesModal } from '../../../../components/MapCoordinates';
+import { useProgramFromIndexedDB } from '../../../../utils/cachedDataHooks/useProgramFromIndexedDB';
```
simonadomnisoru, line 7: Hey,
In order to maintain the `WidgetEnrollment` self-contained we should avoid using anything from `IndexedDB`.To access the program's data, you can either pass the data down from the parent or call the [useProgram](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetEnrollment/hooks/useProgram.js) hook. This ensures that the widget still works in other apps that don't rely on the cache like the capture app. 
Thanks!

src/core_modules/capture-core/components/WidgetEnrollment/Actions/AddLocation/AddLocation.component.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+import { IconLocation16, MenuItem } from '@dhis2/ui';
+import React, { useState, useMemo } from 'react';
+import i18n from '@dhis2/d2-i18n';
+import { dataElementTypes } from '../../../../metaData';
+import { MapCoordinatesModal } from '../../../../components/MapCoordinates';
```
simonadomnisoru, line 6: Hey,
Does it make sense to move the `MapCoordinates` component inside the `WidgetEnrollment` folder? I suggest this in order to avoid adding global '`./../../../components/ `dependencies and try to keep it as self-contained as possible.
Thanks!

simonadomnisoru (approved):
Great job 🎉 !

geethaalwan (approved):

JoakimSM:
**Issues from a user perspective**

- ‘Add area’ shows in 'Enrollment actions' even though feature type is ‘none’

Coordinates:

- When opening the full-view mode when the coordinates have previously been set, the latitude and longtitude is not showing in the input fields. Let’s display the coordinates (but keep the input fields disabled).

- There should be a way to reset the coordinates. Suggestion: Clicking edit and then X (or wiping the coordinate fields) and clicking Set coordinates should reset the field (right now, set coordinates are sometimes disabled and other times doesn’t do anything when clicked).

- If you reopen the full-view after having edited the coordinates, you are still in edit mode. At this point you should be in view mode again.

- Is it possible to get rid of the scroll-bar?

Polygon:

- Should be able reset. Currently when you delete the polygon the set area button is disabled.

JoakimSM:
Looks better, but still some issues to fix.

1. We need validation for the input fields (coordinates). If you type in invalid values the app crashes. Let's follow the pattern from the form, with a red error message below the field. If there are errors you should not be able to save (and the pointer from the map should be removed).
2. Bug scenario: Set a coordinate/polygon from the enrollment Widget. Refresh the page and then remove the coordinate by opening the map, clicking Edit and then Set Area. The app crashes.

JoakimSM:
Thanks for looking into this @simonadomnisoru. Great if you can continue the work here and look into some problems I have found.

- The validation errors below the input fields are not visible unless you scroll down. Can we fix this somehow?
- if you have an invalid field in the modal, click cancel and then reopen the modal, the invalid data is still visible. Can we have the coordinates reset (to the previous value) when you click cancel in the modal? This also causes some weird behaviour (can't reset the coordinate)
- Can the enrollment actions dropdown close when you open the modal?

simonadomnisoru:
> Hey @JoakimSM,
> I added validation and fixed the bug with the latest commits. You can have a look?
> Thank you!

simonadomnisoru:
> Thanks for looking into this @simonadomnisoru. Great if you can continue the work here and look into some problems I have found.
> 
> * The validation errors below the input fields are not visible unless you scroll down. Can we fix this somehow?
> * if you have an invalid field in the modal, click cancel and then reopen the modal, the invalid data is still visible. Can we have the coordinates reset (to the previous value) when you click cancel in the modal? This also causes some weird behaviour (can't reset the coordinate)
> * Can the enrollment actions dropdown close when you open the modal?

Hey @JoakimSM,
I fixed the bugs with the latest commits. Let me know if you find other issues.
Thank you!

superskip (changes requested):
Hey @simonadomnisoru !

First and foremost I suggest we try to refactor `MapCoordinatesModat.component.js` into one component for points and one for polygons. Once that is done, perhaps some of the following points will have disappeared by themselves:

- defaultValues doesn't get initialised (maybe it can be removed)
- Latitude and longitude switched places in the calls to isValidCoordinate
- convertor.js -> converters.js
- MiniMap.container.js -> MiniMap.component.js
- Names on the map are not translated (perhaps this is a good thing)
- Clicking 'Set area' before clicking save cancels all changes without warning.
- Click on a MiniMap containing a polygon. Click the edit polygon button and close modal without first clicking 'Save|Cancel'. Click on the MiniMap again. The 'Save|Cancel' buttons are still displayed. The same is also true for the create and delete buttons.
- No polygon is shown when clicking on 'Edit area on map' in the edit profile modal.

superskip (commented):
Great job with the refactor @simonadomnisoru! Looks good :+1: 

I agree with your decision to disable the Cancel and Save buttons while having an editing tool selected, because it was probably the simplest solution. But now that I've tested it in practice, I found that I would have liked to exit the modal using one of the usual buttons, rather than being required to first find and click on one of the small edit tool buttons. I therefore suggest keeping both the cancel and save button enabled, but that they have slightly different behaviour depending on the state of the component. For instance while editing a polygon it would make sense that the save button saves the polygon that is currently visible on the screen. I guess we might be able to get hold of the partially edited polygon if we maintain a ref to the Map-component. And when clicking the cancel button one should ideally receive a warning whenever the polygon on the screen differs from the original polygon.

Well, the question is whether someone should first make a proper design for this or not... The basic idea seems quite intuitive though, but maybe it will be easier to cover all the cases if one takes the time to make a proper design. On the other hand it might not be very critical to get all the cases just right from the start. What do you think? One more thing to consider is that this behaviour I'm suggesting might not work all that well either (it's only based on intuition).

src/core_modules/capture-core/components/WidgetEnrollment/MapModal/Polygon/Polygon.component.js
```diff
@@ -0,0 +1,194 @@
...
+        </Map>
+    );
+
+    const onFeatureGroupReady = (reactFGref: any, featureCollection: ?FeatureCollection) => {
+        if (!reactFGref) {
+            return;
+        }
+        if (featureCollection) {
+            const leafletGeoJSON = new L.GeoJSON(featureCollection);
+            if (reactFGref) {
```
superskip, line 146: You missed a spot :eyes:

simonadomnisoru:
Hey @superskip,

I've rollback the disabled buttons logic and implemented the same design we currently have when the polygon is created/edited/deleted when enrolling a TEI. This fixes the original issue you reported:
> * Click on a MiniMap containing a polygon. Click the edit polygon button and close modal without first clicking 'Save|Cancel'. Click on the MiniMap again. The 'Save|Cancel' buttons are still displayed. The same is also true for the create and delete buttons.

IMHO I would rather merge the enrollment map with the same design that we currently have when enrolling a TEI. Afterward, if it needs improvements, we can focus on changing all the maps at the same time, not just the enrollment widget map. Let me know if you disagree or find any difference between how the polygons modal works in both places.

Thank you for the review!

superskip (commented):
The inviting blue button is definitely evil now.. I don't think this feels very releasable at the moment, even if the profile widget does the same. If we're in a hurry with releasing this, maybe we should make it look like a useless feature like in the profile widget, by not showing the polygon any more as soon as the page gets reloaded.

If we can make a common map modal component for both widgets, then I agree it would make sense to improve both simultaneously. I also imagine much can be done by improving the enrollment widget first, if one focuses on making the component reusable.

This issue wasn't yours from the start, so we might consider passing it on to me for a while if you're tired of working with it :stuck_out_tongue:

JoakimSM (commented):
@superskip gave me an overview of the problems he saw. Based on this I suggest the following:

- When opening the modal, before any changes have been made, don't display the 'Set area' button at all. We should keep the cancel button, but change the text to 'Close'.
- When starting to draw an area, display 'Close without saving' (same behaviour as current 'Cancel') and 'Set area'. 'Set area' should be disabled (also show a tooltip, 'finish drawing before saving')
- After an area has been drawn, display the buttons we have today (but replace 'Cancel' button text with 'Close without saving')
- Clicking 'cancel' in the drawing should get you back to the previous state (If no changes have been made earlier, we will only display 'Close' for example)
- Remove the option to edit the current polygon (the button that says "edit layers"). If you need to make changes, you will have to draw a new one. We need a proper design here ( https://dhis2.atlassian.net/browse/DHIS2-15970), so we will implement in different ticket.

Also, we should consolidate the map code at a later point: https://dhis2.atlassian.net/browse/TECH-1660

Let's talk if you have questions / concerns

simonadomnisoru:
> * When opening the modal, before any changes have been made, don't display the 'Set area' button at all. We should keep the cancel button, but change the text to 'Close'.
> * When starting to draw an area, display 'Close without saving' (same behaviour as current 'Cancel') and 'Set area'. 'Set area' should be disabled (also show a tooltip, 'finish drawing before saving')
> * After an area has been drawn, display the buttons we have today (but replace 'Cancel' button text with 'Close without saving')
> * Clicking 'cancel' in the drawing should get you back to the previous state (If no changes have been made earlier, we will only display 'Close' for example)
> * Remove the option to edit the current polygon (the button that says "edit layers"). If you need to make changes, you will have to draw a new one. We need a proper design here ( https://dhis2.atlassian.net/browse/DHIS2-15970), so we will implement in different ticket.
> 
> Also, we should consolidate the map code at a later point: https://dhis2.atlassian.net/browse/TECH-1660

Hey @JoakimSM,
I implemented the changes you suggested in the latest commits. Let me know if you find other issues.
Thanks!

JoakimSM:
Much better! Sometimes when I'm deleting (see video) it doesn't actually get deleted. I think the problem always occurs when: First editing an existing area and saving (set area), then clicking the minimap again and deleting (no browser refresh in between).

https://github.com/dhis2/capture-app/assets/12745737/8f64b2b9-1b6f-4cd3-adb1-65fa52bca5ce

This is hopefully the last hurdle.

simonadomnisoru:
> Much better! Sometimes when I'm deleting (see video), it doesn't actually get deleted. I think the problem always occurs when: First editing an existing area and saving (set area), then clicking the minimap again and deleting (no browser refresh in between).
> 
>  Skjermopptak.2023-11-10.kl.15.24.49.mov 
> This is hopefully the last hurdle.

Hey @JoakimSM,
I fixed the delete bug with the latest commit. Let me know if you find other issues.
Thank you!

JoakimSM (approved):
🎉

geethaalwan (approved):
Tested successfully on 2.41,2.40.3,2.39.4,2.38.6 versions

superskip (approved):

PR #3142, opened 2022-12-07
fix: [DHIS2-14340][DHIS2-13197] Sort blank cells in stages & event widget

eirikhaugstulen:
Summary of issue:

- Text values are not sorted - should use the localeCompare to sort asc/desc
- Null values should be treated as bigger that non-null values


jasminenguyennn (approved):
👍

karolinelien:
@JoakimSM Yes, I was actually mentioning the same to @eirikhaugstulen but wanted to have him double-check with you, and at that moment you wrote this comment :) 

Since creating this issue initially, I have had the same convo with Analytics on sorting in the Line Listing app, and they said the same as you, that in the database sorting follows this logic: 
_The NULLS FIRST and NULLS LAST options can be used to determine whether nulls appear before or after non-null values in the sort ordering. By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise._

But it would still be good to fix that blank values are taking into account when sorting.

JoakimSM:
> 

Thanks @karolinelien. Yes, makes sense to follow the default described there: "null values sort as if larger than any non-null value"

JoakimSM (approved):

geethaalwan (approved):
Tested successfully on 2.40,2.39.1,2.38.3  versions

PR #3150, opened 2022-12-13
feat: [TECH-1486] Utility hook for fetching indexed db data

eirikhaugstulen:


src/core_modules/capture-core/utils/reactQueryHelpers/query/useMetadataQuery.types.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import type { UseQueryResult } from 'react-query';
+
+export type CustomOptions = $ReadOnly<{|
+    useCache?: boolean,
+    enabled?: boolean,
+|}>;
+
+export type Result<TResultData> = UseQueryResult<TResultData>;
```
eirikhaugstulen, line 9: Instead of returning just `data`, `loading` and `error`, I changed it to return all `useQuery`-attributes. Open to discussion on this one.

src/core_modules/capture-core/utils/cachedDataHooks/useProgramFromIndexedDB.js
```diff
@@ -1,30 +1,20 @@
 // @flow
```
eirikhaugstulen, line 1: This would not be part of the final PR - just a way to show how these hooks would be used.

src/core_modules/capture-core/utils/reactQueryHelpers/query/useMetadataQuery.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const useMetadataCustomQuery = <TResultData>(
+    queryKey: QueryKey,
+    queryFn: QueryFunction<TResultData>, {
+        cacheTime = 0,
+        enabled = true,
+    }: CustomOptions = {}): Result<TResultData> =>
+        useAsyncMetadata(queryKey, queryFn, { cacheTime, enabled });
+
+export const useIndexedDBQuery = <TResultData>(
```
eirikhaugstulen, line 23: In this example, what is the difference between `useIndexedDBQuery` & `useMetadataCustomQuery`? What is the expected behavior here?

eirikhaugstulen (commented):
Hey @JoakimSM,

Not really sure what to do on this one, as I think your setup work as expected. Left some comments down below that you can check out.

eirikhaugstulen:
The failed tests are because of the `useProgramFromIndexedDB`-hook and that I have not changed the return on multiple files. Will fix/remove before merge.

JoakimSM (approved):
🎆

PR #3157, opened 2022-12-30
feat: [DHIS2-5437] change wording for discard modal

jasminenguyennn:


src/core_modules/capture-core/components/DataEntry/withBrowserBackWarning.js
```diff
@@ -68,11 +68,11 @@ const getEventListener = (InnerComponent: React.ComponentType<any>) =>
                         {...passOnProps}
                     />
                     <ConfirmDialog
-                        header={i18n.t('Unsaved changes')}
-                        text={i18n.t('Leaving this page will discard the changes you made to this event.')}
-                        confirmText={i18n.t('Yes, discard')}
+                        header={i18n.t('Discard unsaved changes?')}
```
simonadomnisoru, line 71: Hey,
Can we reuse the `header`, `text`, `destructiveText` and `cancelText` const values from [ConfirmDialog.constants.js](https://github.com/dhis2/capture-app/blob/2a9930553b4567128db84a43fe9104132e16bbc0/src/core_modules/capture-core/components/Dialogs/ConfirmDialog.constants.js)?
Thanks!

src/core_modules/capture-core/components/Dialogs/ConfirmDialog.component.js
```diff
@@ -3,7 +3,9 @@ import React from 'react';
 import { Modal, ModalTitle, ModalContent, ModalActions, ButtonStrip, Button } from '@dhis2/ui';
 import type { Props } from './confirmDialog.types';
 
-export const ConfirmDialog = ({ open, header, text, confirmText, onConfirm, cancelText, onCancel }: Props) => (
+export const ConfirmDialog = ({
```
simonadomnisoru, line 6: Hey,
IMHO now that the destructive functionality has been added, the name `ConfirmDialog` is no longer appropriate. Can we change it to something more descriptive? Maybe `DiscardChangesDialog` is a good choice.
Thanks!

src/core_modules/capture-core/components/TopBarActions/TopBarActions.container.js
```diff
@@ -16,9 +16,9 @@ const defaultContext = {
 };
 
 const defaultDialogProps = {
-    header: i18n.t('Unsaved changes'),
-    text: i18n.t('Leaving this page will discard the changes you made to this event.'),
-    confirmText: i18n.t('Yes, discard'),
+    header: i18n.t('Discard unsaved changes?'),
```
simonadomnisoru, line 19: The same suggestion here about reusing `header`, `text`, `destructiveText` and `cancelText` const values.

src/core_modules/capture-core/components/WidgetEnrollmentEventNew/AccessVerification/AccessVerification.container.js
```diff
@@ -12,9 +12,9 @@ import { makeEventAccessSelector } from './accessVerification.selectors';
 import type { ContainerProps } from './accessVerification.types';
 
 const dialogConfig = {
-    header: i18n.t('Unsaved changes'),
-    text: i18n.t('Leaving this page will discard the changes you made to this event.'),
-    confirmText: i18n.t('Yes, discard'),
+    header: i18n.t('Discard unsaved changes?'),
```
simonadomnisoru, line 15: The same suggestion here about reusing `header`, `text`, `destructiveText` and `cancelText` const values.

src/core_modules/capture-core/components/Dialogs/DiscardDialog.component.js
```diff
@@ -16,9 +18,12 @@ export const ConfirmDialog = ({ open, header, text, confirmText, onConfirm, canc
...
                     {cancelText}
                 </Button>
-                <Button onClick={onConfirm} primary>
+                {confirmText ? <Button onClick={onConfirm} primary>
                     {confirmText}
-                </Button>
+                </Button> : null }
+                {destructiveText ? <Button onClick={onDestroy} destructive>
+                    {destructiveText}
+                </Button> : null}
```
eirikhaugstulen, line 26: Is the idea here to incorporate both a `ConfirmDialog` and `DiscardDialog` in the same component? 
In that case, my vote would be to make this a general `Dialog` and pass in a `type`-prop. If the component would strictly be used as a destructive modal, I guess `destructiveText` and `onDestroy` should be required props.

src/core_modules/capture-core/components/Dialogs/DiscardDialog.constants.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import i18n from '@dhis2/d2-i18n';
+
+export const defaultDialogProps = {
+    header: i18n.t('Discard unsaved changes?'),
+    text: i18n.t('This event has unsaved changes. Leaving this page without saving will lose these changes. Are you sure you want to discard unsaved changes?'),
+    destructiveText: i18n.t('Yes, discard changes'),
```
eirikhaugstulen, line 7: Same as mentioned in `DiscardDialog.component.js`:
If this dialog is meant to be used as both `Confirm` and `Discard`, then this default prop would still be passed in even if `confirmText` is set. The result would be having two buttons. (📸)

image

src/core_modules/capture-core/components/ScopeSelector/QuickSelector/SingleLockedSelect.component.js
```diff
@@ -140,8 +140,8 @@ const SingleLockedSelectPlain =
                       </Paper>
                   </div>
           }
-          <ConfirmDialog
-              onConfirm={handleConfirm}
+          <DiscardDialog
+              onDestroy={handleConfirm}
```
eirikhaugstulen, line 144: NIT: Should `handleConfirm` also be renamed to `handleDiscard`?

eirikhaugstulen (commented):
Hey @jasminenguyennn,
Some small things that we could probably discuss. Other than that, the refactor looks fine!
Thanks!

src/core_modules/capture-core/components/Dialogs/DiscardDialog.constants.js
jasminenguyennn: hey @eirikhaugstulen, IMO we can leave it generic but it all depends on the design and use case.
I think of a case when we want to have multiple actions 1. No, stay here 2. Commit changes and navigate 3. Discard changes and navigate.
I think it's generally acceptable to have both primary and destructive button in the same scope. 
One of the example is in the app we have the action to Save/Cancel/Delete  
Screenshot 2023-01-05 at 11 18 25

We can discuss further in the team meeting if you want

eirikhaugstulen: No, I definitely see your point in having three actions when you think of it like that 😅 The way I had pictured this widget was that you were only meant to ask the user if they really want to do the action they have clicked - not asking them which action to chose. But I'm not really sure of the use case here, so having three actions might be suitable.

@simonadomnisoru @JoakimSM any opinions?

simonadomnisoru: Hey, 

As of today, I'm not aware of any use case where both actions are needed in the same scope. Therefore, this premature optimization may never be necessary.
On the other hand, this pattern is already used in all forms and is familiar to the users. So if the use case occurs in the future, the component is ready to use.
I don't really have a strong opinion here. I see a good argument for both approaches and whichever path you choose is fine with me.

Thank you!

simonadomnisoru (approved):

src/core_modules/capture-core/components/Dialogs/DiscardDialog.constants.js
karolinelien: If I am understanding it right, I also think just two actions are necessary. 1. Yes, discard changes (and move to wherever user clicked) and 2. No, cancel.

Adding the action "Yes, but save changes" will add too much complexity I think and the user can just click cancel and save in the traditional way.

simonadomnisoru: Hey @eirikhaugstulen,
I updated the code based on @karolinelien answer. The `Discard` button is the only one available and the modal props `destructiveText` and `onDestroy` have been changed to required. Let me know if you find other issues.
Thank you for the feedback! 😄

eirikhaugstulen (approved):
LGTM, @simonadomnisoru 👍

geethaalwan (approved):
Tested successfully on 2.41,2.40.1,2.39.3,2.38.5 versions

PR #3159, opened 2023-01-03
fix: [TECH-1485] Retrieve search groups on demand

eirikhaugstulen:
Retrieve the search groups from indexedDB instead of fetching from the `MemoryObjects`


src/core_modules/capture-core-utils/storage/StorageController.js
```diff
@@ -4,6 +4,13 @@ import isArray from 'd2-utilizr/lib/isArray';
 import log from 'loglevel';
 import { errorCreator } from '../errorCreator';
 
+export class IndexedDBError extends Error {
```
simonadomnisoru, line 7: NIT: move this class to it one `IndexedDBError` file

src/core_modules/capture-core/components/Pages/Search/SearchPage.component.js
```diff
@@ -172,7 +170,10 @@ const Index = ({
     const searchStatusComponents = () => (
         <>
             {searchStatus === searchPageStatus.SHOW_RESULTS && (
-                <SearchResults availableSearchOptions={availableSearchOptions} fallbackTriggered={fallbackTriggered} />
+                <SearchResults
+                    availableSearchOptions={availableSearchOption}
```
simonadomnisoru, line 174: Hey,
Does it make sense to rename `availableSearchOptions` to `availableSearchOption` in order to maintain the consistency of the variable name across different components? 
Thanks!

src/core_modules/capture-core/components/Pages/Search/SearchResults/SearchResults.types.js
```diff
@@ -25,7 +25,7 @@ export type PropsFromRedux ={|
 
 export type OwnProps ={|
     fallbackTriggered: boolean,
-    availableSearchOptions: AvailableSearchOptions,
+    availableSearchOptions: ?AvailableSearchOption,
```
simonadomnisoru, line 28: Same suggestion here about removing `s` from the end of the variable, just like in [SearchPage.types.js ](https://github.com/dhis2/capture-app/pull/3159/files#diff-9dc06406f1668010a7ea187da988e2d9c7e20caa892274965decb1ea72a36831R27)

src/core_modules/capture-core/components/Pages/Search/SearchPage.types.js
```diff
@@ -11,21 +11,20 @@ export type SearchGroups = Array<{|
...
+|}
 
 export type ContainerProps = $ReadOnly<{|
   cleanSearchRelatedInfo: ()=>void,
   navigateToMainPage: ()=>void,
   showInitialSearchPage: ()=>void,
   navigateToRegisterUser: ()=>void,
   trackedEntityTypeId: string,
-  availableSearchOptions: AvailableSearchOptions,
+  availableSearchOption: ?AvailableSearchOption,
```
simonadomnisoru, line 27: Hey,
The user is currently able to search from the main page for tracker programs with 'Display front page list' set to false. Meaning that the `availableSearchOption` variable shouldn't be optional and that the same changes you did in `SearchPage.container` should be applied to the [MainPage.container](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js#L131). 

Thanks!

eirikhaugstulen: Oh, that's a good catch! Changed so we also use the hook in the mainpage. I also agree that `availableSearchOption` should be required, but how do we feel like we should handle the async-nature of this hook? I'm guessing we're only getting more async hooks if we are taking data directly from IndexedDB. Should we try to incorporate all loading states with the `ready`-prop? If so, Flow might still argue for it being undefined. Another way would be to allow undefined values, but return something appropriate in the component. WDYT?

simonadomnisoru: I'm more in favor of allowing undefined values and letting the component handle them instead of using `withLoadingIndicator` HOC and `ready` prop. Depending on the scenarios, the `Example.component.js` can make use of the different helpers from [LoadingMask](https://github.com/dhis2/capture-app/tree/master/src/core_modules/capture-core/components/LoadingMasks) and decide what is shown to the user.

JoakimSM: > Should we try to incorporate all loading states with the ready-prop?
>> I'm more in favor of allowing undefined values and letting the component handle them instead of using withLoadingIndicator HOC and ready prop. Depending on the scenarios, the Example.component.js can make use of the different helpers from LoadingMask and decide what is shown to the user.

In general, I think it depends (and there is no one size fits all solution). If we're only waiting for something from IndexedDB it probably makes sense to not show a loading spinner at all (and this is maybe best dealt with in the component). Other times, it might make sense to incorporate it into to the `ready` state (simplifying the component).

eirikhaugstulen: Hey again,
After a discussion with Joakim we've ending up commenting out the flow issue here. We'll see if we can look into the HOC only returning the correct props using generics in the future, but not critical as of now.

eirikhaugstulen:
Thanks for the review comments @JoakimSM & @simonadomnisoru - should be better now 🙌

geethaalwan (approved):

eirikhaugstulen:
@JoakimSM @simonadomnisoru - Implemented the changes we discussed in the FE check in on Tuesday. Great if you could have another quick look.

simonadomnisoru (approved):

JoakimSM (approved):
Good work!

geethaalwan (approved):

PR #3160, opened 2023-01-04
fix: [DHIS2-14340][DHIS2-13197] Sort blank cells in stages&event-widget

eirikhaugstulen:


simonadomnisoru (approved):

geethaalwan (approved):
Tested on 2.37.10 version by Eirik.

PR #3164, opened 2023-01-11
feat: [DHIS-14336] enter press should trigger search

jasminenguyennn:


simonadomnisoru (commented):
Hey, 
Does it make sense to add a cypress test to this new feature? I believe [cy.type('{enter}')](https://docs.cypress.io/api/commands/type#Arguments) can be used to simulate the `Enter`. 
Other than this suggestion, everything else looks great! 👏 
Thanks!

simonadomnisoru (approved):
Good job!

geethaalwan (approved):

PR #3165, opened 2023-01-12
feat: [DHIS2-14008] use the SelectionBar ui component inside ScopeSelector

simonadomnisoru:

PR #3169, opened 2023-01-16
fix: top selector bar design changes

cooper-joe:
This PR makes design changes to https://github.com/dhis2/capture-app/pull/3165.


simonadomnisoru (approved):
LGTM!

PR #3178, opened 2023-01-30
fix: [DHIS2-14578] fix unique value re enrollment

jasminenguyennn:


src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/hooks/useFormValues.js
```diff
@@ -113,18 +114,23 @@ const buildFormValues = async ({
     formValuesReadyRef: { current: boolean },
     searchTerms?: ?Array<{[key: string]: any}>,
     querySingleResource: QuerySingleResource,
+    removeUniqueValue: boolean
 }) => {
+    const clientAttributesWithoutUniqueValue = removeUniqueValue ?
+        clientAttributesWithSubvalues?.filter(value => !value.unique)
```
simonadomnisoru, line 120: Hey, 
IMHO removing unique data elements from the form values is not the best approach. We should instead make sure that the [logic to detect other TEIs with the same unique ids](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/factory/enrollment/DataElementFactory.js#L103) works as expected. 
From what I tested, our current checks work as intended. I mention it in the Jira ticket and request a retest from @karolinelien. 
Thanks!

simonadomnisoru (approved):
Thanks for looking into this! 😄

geethaalwan (approved):
Tested successfully on 2.40,2.39.2,2.38.4 versions

PR #3184, opened 2023-02-03
fix: [TECH-1495] cypress tests with dynamic dates

superskip:
Backport of https://github.com/dhis2/capture-app/pull/3162

### DIFF
--- a/cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageNavigation.feature
+++ b/cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageNavigation.feature
@@ -9,14 +9,14 @@ Feature: User interacts with Enrollment event page
     Given you open the enrollment page which has multiples events and stages
     And you see the following Enrollment Dashboard
     And the program stages should be displayed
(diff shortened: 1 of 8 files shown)


cypress/integration/EnrollmentEditEventPage/EnrollmentEditEventPageNavigation.feature
```diff
@@ -9,14 +9,14 @@ Feature: User interacts with Enrollment event page
     Given you open the enrollment page which has multiples events and stages
     And you see the following Enrollment Dashboard
     And the program stages should be displayed
-    When the user clicks the event with the report date 2021-07-13
+    When the user clicks the event with the report date 2022-07-13
```
jasminenguyennn, line 12: hey I don't think we should rely on the hardcoded report date, it may change next year. 
the new version [here](https://github.com/dhis2/capture-app/pull/3005/files#diff-21e4d93104a5e6853ab6d318c62ff4d8942c0eab746796f9e21ed7c65d1df3a6R8) doesn't use the report date anymore

simonadomnisoru (approved):

PR #3188, opened 2023-02-09
feat: [DHIS2-14405] Schedule referral event in another orgUnit

eirikhaugstulen:
Some discussion points:

- [x] The actions ending with `WITHOUT_ID` probably does not make sense anymore. We could rewrite them to use the same as the EditEvent-actions, but this would also require a rewrite of the editEvent-page. Then we have to enforce that we always pass in an array of events. (Might be beneficial if we were to have multiple edit modes in the future)

- [x] Even though we are not using the referrals for anything yet, should we implement a way to save the relationship in redux? The same way that we're having optimistic UI for events.

- [x] How do we show an error message in the referral-fields. We need to show a generic error if there is no referral action selected, as well as showing an error on the individual fields. These are not part of the data entry-form and might need a rewrite because of this.

- [x] As of now, this validation is done in the handleSave-function. This is probably not the best place to do it. Open for input on this.


src/core_modules/capture-core/components/WidgetReferral/ReferToOrgUnit/ReferToOrgUnit.container.js
```diff
@@ -0,0 +1,106 @@
...
+            ...prevValues,
+            orgUnit: null,
+        }));
+    };
+
+    return (
+        <div className={classes.wrapper}>
+            <div className={classes.fieldWrapper}>
+                <div className={classes.fieldLabel}>
+                    {i18n.t('Schedule date / Due date', { interpolation: { escapeValue: false } })}
```
simonadomnisoru, line 74: Should use the description of the due date set in Maintenance. In other forms, it is used like this `formFoundation.getLabel('scheduledAt')`. Thanks!

src/core_modules/capture-core/utils/uid/generateUID.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+export const generateUID = (): string => {
```
simonadomnisoru, line 3: Any reason you make our own Id generator and don't use the one from the `uuid` library?

src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/getConvertedReferralEvent.js
```diff
@@ -0,0 +1,67 @@
+import { systemSettingsStore } from '../../../metaDataMemoryStores';
```
simonadomnisoru, line 1: This file is missing // @flow and types for `getConvertedReferralEvent`

simonadomnisoru (commented):
Hey @eirikhaugstulen,
Great progress on this feature! I've added some comments, and below you can find some thoughts on the talking points you mentioned:
1) I am not aware of any plans to have multiple edit modes in the future. But I agree that the `WITHOUT_ID` does not make sense anymore. If the rewrite of the event edit page is complex, perhaps you can just rename the related reducers and actions by replacing `WITHOUT_ID` and `WithoutId` with something else you see fit.
2) I propose that we store them when we use it in the next Jira issues.
3) and 4) Any reason not to use a data entry form for the referral form? You can reuse the existing error handling and required field logic.

Thank you!

src/core_modules/capture-core/utils/uid/generateUID.js
eirikhaugstulen: This is to get an Id that looks like the others in the dhis2-system. It's the function that was used in TC. I guess it's only cosmetic, but not sure how the system would react when using 32 characters (uuid) vs 8 characters (DHIS2). Do you prefer to use the uuid-library?

src/core_modules/capture-core/components/WidgetReferral/ReferralActions/ReferralActions.component.js
```diff
@@ -34,29 +40,47 @@ const styles = () => ({
...
-        <div className={classes.wrapper}>
-            {type === referralStatus.REFERRABLE ? Object.keys(mainOptionTranslatedTexts).map(key => (
-                <Radio
-                    key={key}
-                    name={`referral-action-${key}`}
-                    checked={key === selectedAction}
-                    label={mainOptionTranslatedTexts[key]}
-                    onChange={(e: Object) => setSelectedAction(e.value)}
-                    value={key}
+    const { programStage } = useProgramStageInfo(selectedType.toConstraint.programStage.id);
```
simonadomnisoru, line 50: Hey,
I get an error for a relationship without `toConstraint.programStage`. Maybe you can add a `?` to fix it. Something like this: `useProgramStageInfo(selectedType.toConstraint.programStage?.id);`
Thanks!

src/core_modules/capture-core/utils/uid/generateUID.js
simonadomnisoru: I was not aware that this is the shape of id used by the old tracker and across the dhis2-system. Better leave it as it is then. Thanks for the explanation! 😄

src/core_modules/capture-core/components/WidgetReferral/ReferToOrgUnit/ReferToOrgUnit.container.js
eirikhaugstulen: Yes, good point!
As we don't have a formFoundation for the constraining programStage, I built a hook to fetch it from the api. Could get it through IndexedDB, but it would require fetching a lot more data than we need. Open for suggestions.

src/core_modules/capture-core/components/WidgetReferral/ReferralActions/ReferralActions.component.js
eirikhaugstulen: Good catch! This was never meant to be hardcoded to the `toConstraint`-object either way, but the other constraint regardless of `from` or `to`. Should be fixed now.

eirikhaugstulen:
Hey,
So since our chat last week I did some changes.
- I tried to reuse most of the DataEntry-form components, but since some are coupled with redux I had to create my own. (ref. DateFieldForReferral).
- Added some handling for showing and removing error messages related to fields
- Passed in a ref that exposes functions for checking validity and extracting values

src/core_modules/capture-core/components/WidgetReferral/ReferToOrgUnit/ReferToOrgUnit.container.js
simonadomnisoru: If the Widget is not self-contained, we tend to avoid extra fetches and use the metadata cached in IndexDB instead. IMHO even if more data is fetched once, this is preferable then fetching the scheduledAt metdata label each time `WidgetReferral` is mounted. WDYT?

simonadomnisoru (commented):
Hey,

Great job with the `DateFieldForReferral` and `OrgUnitSelectorForReferral`. I only have some small suggestions.

Firstly, there are a couple of places where `Object` is used and perhaps a more descriptive type shape could be used instead. I suggested ideas for most places. Can you take a look?

Secondly, what is the plan to test this feature? I couldn't find any relationship with `referral` set to true in the DB. Do you plan to change the DB and add some Cypress tests later on?

Thank you!

src/core_modules/capture-core/components/WidgetReferral/ReferToOrgUnit/ReferToOrgUnit.container.js
JoakimSM: @eirikhaugstulen and I discussed this a bit earlier @simonadomnisoru. I agree, we prefer getting metadata from IndexedDB when we are outside of a self-contained Widget. 

At the same time I wasn't sure we should redo it now since this would be the way to do it if we make it self-contained (In self-contained mode we would probably be able to pass in the metadata, but if it is not passed in we would grab it from the api in the Widget itself).

But yeah, ideally we would grab it from IndexedDB like you said. We should also extend the IndexedDB interface with a callback method like we have on getAll to limit the data the calling method gets back (called `project` for getAll).

Even though we agreed earlier to leave it for now @eirikhaugstulen, maybe it's good to set precedence here and change it to IndexedDB after all (the `project` method you can decide if you want to implement here, or if we do it later)

eirikhaugstulen:
Hey @simonadomnisoru - thanks for the great contribution! I was meant to clean it up a bit, but it must've went under my radar. 

When it comes to the fetching of metadata, this is an ongoing discussion now that we're sort of in the middle of transition to self-contained widgets. The reasoning here is that this referral widget might someday be a self contained widget, therefore not using `<DataEntry />` and `indexedDB`. The overall idea is that in our case, we could fetch metadata from indexedDB and feed it into the widget. In the event that the metadata is not provided, say when an external app or developer is using it, we would still need to fetch data from the API. Therefore, we still need to build the api-handling, but disabling it if we pass in metadata from the outside. We could also fetch the metadata and pass it in now, but we decided that it was not necessary just yet. We could discuss this in the meeting tomorrow if you want.
(Btw, we're using a pretty strict caching strategy on the api-call so the fetch should only happen once per app initialization)

JoakimSM (commented):
Thanks @eirikhaugstulen, I've made some comments in code. Additionally, I would like us to show error messages after you have touched a field (the same pattern as for the other forms). 

But certainly looks good overall @eirikhaugstulen!

src/core_modules/capture-core/components/WidgetReferral/useReferral.js
```diff
@@ -1,48 +1,50 @@
...
-        } else if (selectedRelationshipTypes.length > 1) {
-            setReferralStatus(referralStatus.AMBIGUOUS_REFERRALS);
         }
     }, [
-        setSelectedRelationshipTypes,
-        selectedRelationshipTypes,
+        relationshipTypes,
         setReferralStatus,
         programStageId,
     ]);
```
JoakimSM, line 43: I know you didn't write this from scratch, but can we refactor this to use useMemo (ie, get rid of useEffect and the useState)? I don't see any side-effects here.

eirikhaugstulen:
New updates:
- Updated the errorHandling and formValidation to also show error when field is touched.
- Tried to decouple the code to separate `WidgetReferral` and `WidgetEnrollmentEventNew
- Moved the creation of relationship and referralEvent to `Validated.container.js`
- Added `project`-filtering to the get method in `IndexedDbAdapter`.
    - Did not see any value in converting it from a static method to regular (same as `getAll()`). Let me know if you disagree. 
- Some general refactoring

src/core_modules/capture-core/components/WidgetReferral/WidgetReferral.component.js
```diff
@@ -1,11 +1,100 @@
...
+        const { referralEvent, relationship } = getConvertedReferralEvent({
+            referralDataValues,
+            currentProgramStageId: programStageId,
+            currentEventId: eventId,
+            programId,
+            teiId,
+            enrollmentId,
+            // $FlowFixMe - selectedRelationshipType is not null
+            referralType: selectedRelationshipType,
+        });
```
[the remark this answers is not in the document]
eirikhaugstulen: Thanks for the feedback, I think I've loosened the code a bit in the newest commit.

src/core_modules/capture-core/components/WidgetReferral/FormComponents/OrgUnitSelectorForReferral.js
```diff
@@ -0,0 +1,77 @@
...
+    const handleDeselect = () => {
+        setTouched(true);
+        onDeselectOrgUnit();
+    };
+
+    const shouldShowError = (saveAttempted || touched);
+
+    return (
+        <OrgUnitFieldForForm
+            label={i18n.t('Organisation unit', { interpolation: { escapeValue: false } })}
```
simonadomnisoru, line 68: NIT: `'Organisation unit'` is plain text, `interpolation` is only needed using a `{{variable}}`

simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetReferral/FormComponents/DateFieldForReferral.js
```diff
@@ -0,0 +1,66 @@
...
+import React, { useState } from 'react';
+import { convertStringToDateFormat } from '../../../utils/converters/date';
+import {
+    DateField,
+    withDefaultFieldContainer,
+    withDisplayMessages,
+    withInternalChangeHandler,
+    withLabel,
+} from '../../FormFields/New';
+import labelTypeClasses from '../../WidgetEnrollmentEventNew/DataEntry/dataEntryFieldLabels.module.css';
```
JoakimSM, line 11: Can we duplicate what we need from here and put it inside the WidgetReferral folder hierarchy?

src/core_modules/capture-core/components/WidgetReferral/FormComponents/OrgUnitSelectorForReferral.js
```diff
@@ -0,0 +1,77 @@
...
+import React, { useState } from 'react';
+import i18n from '@dhis2/d2-i18n';
+import {
+    SingleOrgUnitSelectField,
+    withDefaultFieldContainer,
+    withDisplayMessages,
+    withInternalChangeHandler,
+    withLabel,
+} from '../../FormFields/New';
+import labelTypeClasses from '../../WidgetEnrollmentEventNew/DataEntry/dataEntryFieldLabels.module.css';
```
JoakimSM, line 11: Same as the the other comment.

JoakimSM (approved):
Looks great @eirikhaugstulen! (couple of minor changes (see comments) and we are good to go for testing)

PR #3198, opened 2023-02-27
feat: [DHIS2-9661] first stage registration

jasminenguyennn:


src/core_modules/capture-core/hooks/useTrackedEntityTypesWithCorrelatedPrograms.js
```diff
@@ -32,9 +33,28 @@ export const useTrackedEntityTypesWithCorrelatedPrograms = (): TrackedEntityType
...
+                useFirstStageDuringRegistration,
             }: TrackerProgram) => {
                 const accumulatedProgramsOfTrackedEntityType =
                   acc[trackedEntityTypeId] ? acc[trackedEntityTypeId].programs : [];
+                const programMetadata = { programId, programName, searchGroups, enrollment };
+                if (useFirstStageDuringRegistration) {
+                    const accessibleStages = [...stages.values()].filter(({ access }) => access.write);
+                    if (accessibleStages.length) {
+                        const firstAccessibleStage = accessibleStages[0];
+                        firstAccessibleStage.stageForm.getSection(Section.MAIN_SECTION_ID).name
```
eirikhaugstulen, line 46: This breaks if the data entry form of the first stage is a section form. We should probably still allow different form types. IMO we should also not overwrite the section header if this is custom. (But I agree that the design and UI seems like this is intended)

src/core_modules/capture-core/hooks/useTrackedEntityTypesWithCorrelatedPrograms.js
```diff
@@ -32,9 +33,28 @@ export const useTrackedEntityTypesWithCorrelatedPrograms = (): TrackedEntityType
...
+                stages,
+                useFirstStageDuringRegistration,
             }: TrackerProgram) => {
                 const accumulatedProgramsOfTrackedEntityType =
                   acc[trackedEntityTypeId] ? acc[trackedEntityTypeId].programs : [];
+                const programMetadata = { programId, programName, searchGroups, enrollment };
+                if (useFirstStageDuringRegistration) {
+                    const accessibleStages = [...stages.values()].filter(({ access }) => access.write);
+                    if (accessibleStages.length) {
+                        const firstAccessibleStage = accessibleStages[0];
```
eirikhaugstulen, line 45: Even if it seems that the order of our stages is correct, this order is really important to show the correct stage on the registration page. Perhaps we should rewrite this to use the `sortOrder`-flag on the stage itself?

src/core_modules/capture-core/hooks/useTrackedEntityTypesWithCorrelatedPrograms.js
```diff
@@ -32,9 +33,28 @@ export const useTrackedEntityTypesWithCorrelatedPrograms = (): TrackedEntityType
                 },
                 searchGroups,
                 enrollment,
+                stages,
+                useFirstStageDuringRegistration,
```
eirikhaugstulen, line 37: I know that this is the default DHIS2 terminology, but in this case, it really seamed like you were trying to pass a hook as a prop because of the `use`-naming. How does the team feel about this? Could we consider renaming this internally?

src/core_modules/capture-core/hooks/useTrackedEntityTypesWithCorrelatedPrograms.js
```diff
@@ -1,7 +1,8 @@
 // @flow
```
eirikhaugstulen, line 1: We're currently pushing to remove these programCollection objects as part of the plugin-epic. The registration page is one of the places we have rewritten in https://github.com/dhis2/capture-app/pull/3200 and we should probably align these PR's.

src/core_modules/capture-core/metaData/Program/ProgramStage.js
```diff
@@ -20,6 +20,8 @@ export class ProgramStage {
     _openAfterEnrollment: boolean;
     _allowGenerateNextVisit: boolean;
     _generatedByEnrollmentDate: boolean;
+    _access: Object;
```
eirikhaugstulen, line 23: Could we be more specific here?

src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.epics.js
```diff
@@ -148,15 +174,39 @@ export const startSavingNewTrackedEntityInstanceWithEnrollmentEpic: Epic = (
...
-            const values = formsValues['newPageDataEntryId-newEnrollment'] || {};
+            const { occurredAt, enrolledAt, geometry, stageComplete, stageOccurredAt, stageGeometry } =
+                dataEntriesFieldsValue[formId] || {};
+            const { trackedEntityType, stages, useFirstStageDuringRegistration } = getTrackerProgramThrowIfNotFound(programId);
+
+            const values = formsValues[formId] || {};
+            const stageValues = {};
+            if (useFirstStageDuringRegistration) {
+                const firstStage = [...stages][0][1];
+                const stageDataElementIds = [...firstStage.stageForm.sections].reduce((acc, section) => {
```
eirikhaugstulen, line 188: The RenderFoundation class has a method called `getElements()` that returns an array of all DataElements. This is very important for the plugin epic as the section.elements might not only contain data elements anymore. Perhaps we could create a method in the RenderFoundation class called `getElementIds()`?

src/core_modules/capture-core/components/DataEntries/EnrollmentRegistrationEntry/EnrollmentRegistrationEntry.types.js
```diff
@@ -21,6 +21,7 @@ export type OwnProps = $ReadOnly<{|
   teiId?: ?string,
   skipDuplicateCheck?: ?boolean,
   trackedEntityInstanceAttributes?: Array<InputAttribute>,
+  firstStageForm?: Object,
```
eirikhaugstulen, line 24: Could we be more specific here?

src/core_modules/capture-core/components/DataEntries/common/useRegistrationFormInfoForSelectedScope.js
```diff
@@ -27,13 +29,14 @@ const useRegistrationOptions = (): RegistrationOptions => {
                     name: trackedEntityTypeName,
                     formId: `registrationPageForm-${trackedEntityTypeId}`,
                 },
-                ...programs.reduce((accumulated, { programId, programName, enrollment }) => ({
+                ...programs.reduce((accumulated, { programId, programName, enrollment, firstStageForm }) => ({
```
eirikhaugstulen, line 32: Same as the comment on `useTrackedEntityTypesWithCorrelatedPrograms`. We're trying to move away from these memoryObjects, and should therefore try to align these PR's.

eirikhaugstulen:
Hey @jasminenguyennn,
Good job on this! Some comments about aligning some PRs that we could talk a bit about if you want. Also some other general concerns here and there, but nothing major 😄 Let me know if anything is unclear!

PR #3210, opened 2023-03-13
fix: [DHIS2-12520] change enrollment event from url

superskip:
Prevents the app from crashing when the users changes an event id through the url.
If the event id is invalid an error message is displayed.
Also, if the user removes the event id from the url, then the app navigates back to the main page.

The reported problem with the rule engine seems to have been already resolved in [DHIS2-14740](https://dhis2.atlassian.net/browse/DHIS2-14740).

[DHIS2-14740]: https://dhis2.atlassian.net/browse/DHIS2-14740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.js
```diff
@@ -44,24 +46,38 @@ const getPageStatus = ({ orgUnitId, enrollmentSite, teiDisplayName, trackedEntit
...
+    useEffect(() => {
+        if (!urlEventId) {
+            // return to main page
+            history.push(`/?${buildUrlQueryString({ orgUnitId })}`);
+        } else if (eventId !== urlEventId) {
+            dispatch(changeEventFromUrl(urlEventId, pageKeys.ENROLLMENT_EVENT));
+        }
+    }, [dispatch, history, eventId, urlEventId, orgUnitId]);
+
+    return !loading && eventId === urlEventId || error ? (
```
simonadomnisoru, line 67: IMHO the mix of `&&` and `||` is a bit confusing. Can you add some parentheses to clarify the intended order of operations? 
Thank you!

src/core_modules/capture-core/components/Pages/ViewEvent/ViewEventComponent/viewEvent.actions.js
```diff
@@ -4,6 +4,7 @@ import type { OrgUnit } from '@dhis2/rules-engine-javascript';
 
 export const actionTypes = {
     VIEW_EVENT_FROM_URL: 'ViewEventFromUrl',
+    CHANGE_EVENT_FROM_URL: 'viewEventFromUrl',
```
simonadomnisoru, line 7: To avoid confusion with the already existing `VIEW_EVENT_FROM_URL` key, does it make sense to use ``changeEventFromUrl'' instead? Something like this: `CHANGE_EVENT_FROM_URL: 'changeEventFromUrl'`
Thanks!

superskip: Actually that entry ended up not being used at all, so I'll remove it.

simonadomnisoru (approved):

JoakimSM (approved):
This is definitely an improvement, and the problem the ticket describes is fixed 👍 At some point we have to refactor the loading of the form in view event page though. Made a new ticket we can dig into at a later point: https://dhis2.atlassian.net/browse/TECH-1529

geethaalwan (approved):
Tested successfully on 2.41,2.40.1,2.39.3,2.38.4 versions

PR #3211, opened 2023-03-15
fix: [DHIS2-14190] restrict which values can be assigned to option sets

superskip:
Checks if a program rule "assign effects" is targeting an option set, and if so, checks that the value to be assigned lies in the option set. If the value does not lie in the option set, it is replaced with `null` (meaning no selected option).


JoakimSM (approved):
Looks good!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.41, 2.40.1,2.39.2,2.38.4 versions

PR #3213, opened 2023-03-15
feat: [DHIS2-14574] store program stage working lists

simonadomnisoru:
Load, add, update, delete, and share a program stage working list by consuming the new `/programStageWorkingLists` endpoint introduced in 2.40. The TEI and program stage working lists are shown together on the main and search page and the user can switch between the different types.

I introduced the concept of an altered default template to be able to maintain the configuration of one type of WL and use it and save it into another type of WL. The following 2 scenarios explain the steps:
- A TEI WL is selected and loaded. Add program stage from buttons. The TEI WL is unselected and the page redirects to the default template while keeping the TEI-specific data. A Program stage WL can then be created based on the initial TEI configuration.
- With PS WL is selected and loaded. Remove the program stage from the buttons. The program stage WL is unselected and the page redirects to the default template while keeping TEI-specific data. A TEI WL can then be created based on the initial program stage configurations that make sense for a TEI.


JoakimSM (approved):
Nice! Good work @simonadomnisoru

geethaalwan (approved):
Tested successfully on 2.41,240.1,2.40.0 versions

PR #3217, opened 2023-03-23
feat: [DHIS2-9661] first stage registration

jasminenguyennn:


simonadomnisoru:
The work on DHIS2-9661 will continue in https://github.com/dhis2/capture-app/pull/3267

PR #3226, opened 2023-04-04
fix: [DHIS2-14868] extend the clickable zone for search header

jasminenguyennn:


geethaalwan (approved):
Tested successfully on 2.40.1,2.40.0,2.41,2.39.2,2.38.4 versions

PR #3227, opened 2023-04-04
fix: [DHIS2-14868] extend the clickable zone for search header v37

jasminenguyennn:


geethaalwan (approved):
Tested successfully on 2.37.10 version

PR #3234, opened 2023-04-12
fix: [DHIS2-15109] eventDate breaks saved working lists

eirikhaugstulen:
Hey all,

When upgrading to the new tracker endpoints in 2.38, for some reason, we ended up saving the occurredAt-column in the working lists as eventDate. Not sure if this was a bug in the API at the time, or if it was just a slip up from my side, but it lead to breaking tests now. The API, which would return a status code 200 until now, has changed to return a 400 bad request. 

This PR resolves that we solve both saving correct label in the `eventFilters`-endpoints and ensures correct order-label. 

Steps to reproduce bug:

1. Open `/#/?orgUnitId=DiszpKrYNg8&programId=eBAyeGv0exc`
2. Do any changes (for example changing sort order on the event date)
3. Try to save a working list view

The API should throw an error and the app will crash.

### DIFF
--- a/src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/helpers/eventFilters/clientConfigToApiEventFilterQueryCriteriaConverter/convertToEventFilterEventQueryCriteria.js
+++ b/src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/helpers/eventFilters/clientConfigToApiEventFilterQueryCriteriaConverter/convertToEventFilterEventQueryCriteria.js
@@ -181,7 +181,7 @@ const structureFilters = (apiFilters: Array<Object>, columns: ColumnsForConverte
         dataFilters: [],
     });
 
-const getSortOrder = (sortById: string, sortByDirection: string) => `${sortById === 'occurredAt' ? 'occurredAt' : sortById}:${sortByDirection}`;
+const getSortOrder = (sortById: string, sortByDirection: string) => `${sortById}:${sortByDirection}`;
 
 const getColumnsOrder = (columns: Array<ColumnForConverter>) =>
     columns
@@ -204,7 +204,7 @@ export const convertToEventFilterEventQueryCriteria = ({
         () => typeConvertFilters(filters, columns),
         convertedFilters => structureFilters(convertedFilters, columns),
     )();
-    const displayColumnOrderCriteria = getColumnsOrder([...columns.values()])?.map(column => (column === 'occurredAt' ? 'occurredAt' : column));
+    const displayColumnOrderCriteria = getColumnsOrder([...columns.values()]);
 
     return {
         ...filtersCriteria,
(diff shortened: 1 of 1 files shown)


src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/helpers/eventFilters/clientConfigToApiEventFilterQueryCriteriaConverter/convertToEventFilterEventQueryCriteria.js
```diff
@@ -181,7 +181,7 @@ const structureFilters = (apiFilters: Array<Object>, columns: ColumnsForConverte
         dataFilters: [],
     });
 
-const getSortOrder = (sortById: string, sortByDirection: string) => `${sortById === 'occurredAt' ? 'eventDate' : sortById}:${sortByDirection}`;
+const getSortOrder = (sortById: string, sortByDirection: string) => `${sortById === 'occurredAt' ? 'occurredAt' : sortById}:${sortByDirection}`;
```
jasminenguyennn, line 184: can we simplify it to 
```
const getSortOrder = (sortById: string, sortByDirection: string) => `${sortById}:${sortByDirection}`;
```

src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/helpers/eventFilters/clientConfigToApiEventFilterQueryCriteriaConverter/convertToEventFilterEventQueryCriteria.js
```diff
@@ -204,7 +204,7 @@ export const convertToEventFilterEventQueryCriteria = ({
         () => typeConvertFilters(filters, columns),
         convertedFilters => structureFilters(convertedFilters, columns),
     )();
-    const displayColumnOrderCriteria = getColumnsOrder([...columns.values()])?.map(column => (column === 'occurredAt' ? 'eventDate' : column));
+    const displayColumnOrderCriteria = getColumnsOrder([...columns.values()])?.map(column => (column === 'occurredAt' ? 'occurredAt' : column));
```
jasminenguyennn, line 207: can we simplify this to 
```
const displayColumnOrderCriteria = getColumnsOrder([...columns.values()]);
```

superskip:
The test works on https://test.e2e.dhis2.org/ca-test-2.39 though. Which must mean that the problem was caused by a change in the backend. We can use the fix if it works with the older backend versions, let's do that by adding an appropriate label on this PR...

eirikhaugstulen:
The test passed, yes, but the functionality did not work. Previously, this was just treated as a silent error in the backend I guess, but now it treats it as a 400 - bad req. I think this fix is safe to backport. @superskip

superskip (approved):
LGTM! (just remember to push the changes suggested by @jasminenguyennn before merging)

JoakimSM (approved):

geethaalwan (approved):
Tested successfully on 2.41,2.40.0,2.40.1,2.39.2,2.38.4 versions

PR #3235, opened 2023-04-14
chore: non-mandatory dev-tests

JoakimSM:
As discussed: change the verify workflow to continue even though the dev-tests fail


eirikhaugstulen (approved):
👍

PR #3248, opened 2023-05-11
fix: [DHIS2-15153] ask user to create new event 

simonadomnisoru:
Improve the availableProgramStages in which is possible to add a new enrollment event


superskip (commented):
Looks good! :+1:

Shall we make one final adjustment by changing from `reduce` to `map`?

src/core_modules/capture-core/hooks/useAvailableProgramStages.js
```diff
@@ -20,24 +20,27 @@ export const useAvailableProgramStages = (programStage: ProgramStage, teiId: str
...
-
-            return accStage;
-        }, []), [
-        enrollment?.events,
-        program?.programStages, programLoading,
-        programStage.allowGenerateNextVisit,
-        programStage.id,
-    ]);
+                return { id: currentStage.id, isAvailableStage, eventCount, currentStage };
+            }, []),
```
superskip, line 35: Let's remove this extra empty bracket argument to avoid confusion. Looks good otherwise :+1:

superskip (commented):
Found one issue, see comment.

eirikhaugstulen (approved):
LGTM when @superskip's comment is fixed 👍

superskip (approved):

geethaalwan (approved):
Tested successfully on 2.41,2.40.1,2.39.3,2.38.4 versions

PR #3250, opened 2023-05-15
refactor: [TECH-1538] replace Material-UI Button

simonadomnisoru:
[TECH-1538](https://dhis2.atlassian.net/browse/TECH-1538)

**Tech summary**
Replace material-ui Button and clean some unused code

[TECH-1538]: https://dhis2.atlassian.net/browse/TECH-1538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eirikhaugstulen:
Hey @simonadomnisoru,
In the cases where we import from `components/Buttons/Button.component`, this component seems like it's already using the button from `@dhis2/ui`. I'm guessing the extra wrapper component is there so that we don't get to dependent on the ui-library in the future. Do you think it's sensible to continue using this wrapper component, or is there a reason for removing it?

JoakimSM (approved):
@eirikhaugstulen @simonadomnisoru Pros and cons definitely, but I'm in favour of removing the wrappers if they don't provide any functionality. A wrapper could make it easier to replace the dhis2 ui button as we wouldn't have to update the imports, but we would probably have to map props etc inside our wrapper component (add complexity).

(The wrapper component might have had some value when we were using material-ui, can't remember exactly)

PR #3267, opened 2023-05-25
feat: [DHIS2-9661][DHIS2-14830] first stage registration 

simonadomnisoru:
[DHIS2-9661](https://dhis2.atlassian.net/browse/DHIS2-9661). The PR builds on work done previously in https://github.com/dhis2/capture-app/pull/3217.

[DHIS2-9661]: https://dhis2.atlassian.net/browse/DHIS2-9661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**Tech summary**
Currently, the [D2Form](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/DataEntry/DataEntry.component.js#L304) is rendered between the TOP and BOTTOM dataEntrySections definitions. I introduced the concept of [placements.BEFORE_METADATA_BASED_SECTION](https://github.com/dhis2/capture-app/pull/3267/files#diff-2e6ef6a7cb7e9a1bccfe8a56dde791a9b96b50709cca7847af6dba187ff29315R5) to be able to render the stage dataEntrySections definitions in between the concatenated enrollment metadata and the stage metadata. The [getCustomContent](https://github.com/dhis2/capture-app/pull/3267/files#diff-fe42b756072b65cb18ced425bbd143771f52a77aa526f1546d75fe6e8f0eb146R279) callback implements the logic to render the `placements.BEFORE_METADATA_BASED_SECTION` dataEntrySections based on `beforeSectionId` value.

- [useBuildFirstStageRegistration](https://github.com/dhis2/capture-app/pull/3267/files#diff-bb9712267a40d0e8863bce92796e538fe7013fade6c30d57441f3de76e545c7eR6) retrieve the firstStageMetaData from IndexDB and computes the firstStage with access
- [useMergeFormFoundationsIfApplicable](https://github.com/dhis2/capture-app/pull/3267/files#diff-f3978cc79c6ad283ba98794118a188735c379feea555838ca539d5fc13616c36R16) when the `useFirstStageDuringRegistration` flag is active, the hook will create a new RenderFoundation which concatenates the enrollment and first program stage metadata. The concatenated formFoundation will be used by the `EnrollmentWithFirstStageDataEntry` and the program rules execution.
- [useDataEntrySections](https://github.com/dhis2/capture-app/pull/3267/files#diff-c22090259d1554a9d501267ff9dfac6429b980f1ea89893b9b16f41049515d27R8) defines the dataEntrySections of the EnrollmentWithFirstStageDataEntry form with the help of the new `placements.BEFORE_METADATA_BASED_SECTION` concept.
- [EnrollmentWithFirstStageDataEntry.component.js](https://github.com/dhis2/capture-app/pull/3267/files#diff-9ad47db2868d70e93cb93df54e6d0819829ebfd81f70c1e0d0c592ab44c19a61R1) is a DataEntry wrapper to build the form for the first program stage event.
- [deriveFirstStageDuringRegistrationEvent](https://github.com/dhis2/capture-app/pull/3267/files#diff-2e319649b110cbd2a562a39a908ce5ced7259b9d3541587f8510a94df3e4f93dR11) prepares the event data in the format expected by the API.
- [getStageWithOpenAfterEnrollment](https://github.com/dhis2/capture-app/pull/3267/files#diff-63523ed335ad1e9f1489cc31b35aa9a7f14a7a64d6283f5b0cd2bea31b2a1ee1R13) checks if the user should be redirected to /enrollmentEventNew or /enrollmentEventEdit after creating the TEI. The place where it is redirected depends if the event will be created or not. The event can be created either during first-stage registration or autogenerated.
- In RegistrationDataEntry.epics.js, the [eventIndex](https://github.com/dhis2/capture-app/pull/3267/files#diff-aa3ac034a144b650fbeb7c3c41b167548e367d86930581029188a9133fdab448R131) is calculated and will be used when the user is redirected to [/enrollmentEventEdit](https://github.com/dhis2/capture-app/pull/3267/files#diff-aa3ac034a144b650fbeb7c3c41b167548e367d86930581029188a9133fdab448R206). 
- Adapted the [convertGeometryOut](https://github.com/dhis2/capture-app/pull/3267/files#diff-edb681f429718c90bde6eb6d78d55898efb289757a59d0f14bf9b7232f8c8074R4) to be able to convert not only the `foundation.featureType` but also any customFeatureType like the [stageGeometry](https://github.com/dhis2/capture-app/pull/3267/files#diff-e0e2c9c45ca505532bab32b63946470793386cb1c742109e660d000f1cf5d8ceR22)
- Adapted the [getApplicableRuleEffectsForTrackerProgram](https://github.com/dhis2/capture-app/pull/3267/files#diff-2444c9088ff23fbc2c572ae8b50b722a0748dcd01a0c923afbb9fd41b5e56830R77) to accept a custom `formFoundation` and use it as the base for `foundationForPostProcessing`
- [convertAndGroupBySection](https://github.com/dhis2/capture-app/pull/3267/files#diff-c8374a338fd29eaa4050a7fd838fc55415fe63223adc505f2bed1908faf64041R167) in RenderFoundation 
    - loop the form values 
    - grab the associated data element
    - convert the value
    - group it by the section group
- [initMode](https://github.com/dhis2/capture-app/pull/3267/files#diff-aa3ac034a144b650fbeb7c3c41b167548e367d86930581029188a9133fdab448R210) - a new URL param used when navigating to the `/enrollmentEventEdit` page. When initMode is set up, this overrides any page logic on the initial load.


simonadomnisoru:
> Hey @simonadomnisoru, I've run into this issue where the required elements in the first stage is not treated the same ways of TEAs. The form treats them as valid after touch. See video below.
> 
>  Skjermopptak.2023-06-20.kl.16.11.06.mov

Hey @eirikhaugstulen,
Good catch! The bug is fixed by the latest commits. Let me know if you find any other issues.
Thanks for the feedback!

eirikhaugstulen (approved):
Hey @simonadomnisoru,
I think this looks really good. Approving as I think this is great as is 🎉
However, adding one small comment here, just so we can follow up wether it's a use case or not:
When running a rule in the first stage, trying to assign a value in the regular enrollment forms does not work. I.e. if you add a rule that only runs in the birth stage, having an action that assigns a value to the first name does not work. I'm guessing it's because the forms are encapsulated by a different key. Is this a use case or is it intentionally not meant to work this way?

simonadomnisoru:
> Hey @simonadomnisoru, I think this looks really good. Approving as I think this is great as is 🎉 However, adding one small comment here, just so we can follow up wether it's a use case or not: When running a rule in the first stage, trying to assign a value in the regular enrollment forms does not work. I.e. if you add a rule that only runs in the birth stage, having an action that assigns a value to the first name does not work. I'm guessing it's because the forms are encapsulated by a different key. Is this a use case or is it intentionally not meant to work this way?

Hey @eirikhaugstulen,
Thanks for mentioning this use case. We decided to divide the work into 2 tickets, so in this PR the program rules are not properly executed. I will have a look at them in [DHIS2-14830](https://dhis2.atlassian.net/browse/DHIS2-14830) and take your feedback into account. 
Thank you!

JoakimSM:
I think we should set up a meeting for this one, I have some input on the approach and structure. Some notes:

- `FirstStageRegistration.component`: We render two different instances of DataEntry with the same key. I'm very worried this is error prone. A problem I can think of from the top of my head: We set the loaded status in the Redux store for the key. At the moment the loaded status will be set to true when the first of the two instances is ready. This probably means that user can click save at this point before the second instance is ready and send invalid data (we could have some async operations running, like generating unique ids). We should rather concatenate the metadata and use one DataEntry instance.
- I think we should move all the files we added to `DataEntryDhis2Helpers`.
- I don't think `useBuildFirstStageRegistration` belongs in `useRegistrationFormInfoForSelectedScope`.

Also, not directly related to this ticket (but great to fix here I think), we should update the logic related to "first stage on registration page", "auto-generate event" and "open data entry after enrollment". I saw some bugs here, but we should do a refactor here anyway. Let's talk about it.

simonadomnisoru:
> I think we should set up a meeting for this one, I have some input on the approach and structure. Some notes:
> 
> * `FirstStageRegistration.component`: We render two different instances of DataEntry with the same key. I'm very worried this is error prone. A problem I can think of from the top of my head: We set the loaded status in the Redux store for the key. At the moment the loaded status will be set to true when the first of the two instances is ready. This probably means that user can click save at this point (before the second instance is ready) and send invalid data (we could have some async operations running, like generating unique ids). We should rather concatenate the metadata and use one DataEntry instance.
> * I think we should move all the files we added to `DataEntryDhis2Helpers`.
> * I don't think `useBuildFirstStageRegistration` belongs in `useRegistrationFormInfoForSelectedScope`.
> * AOC?
> 
> Also, not directly related to this ticket (but great to fix here I think), we should update the logic related to "first stage on registration page", "auto-generate event" and "open data entry after enrollment". I saw some bugs here, but we should do a refactor here anyway. Let's talk about it.

Hey @JoakimSM, 
I've updated the code with the approach and structure you suggested. Can you take a look?
Thanks!

JoakimSM (commented):
I have some suggestions on the introduced middle placement. The suggestions will require some refactoring of the `DataEntry.component`, but I think this is well overdue as the current logic is hard to reason about and not easily maintainable. But starting from the top, this is the idea:

Add a prop called `beforeSectionId` to the dataEntrySections that will be used when placement is `BEFORE_METADATA_BASED_SECTION`(`BEFORE_METADATA_BASED_SECTION` replaces `MIDDLE`). For the custom fields that goes into the sections, let’s remove the placement prop (i.e., we should only specify the section).

In `DataEntry.component` we should be able to reproduce the current behaviour, but hopefully with improved readability. For the sections with placement `BEFORE_METADATA_BASED_SECTION`, the callback idea you had is good, but I would modify it slightly. A callback like the following makes sense to me: 
customContent: `{ [beforeSectionId: string]: (beforeSectionId: string) => React.Node }`

A added bonus here is that we don’t have to add anything to RenderFoundation, a file I would really like to clean up at some point.

Minor one for the `DataEntry.component` refactor: Can we rename `FieldSection` to `Field` (or something along those lines)?

simonadomnisoru:
> I have some suggestions on the introduced middle placement. The suggestions will require some refactoring of the `DataEntry.component`, but I think this is well overdue as the current logic is hard to reason about and not easily maintainable. But starting from the top, this is the idea:
> 
> Add a prop called `beforeSectionId` to the dataEntrySections that will be used when placement is `BEFORE_METADATA_BASED_SECTION`(`BEFORE_METADATA_BASED_SECTION` replaces `MIDDLE`). For the custom fields that goes into the sections, let’s remove the placement prop (i.e., we should only specify the section).
> 
> In `DataEntry.component` we should be able to reproduce the current behaviour, but hopefully with improved readability. For the sections with placement `BEFORE_METADATA_BASED_SECTION`, the callback idea you had is good, but I would modify it slightly. A callback like the following makes sense to me: customContent: `{ [beforeSectionId: string]: (beforeSectionId: string) => React.Node }`
> 
> A added bonus here is that we don’t have to add anything to RenderFoundation, a file I would really like to clean up at some point.
> 
> Minor one for the `DataEntry.component` refactor: Can we rename `FieldSection` to `Field` (or something along those lines)?

Hey @JoakimSM,
I've updated the code with the `beforeSectionId` and `BEFORE_METADATA_BASED_SECTION` approach you suggested. Can you take a look?
Thank you!

JoakimSM (commented):
For the rules engine execution we will have to make sure we convert all the values (noticed that event values are not being converted to client-values).

My suggested approach (we can have a chat before you start implementing):

- Add a Section property to the DataElement class. The section property is optional and should reference the Section it belongs to.

- In The Section class, add a string property called group. Group is completely custom, in the case of first stage in registration form we can use «TRACKED_ENTITY» or «EVENT» as the value.

- In RenderFoundation add a function to group values based on their section group.
  - Loop values and for each value grab the associated data element. Look up the section group for the data element and save it in the following object structure: { [groupId: string]: { [dataElementId: string]: any }}

In runRulesOnEnrollmentUpdate in enrollment.epics:
Use the formFoundation you passed in to convert values, split the values in groups (as described above) etc (i.e. don’t use the program collection in memory directly for this anymore). 

I suggest we also implement more of the logic for for putting the values in the right place in the runRulesOnEnrollmentUpdate function (getCurrentEventValuesFromStage in runRulesOnUpdateFieldBatch I would like us to remove for example). 

I also noticed that the event `occurredAt` value was passed in too the rules execution among the enrollment values.

Also, I couldn’t get the AOC selection to show, but I didn’t investigate this part thoroughly.

But we are getting there, just a few more hurdles 👍

simonadomnisoru:
> For the rules engine execution we will have to make sure we convert all the values (noticed that event values are not being converted to client-values).
> 
> My suggested approach (we can have a chat before you start implementing):
> 
> * Add a Section property to the DataElement class. The section property is optional and should reference the Section it belongs to.
> * In The Section class, add a string property called group. Group is completely custom, in the case of first stage in registration form we can use «TRACKED_ENTITY» or «EVENT» as the value.
> * In RenderFoundation add a function to group values based on their section group.
>   
>   * Loop values and for each value grab the associated data element. Look up the section group for the data element and save it in the following object structure: { [groupId: string]: { [dataElementId: string]: any }}
> 
> In runRulesOnEnrollmentUpdate in enrollment.epics: Use the formFoundation you passed in to convert values, split the values in groups (as described above) etc (i.e. don’t use the program collection in memory directly for this anymore).
> 
> I suggest we also implement more of the logic for for putting the values in the right place in the runRulesOnEnrollmentUpdate function (getCurrentEventValuesFromStage in runRulesOnUpdateFieldBatch I would like us to remove for example).
> 
> I also noticed that the event `occurredAt` value was passed in too the rules execution among the enrollment values.
> 
> Also, I couldn’t get the AOC selection to show, but I didn’t investigate this part thoroughly.
> 
> But we are getting there, just a few more hurdles 👍


Hey @JoakimSM,
I've updated the code with the approach you suggested for rules engine execution. Can you take a look? 
Regarding `AOC` selection, it should appear when you configure the program as described in [EnrollmentDataEntry.component.js](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/DataEntries/Enrollment/EnrollmentDataEntry.component.js#L308). Did you find any specific cases where `getAOCSettingsFn` doesn't work correctly?
Thanks!

JoakimSM (commented):
We talked about making some changes here.

We should show the AOC selection if there are any program stages with «auto-generate event» or «show first stage on registration page» is selected for the program. We should use the same logic for the new and old dashboard.

This means that we should generate events for program stages with «auto-generate event» and «open data entry form after enrollment», even if we are using the new dashboard. The effect of «open data entry form after enrollment» will be that we navigate to the edit event page (not the new event page). We will have to add another url param to make this work I think (initMode=edit, this overrides any page logic on the initial load).

We should follow the logic described here for auto-generating the event: https://docs.google.com/document/d/1E5dBP5_BZZ0pLt573Xd0GQB6kve2WUpNdoew3kCFnUo/edit#bookmark=id.12acp8ay7jwz

simonadomnisoru:
> We should show the AOC selection if there are any program stages with «auto-generate event» or «show first stage on registration page» is selected for the program. We should use the same logic for the new and old dashboard.
> 
> This means that we should generate events for program stages with «auto-generate event» and «open data entry form after enrollment», even if we are using the new dashboard. The effect of «open data entry form after enrollment» will be that we navigate to the edit event page (not the new event page). We will have to add another url param to make this work I think (initMode=edit, this overrides any page logic on the initial load).
> 
> We should follow the logic described here for auto-generating the event: https://docs.google.com/document/d/1E5dBP5_BZZ0pLt573Xd0GQB6kve2WUpNdoew3kCFnUo/edit#bookmark=id.12acp8ay7jwz

Hey @JoakimSM, code updated. Can you take a look? Thanks!

JoakimSM (approved):
🎉 Nice, good work @simonadomnisoru.

geethaalwan (approved):
Tested successfully on 2.41,2.40.2,2.39.3,2.38.5 versions

PR #3269, opened 2023-05-26
fix: [DHIS2-15342] program stage WL trackedEntities request

simonadomnisoru:
The solution was to sanitize the `trackedEntityIds` by removing the first `;`. The PR also fixed the [Cypress tests](https://github.com/dhis2/capture-app/wiki/Pull-Request-Guidelines#cypress-tests-to-follow-up-on-the-tests-are-commented-out-from-our-code-base) related to the Program stage WL


superskip (approved):
It would probably be more efficient to use `push` instead of spread syntax in this kind of situation, but never mind, as this function doesn't appear to be performance critical.

EDIT: Never mind that, looks good to me :+1: :smile:

geethaalwan (approved):
Tested successfully on 2.41 version

PR #3293, opened 2023-06-05
fix: [DHIS2-15312] AOC for event programs

JoakimSM:
Disable editing AOC for event programs. Need a design before we implement.


simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2..41,2.40.1,2.39.3,2.38.5 versions

PR #3312, opened 2023-06-13
fix: [DHIS2-6335] use custom map center point

simonadomnisoru:
[DHIS2-6335](https://dhis2.atlassian.net/browse/DHIS2-6335)

**Tech summary**
- The orgUnit geometry is retrieved using `useApiMetadataQuery`(react query) by [useCenterPoint](https://github.com/dhis2/capture-app/pull/3312/files#diff-93958c46b8aeb88c9e3a29c0e5c181689694ffb9a825e5b71fbbab3790cec05fR18) hook in the WidgetEnrollment and [withCenterPoint](https://github.com/dhis2/capture-app/pull/3312/files#diff-299b3e0090590c612721fa80c9725df6516156ac83d44f8fd4b23067c18f6267R56) HOC in the rest of the forms. 
- The geometry is only retrieved when we open the map and no coordinate/polygon has been saved previously.
- When the orgUnit is missing the geometry information, a fallback is used to search the parent orgUnit until a viable one is found. The Map will default to London coordinates if no viable geometry info is found.
- The geometry information is converted based on its type (Point or Polygon) to Array<number> and passed to the Map component.

[DHIS2-6335]: https://dhis2.atlassian.net/browse/DHIS2-6335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/DataEntry/withCenterPoint.js
```diff
@@ -24,28 +32,24 @@ const getCenterPoint = (InnerComponent: ComponentType<any>) =>
...
+    if (orgUnit && !called) {
+        refetch({ variables: { orgUnitId: orgUnit.id } });
+    }
+    const center = useMemo(() => {
+        if (data?.organisationUnits) {
+            const { geometry, parent } = data.organisationUnits;
+            if (geometry) {
+                return convertToClientCoordinates(geometry);
+            } else if (parent?.id) {
+                refetch({ variables: { orgUnitId: parent.id } });
```
superskip, line 41: Elegant backtracing :+1:

src/core_modules/capture-core/components/WidgetEnrollment/Actions/AddLocation/AddLocation.component.js
superskip, line 1: Lots of changes disappeared here. Was that an accident?

superskip (commented):
Looks good in general, just wondering about some deletions in the last commit.

src/core_modules/capture-core/components/WidgetEnrollment/Actions/AddLocation/AddLocation.component.js
superskip: Since this file came from DHIS2-13237 there probably shouldn't be any problem with this I guess.

superskip (approved):

JoakimSM (commented):
Correct me if I'm wrong, but from my understanding we only need the center point when we open the map and no coordinate/polygon previously has been saved? If my assumption is correct, it would be great if we incorporate this into one of the map components that are loaded when we open the map (I didn't dig into exactly which one), so the api-request is made only when we actually need the data. I would use the `useApiMetadataQuery` (`useMetadataQuery` file) to do the fetching with react query. The hook will give us free caching as long as we only request one org.unit at a time (works perfectly here).

Can you have a look and see if this is feasible and then we can have a chat before you execute?

geethaalwan (approved):
Tested successfully on 2.41,2.41,2.40.4,2.39.6,2.38.7 versions

PR #3350, opened 2023-06-26
feat: [DHIS2-14334] edit enrollment date

superskip:
[DHIS2-14334](https://dhis2.atlassian.net/browse/DHIS2-14334)

Don't be scared by the large number of line changes, a lot of them comes from `yarn.lock`.

@cooper-joe Could you please have a look at the `style`-object in `EnrollmentDate.component`? I think the units of measurement might be wrong.

[DHIS2-14334]: https://dhis2.atlassian.net/browse/DHIS2-14334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.component.js
```diff
@@ -0,0 +1,147 @@
...
+        paddingTop: '6px',
+    },
+    buttonStrip: {
+        display: 'flex',
+        gap: `${spacersNum.dp4}px`,
+        margin: `${spacersNum.dp4}px 0`,
+    },
+    note: {
+        fontSize: '12px',
+        color: colors.grey600,
```
cooper-joe, line 60:
```suggestion
        color: colors.grey700,
```

cooper-joe (approved):
LGTM

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.js
```diff
@@ -74,13 +75,17 @@ export const EnrollmentPageDefault = () => {
...
+
     const onUpdateTeiAttributeValues = useCallback((updatedAttributeValues, teiDisplayName) => {
         dispatch(updateEnrollmentAttributeValues(updatedAttributeValues
             .map(({ attribute, value }) => ({ id: attribute, value })),
         ));
         dispatch(updateTeiDisplayName(teiDisplayName));
     }, [dispatch]);
 
+    const onUpdateEnrollmentDate = useCallback(enrollmentDate =>
+        dispatch(updateEnrollmentDate(enrollmentDate)), [dispatch]);
```
simonadomnisoru, line 87: Hey @superskip,
IMHO the enrollment date in the top ScopeSelector should be updated without requiring a page refresh. The `WidgetProfile` currently handles it in`onUpdateTeiAttributeValues` and `updateTeiDisplayName(teiDisplayName)`. Does it make sense to implement something similar for the enrollment date? 

https://github.com/dhis2/capture-app/assets/6817073/e0ce501e-bacd-45e1-b62b-db550a519d80

Thanks you!

superskip: Good observation! Will fix :)

simonadomnisoru (approved):

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.component.js
```diff
@@ -106,8 +107,10 @@ export const EnrollmentPageDefaultPlain = ({
                     teiId={teiId}
                     enrollmentId={enrollmentId}
                     programId={program.id}
+                    mayEditDate
```
JoakimSM, line 110: Can we go with `readOnlyMode` even though the mode is not complete in this PR? I understand your thinking, but I think it will be easier for someone to finish up this later if we introduce it here.

And `readOnlyMode` should be optional (defaults to false). Meaning you can omit it in this file.

JoakimSM (commented):
Thanks @superskip.

Some comments in the code, but in addition we should implement the same for "incident date" (you might have mentioned that you wanted a review first). The incident date should have the same information regarding auto-generated events.

src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.component.js
```diff
@@ -0,0 +1,147 @@
...
+    };
+
+    return editMode ? (
+        <div data-test="widget-enrollment-enrollment-date">
+            <CalendarInput
+                calendar="gregory"
+                dense
+                className={classes.calendar}
+                label={enrollmentDateLabel}
+                inputWidth="50%"
```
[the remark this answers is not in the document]
superskip: There is already lower bound on how narrow the enrollment widget can get. If we use `inputWidth="200px"` then this lower bound will increase by a little. If we instead say `inputWidth="180px"` then the lower bound will increase even less (if at all) while at the same time we will get a width matching better with that shown in the pictures in the jira ticket. By the way there is no `maxWidth` prop, so I'll go with `inputWidth="180px"`.

src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.container.js
```diff
@@ -0,0 +1,51 @@
...
+    displayAutoGeneratedEventWarning,
+    onUpdateDate,
+    classes,
+}: Props) => {
+    const [updateMutation] = useDataMutation(enrollmentUpdate);
+
+    const saveHandler = (selectedDate: string) => {
+        enrollment.enrolledAt = convertValueClientToServer(selectedDate, dataElementTypes.DATE);
+        updateMutation(enrollment);
+        onUpdateDate && onUpdateDate(enrollment.enrolledAt);
```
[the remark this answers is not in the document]
superskip: Which part of the UI are you thinking of here? The date in the enrollment widget is rendered when the states `editMode` and `selectedDate` (in enrollmentDate.compontent) changes. Anyway, I think perhaps the change you had in mind got implemented in the error handling commit :slightly_smiling_face:

src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.component.js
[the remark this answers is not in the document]
JoakimSM: I think a maxWidth of 200 would be best. Let's go for a wrapping div with a maxWidth of 200.

src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.container.js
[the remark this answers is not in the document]
JoakimSM: Not exactly what I was looking for.

We should use optimistic UI and update the enrollment date instantly when the user changes it. Meaning; Before we get confirmation from the api, the enrollment date in the Widget UI should update and we should invoke the external onUpdate callback. If the request fails, we should revert the value to the original one. Also, in your saveHandler (EnrollmentDate.container) it is fine to invoke updateMutation directly, don't need to use an useEffect.

I think it makes most sense to update the enrollment object from useEnrollment (EnrollmentDate.container) instead of using a local component state.

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.container.js
```diff
@@ -33,13 +45,17 @@ export const WidgetEnrollment = ({ teiId, enrollmentId, programId, onDelete, onA
...
             program={program}
             refetchEnrollment={refetchEnrollment}
             refetchTEI={refetchTEI}
             ownerOrgUnit={{ id: ownerOrgUnit, displayName }}
             loading={!(enrollment && program && displayName)}
             onDelete={onDelete}
             onAddNew={onAddNew}
+            onUpdateEnrollmentDate={onUpdateEnrollmentDate}
+            onUpdateIncidentDate={onUpdateIncidentDate}
             error={error}
```
JoakimSM, line 59: can we change this to `initError`? (hard to keep track of the errors now when we also have error callbacks when updating)

src/core_modules/capture-core/components/WidgetEnrollment/EnrollmentDate/EnrollmentDate.component.js
```diff
@@ -0,0 +1,151 @@
+// @flow
```
JoakimSM, line 1: NIT: Can we change the name of this file since we are using it for both enrollment date and incident date? A bit hard since it is the enrollment Widget (so EnrollmentDate makes some sense). EnrollmentWidgetDate maybe? (feel free to come up with something better)

superskip:
Renamed `EnrollmentDate` -> `Date`.

@JoakimSM - I didn't like how the code turned out after doing the thing we discussed (exposing `setEnrollment`).
Instead I made it so that `enrollment` gets refetched once the update request succeeds. As a consequence the `enrolledAt` and `occurredAt` values will not be up to date while the request is being processed. Therefore, to display the new date value while the request gets processed I had to introduce a state variable (in `Date.container.js`).

JoakimSM (approved):
Good job 👍

geethaalwan (approved):
Tested successfully on 2.41,2.40.2,2.38.5,2.39.3 versions

PR #3393, opened 2023-07-28
feat: [DHIS2-12615] Add new TEI-relationship (search)

eirikhaugstulen:
Add a new TEI relationship through search (Capture-parity-version)

Tech-summary:
- Added an optional prop called `renderTrackedEntitySearch` with props and callbacks
- Copied the code from `TeiSearch` (old relationships) and refactored to fit the workflow of the relationship widget
- Build a mutation-hook that should work for both search and add new TEI. This mutation takes in `apiData` which is sent to the server and a `clientData` that updates the UI
- Relationships are invalidated and refetched when the mutation completes

Question points:
- [ ] We do not show that the optimistically updated relationship is not persisted. Should this follow the same standard of `StagesAndEventsWidget` with a greyed out row?
- [ ] When adding a relationship we do not have information about the dataElementType of the TEI attributes, which might lead to wrong formatting
- [ ] Should we follow the same structure of removing the relationship when there's an error?


simonadomnisoru (commented):
Hey Erik,
Great job on this PR 👏 ! Some initial thoughts:

- Currently, the search returns TEIs that are already linked. When linking to these entities, the API returns a `409` error. Should we just remove them from the displayed search results or perhaps let the query fail and explain the error to the user in the bottom SnackBar?
- Is there a technical reason why you copied `TeiSearch` to the Pages folder instead of adapting and reusing the existing `TeiSearch` component?

Suggestions for your question points:
1) It makes sense to me to use the same approach as `StagesAndEventsWidget` (greyed out row, click disabled, and a tooltip), but perhaps you can confirm it first with the design team.
2) The data must be converted to the server format before adding the relationship. The TEIs attributes `valueType` is available in `item.tei.attributes` inside [onAddRelationship](https://github.com/dhis2/capture-app/blob/eh/feat/DHIS2-12615_SearchExistingTEIRelationship/src/core_modules/capture-core/components/Pages/NewRelationship/TeiRelationship/SearchResults/TeiRelationshipSearchResults.component.js#L80). You can pass it to the callback and then use it in [onLinkToTrackedEntityFromSearch](https://github.com/dhis2/capture-app/pull/3393/files#diff-82f539443aea0d260560f1d5a4fb765ef1a507c6ea82dc5e2ca6f5e87478251bR59)
3) I agree that it makes sense to remove the relationship when there is an error, just as `StagesAndEventsWidget` currently does. 

Let me know if anything is unclear. Thanks!

eirikhaugstulen:
Superseded by #3401

PR #3398, opened 2023-08-01
fix: [DHIS2-15365] remove orgUnit and ouMode query args

simonadomnisoru:
[DHIS2-15365](https://dhis2.atlassian.net/browse/DHIS2-15365)


superskip (approved):

linadhis2:
Verified that the query args are no longer present using debug environments on dev and DHIS2 Version 2.40.1-SNAPSHOT on DHIS2 Revision aebca38.

PR #3399, opened 2023-08-07
chore: [TECH-1616] adapt cypress tests

simonadomnisoru:


superskip (commented):
Thanks for looking into this @simonadomnisoru :pray: 
I see there are still some failing tests on the e2e servers, are those something that should be handled as well in this PR?

simonadomnisoru:
> Thanks for looking into this @simonadomnisoru 🙏 I see there are still some failing tests on the e2e servers, are those something that should be handled as well in this PR?

Hey @superskip,
This PR fixes the failures in `ca-test-dev` but causes the [same tests](https://cloud.cypress.io/projects/322xnh/runs/9490/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D) to fail against the `ca-test-2.40/ca-test-2.39/ca-test-2.38` instances. This happens because we test against stable releases and the backported API fix for [DHIS2-15325](https://dhis2.atlassian.net/browse/DHIS2-15325) has not yet been released. 
We have to wait for `2.40.1`, `2.39.3` and `2.38.5` to be released and then redeploy the `ca-test-2.40/ca-test-2.39/ca-test-2.38` instances. I'm not sure what the release schedule is and it may take a few weeks/months. To speed up the process and avoid waiting, one solution is to test against the latest 2.40dev/2.39dev/2.38dev until their stable versions are deployed. What do you think @superskip and @JoakimSM?
Thank you!

superskip (approved):
> To speed up the process and avoid waiting, one solution is to test against the latest 2.40dev/2.39dev/2.38dev until their stable versions are deployed.

I think this sounds like a very good idea :slightly_smiling_face:
Does it mean we go to AWX, remove all ca-test-2.xx instances, and spin up new three new instances called ca-test-2.xx using the war files named 2.xx? If so I guess this PR is ready for approval :heavy_check_mark:

PR #3403, opened 2023-08-15
feat: [DHIS2-15299] escape value for attribute filter

simonadomnisoru:
[DHIS2-15299](https://dhis2.atlassian.net/browse/DHIS2-15299)

**Tech summary**
Escape `,:/` in the value of the filter in the following places:
- working list - download events
- working lists (init and update) 
    - TEIs
    - program stage events
    - single events 
- search
    - by unique id
    - by attributes
    -  possible duplicate
    - from adding a relationship
- validate unique values
    - new enrolment
    - widget profile
    - new TET

### DIFF
--- a/cypress/integration/SearchPage.feature
+++ b/cypress/integration/SearchPage.feature
@@ -37,14 +37,6 @@ Feature: User interacts with Search page
     #     And you click search
     #     Then you should see no results found
 
-    Scenario: Searching using attributes in Tracker Program throws error
-        Given you are on the default search page
-        When you select the search domain Malaria Case diagnosis
-        And you expand the attributes search area
-        And you fill in the first name with values that will return an error
-        And you click search
-        Then there should be an generic error message
-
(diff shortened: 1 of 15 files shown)


cypress/integration/SearchPage.feature
```diff
@@ -37,14 +37,6 @@ Feature: User interacts with Search page
     #     And you click search
     #     Then you should see no results found
 
-    Scenario: Searching using attributes in Tracker Program throws error
-        Given you are on the default search page
-        When you select the search domain Malaria Case diagnosis
-        And you expand the attributes search area
-        And you fill in the first name with values that will return an error
```
simonadomnisoru, line 44: With the filter value escaped, this scenario is obsolete. Please let me know if you know of another way to make the API request fail and I can adapt the test instead of removing it.

superskip (approved):
LGTM!

(Can't think of any values that would make a tei attribute search fail)

PR #3407, opened 2023-09-12
feat: [DHIS2-15830] Add orgUnitId and programId to plugin context

eirikhaugstulen:
Tech summary:
We need to pass in ProgramId and orgUnitId as part of the context for the form field plugin.

### DIFF
--- a/src/core_modules/capture-core/components/D2Form/FormFieldPlugin/FormFieldPlugin.container.js
+++ b/src/core_modules/capture-core/components/D2Form/FormFieldPlugin/FormFieldPlugin.container.js
@@ -6,11 +6,13 @@ import { usePluginMessages } from './hooks/usePluginMessages';
 import { usePluginCallbacks } from './hooks/usePluginCallbacks';
 import { usePluginValues } from './hooks/usePluginValues';
 import { formatPluginConfig } from './formatPluginConfig';
+import { useLocationQuery } from '../../../utils/routing';
 
 export const FormFieldPlugin = (props: ContainerProps) => {
     const { pluginSource, fieldsMetadata, formId, onUpdateField, pluginContext } = props;
     const metadataByPluginId = useMemo(() => Object.fromEntries(fieldsMetadata), [fieldsMetadata]);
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/components/D2Form/FormFieldPlugin/FormFieldPlugin.container.js
```diff
@@ -6,11 +6,13 @@ import { usePluginMessages } from './hooks/usePluginMessages';
 import { usePluginCallbacks } from './hooks/usePluginCallbacks';
 import { usePluginValues } from './hooks/usePluginValues';
 import { formatPluginConfig } from './formatPluginConfig';
+import { useLocationQuery } from '../../../utils/routing';
 
 export const FormFieldPlugin = (props: ContainerProps) => {
     const { pluginSource, fieldsMetadata, formId, onUpdateField, pluginContext } = props;
     const metadataByPluginId = useMemo(() => Object.fromEntries(fieldsMetadata), [fieldsMetadata]);
     const configuredPluginIds = useMemo(() => Object.keys(metadataByPluginId), [metadataByPluginId]);
+    const { programId, orgUnitId } = useLocationQuery();
```
simonadomnisoru, line 15: Hey,
What about cases where `programId` is missing from URL parameters (such as on `/viewEvent` or `/enrollmentEventEdit` pages)? Should we take it from somewhere else in these cases? 🤔 
Thanks!

eirikhaugstulen: That's a good point. Also, I didn't think this quite through I guess. Instead of sending in programId, we should perhaps send in the contextId, i.e. either programId or TETid? Also, they configure the plugin by this contextId, so it should be familiar. WDYT?

simonadomnisoru: Do they need to know if the plugin is injected within a program or a TET context? If not, then a generic `contextId` makes sense to me.

JoakimSM (approved):

PR #3413, opened 2023-09-13
fix: [DHIS2-15383] align mandatory behaviour for all value types

superskip:
Jira: [DHIS2-15383](https://dhis2.atlassian.net/browse/DHIS2-15383)

The first three commits are completely independent of each other, so I recommend viewing them separately.

**The first commit** is a bit complex, so let me break it down a bit:

It concerns itself with a home-made low-level component, acting as a container for a collection of checkboxes.
This component supports keyboard input: arrow keys to move between checkboxes, enter/space to select a checkbox.
However, the original implementation didn't properly handle the case when a key is being held down over time.
The new file `withKeyboardNavigation` aims to improve the handling in this situation.

Another central feature of this commit is that it introduces an array of refs to the checkboxes.
These are used to keep the container from validating when the focus moves within it (which typically happens when an arrow key is pressed).

**The third commit** doesn't really address the reported issue, but it is sort of related in that it also contributes to alignment of the behaviour of all fields, specifically the result of `d2:hasValue` in the rule engine.

All the subsequent commits are only minor adjustments needed for passing the cypress tests.

[DHIS2-15383]: https://dhis2.atlassian.net/browse/DHIS2-15383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JoakimSM (commented):
Some minor stuff in addition to what we talked about in person.

superskip:
The merge I did from master accidentally got split up into several commits during a rebasing operation..

@JoakimSM - I fixed the keyboard navigation problem and the file selector button problem that you showed me, as well as correcting what you pointed out in the comments.

simonadomnisoru (commented):
Hey @superskip,
The mandatory error is not displayed when leaving a field with value type `MULTI_TEXT` (probably some logic is missing in [MultiSelectField.component.js](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/FormFields/New/Fields/MultiSelectField/MultiSelectField.component.js#L47)). 
In addition, after submitting the form, the required error text is displayed, but the field background is not red. Can take a look?



https://github.com/dhis2/capture-app/assets/6817073/8b000ad5-71ea-498e-a869-d38560451ed9


Besides this, the rest looks great. Thanks!

superskip:
Thanks for the review @simonadomnisoru !
I ended up wrapping `MultiSelecFieldUI` in a div and get a lot of line changes because of increased indentation :neutral_face:
Possibly it would have been better to change the UI library component so that it calls the `onBlur` prop when tabbing out.

simonadomnisoru (approved):
LGTM

JoakimSM (approved):
nice

geethaalwan (approved):
Tested successfully on 2.41,2.40.2,2.39.4,2.38.6 versions

PR #3415, opened 2023-09-14
chore: [TECH-1539] Use Capture-UI IconButton

eirikhaugstulen:
Summary:
- Replaced Material-UI IconButtons with Capture-UI
- Created a ChevronIcon with the animations and etc. (Placed it in Capture-UI, but not sure if that's correct. Open for input)
- Added some styling to the IconButton component, as well as some additional styling here and there


src/core_modules/capture-ui/IconButton/IconButton.component.js
```diff
@@ -1,10 +1,36 @@
...
-import classes from './iconButton.module.css';
+import { withStyles } from '@material-ui/core/styles';
+import { colors } from '@dhis2/ui';
 import type { Props } from './iconButton.types';
 
-export const IconButton = ({ children, className, dataTest, onClick, ...passOnProps }: Props) => {
+const styles = {
+    button: {
+        cursor: 'pointer',
+        borderRadius: '3px',
```
simonadomnisoru, line 11: To maintain the current round border can we use `border-radius: 50%;` instead?

src/core_modules/capture-ui/IconButton/IconButton.component.js
```diff
@@ -15,12 +41,18 @@ export const IconButton = ({ children, className, dataTest, onClick, ...passOnPr
...
+        return onClick(event);
+    }, [disabled, onClick]);
+
     return (
         <span
             {...passOnProps}
-            onClick={onClick}
+            onClick={onClickHandler}
+            disabled={disabled}
             data-test={dataTest}
```
simonadomnisoru, line 54: It seems that many Cypress tests are failing due to missing `dataTest` prop. Can you have a look?

src/core_modules/capture-ui/IconButton/IconButton.component.js
eirikhaugstulen: Good point, but it was decided that these should look like the buttons provided from UI and the already existing "IconButtons" in our widgets. Is that ok or should we mention it in the design-meeting? 😊

simonadomnisoru: Sounds good! If you add a comment in the ticket about this decision should be enough. Thanks!

simonadomnisoru (approved):

PR #3421, opened 2023-09-27
fix: [DHIS2-15879] Unable to schedule event with note

eirikhaugstulen:
TECH Summary:
The API throws an error in the current scheduling functionality, as the provided ID is not in the correct format. Therefore, I have switched to the same algorithm we use in Relationships and Referrals


simonadomnisoru (approved):
LGTM! 👏 
NIT: Add a Cypress scenario to check that the user can successfully schedule an event with a note. Thanks!

geethaalwan (approved):
Tested successfully on 2.41,2.40.2,2.39.4,2.38.5 versions

PR #3429, opened 2023-10-02
chore: [TECH-1654] upgrade cypress to v12

eirikhaugstulen:
Ran into some performance issues, so decided trying to update to v12.
There were however some breaking changes, and because of this, VCS got really hard to track. Therefore, here are the main headers:
- All of our e2e tests are now moved from cypress/integration/**  => cypress/e2e/**
- Cucumber was deprecated by the previous maintainer, so I switched to the @badeball/cypress-cucumber-preprocessor package. (same code, just updated)
- `And` is deprecated by cucumber, but for now we import { defineStep as And } and it will work the same way
- We now cache both network calls and user sessions, which should greatly improve performance (at least it does locally)
- Cypress is now a lot more strict on duplicate test step names, so I've deleted a lot of tests that were both in sharedSteps and the index-files.


superskip:
> * We now cache both network calls and user sessions, which should greatly improve performance (at least it does locally)

Sounds awesome :smile:

simonadomnisoru (approved):
LGTM!

JoakimSM (approved):
🎉

PR #3435, opened 2023-10-16
fix: [DHIS-15621] row is broken up in sections

henrikmv:


src/core_modules/capture-core/components/D2Form/D2Section.component.js
```diff
@@ -87,7 +88,10 @@ class D2SectionPlain extends React.PureComponent<Props> {
             );
         }
         return (
-            <div>
+            <div
+                className={applyCustomFormClass ? this.props.classes.containerCustomForm : ''}
+                data-test="d2-form-component"
```
simonadomnisoru, line 93: Hey @henrikmv,
There seem to be a lot of broken cypress tests. Could it be related to the `data-test="d2-form-component"` selector you moved?
Thanks!

superskip: @henrikmv Note that you don't need to fix all tests: you can ignore all where an orgunit name is missing (looks like there are six of them - https://cloud.cypress.io/projects/322xnh/runs/9730/overview?roarHideRunsWithDiffGroupsAndTags=1).

simonadomnisoru: @henrikmv We track known cypress bugs on the [repository wiki](https://github.com/dhis2/capture-app/wiki/Pull-Request-Guidelines#cypress-tests-to-follow-up-on-the-tests-are-commented-out-from-our-code-base). You only need to look into the failed tests that are not on that list. Thanks!

superskip: @simonadomnisoru  Hmm, it doesn't seem to quite agree with the errors I got :thinking:

PR #3436, opened 2023-10-17
refactor: [TECH-1657] standardize format of redux orgunits

superskip:
[TECH-1657](https://dhis2.atlassian.net/browse/TECH-1657)

I've put some effort into arranging the commits.
Might be a good idea to open the first commit in a separate window for easy reference.

[TECH-1657]: https://dhis2.atlassian.net/browse/TECH-1657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superskip:
Hey @JoakimSM, thanks for the review!
What better way to describe this type of orgunit than naming it's origin? Maybe what it is compatible with, like `RulesEngineCompatibleOrgUnit`. But it is supposed to be compatible to more than just the rules engine, so that particular name would be a bit misleading. Adding all other current compatibility dependencies to the name has obvious maintainability issues. Simply `DetailedOrgUnit` again looks rather generic. Maybe something like `TonyOrgUnit` (or any other name-like prefix) is the best option. The fact that the object value is cached locally, though, would that be invariably true in the future? If so we should maybe name it after this property. Oh, maybe a better property which I also think is sort of true is that it wants to be everybody's friend. Actually, how about `UniversallyCompatibleOrgUnit` (too long, potentially misleading) or `CompatibleOrgUnit` (maybe too generic). I like the ring to `ReduxOrgUnit`, also it has got a name-like prefix and has cached value connotations...

JoakimSM (approved):
Good job 👍

PR #3451, opened 2023-10-30
docs: [DHIS2-15483] assign user when reporting or scheduling an enrollment event

simonadomnisoru:


docs/user/using-the-capture-app.md
```diff
@@ -1193,6 +1193,15 @@ Instead of reporting an event the user can select to schedule an event for later
 
 The suggested date for the first event of a program stage in an enrollment is always based on the enrollment date or the incident date (depending on the program configuration). The program stage configuration setting "scheduled days from start" will be added to the base date to cumpute the suggested date.
 
+### Assigning new events
+When reporting or scheduling an event, you can assign a user to it. This feature must be enabled per program stage in a tracker program by clicking the "Allow user assignment of events" check box.
+
+You will find the assignee section near the bottom of the data entry page. Search for and select the user you would like to assign the event to. The assignee will be preserved when you save the event.
+
+![](resources/images/user_assignment_report_new.png)
```
eirikhaugstulen, line 1201: We don't usually mention how to set up and configure features in our user docs, but I'm not sure where else to put it. If this is right, then it looks good! 🎉

@karolinelien any insights?

eirikhaugstulen (approved):
👍

docs/user/using-the-capture-app.md
karolinelien: From my side I am fine with explaining how to enable this feature here, I definitely haven't followed the rule of not mention how to configure things before 👀

karolinelien:
My only comment is that I would perhaps have the title be "Assigning user to new events", not just "Assigning new events"

karolinelien (approved):
LGTM!

PR #3453, opened 2023-10-30
feat: [DHIS-11419] display assigned users on events in enrollment overview page

superskip:
Jira issue description: 

>Stages and events Widget: If the option "Allow user assignment of events" is set for the program stage we should display the assignee in the table.

The program Malaria focus registration can be used for testing.


src/core_modules/capture-core/converters/clientToList.js
```diff
@@ -92,7 +92,7 @@ const valueConvertersForType = {
     [dataElementTypes.FILE_RESOURCE]: convertResourceForDisplay,
     [dataElementTypes.IMAGE]: convertResourceForDisplay,
     [dataElementTypes.ORGANISATION_UNIT]: (rawValue: Object) => rawValue.name,
-    [dataElementTypes.ASSIGNEE]: (rawValue: Object) => `${rawValue.name} (${rawValue.username})`,
+    [dataElementTypes.ASSIGNEE]: (rawValue: Object) => `${rawValue.name || rawValue.displayName} (${rawValue.username})`,
```
simonadomnisoru, line 95: **NIT**: There is an edge case where `displayName` is undefined. To reproduce, create a new enrollment event, you will be redirected to the enrollment overview and the `undefined (undefined)` text will appear in the program stage table. You must refresh the page to see the correct value of `displayName`. 

Screenshot 2023-11-01 at 15 58 47


This happens because only the `uid` property is stored in Redux (within `enrollmentDomain`) when the event is created. The solution is quite straightforward and I already needed to fix it in one of the other related tickets in the epic [DHIS2-15480](https://github.com/dhis2/capture-app/pull/3412/files#diff-af963d544e60c1aea70db6cc2be6e5cd4f0bb4901cdbf40de87648e3b5a2c45bR29).

Thanks!

simonadomnisoru (approved):
LGTM! 👏

src/core_modules/capture-core/converters/clientToList.js
superskip: Thanks @simonadomnisoru! Implemented the fix so that it won't come up as an issue during testing. I tried undoing the change I did on this line, but that didn't work. Seems like removing `rawValue.name` works, though. Do you think I should?

simonadomnisoru: To undo the change, you can add a converter for `dataElementTypes.ASSIGNEE` inside [serverToClient](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/converters/serverToClient.js#L38). Something like this should work:

```
    [dataElementTypes.ASSIGNEE]: (serverValue: Object) => ({
        id: serverValue.uid,
        name: serverValue.displayName,
        username: serverValue.username,
        firstName: serverValue.firstName,
        surname: serverValue.surname,
    }),
```
This ensures that when the assignedUser value reaches the `clientToList` converter, it is in the correct format. Thanks!

superskip: No luck with that one.. I suppose we'd better keep both `rawValue.name

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/hooks/useProgramMetadata.js
```diff
@@ -8,7 +8,7 @@ const query = {
         id: ({ id }) => id,
         params: {
             fields:
-            ['programStages[id,repeatable,hideDueDate,programStageDataElements[displayInReports,dataElement[id,valueType,displayName,displayFormName,optionSet[options[code,name]]]'],
+            ['programStages[id,repeatable,hideDueDate,enableUserAssignment,programStageDataElements[displayInReports,dataElement[id,valueType,displayName,displayFormName,optionSet[options[code,name]]]'],
```
JoakimSM, line 11: Do you know the reason why `useProgramMetadata` is retrieving data from the api instead of using the IndexedDB cache? If not, can you change it to use `useIndexedDBQuery`?

superskip: I'm not sure, but it could perhaps be related to translations, i.e. `displayName` needs to come from the server?

JoakimSM: We have a ticket to fix the displayName cache problem. Great if you change this one to use `useIndexedDBQuery`

JoakimSM: If I understand you correctly, there is a separate store/table for data elements. Should be retrieved from there.

superskip: Changed the metadata source in this case from the server to IndexedDB. The code got rather complicated the way I did it though. Got any suggestions for making it less complicated?

JoakimSM: It would be beneficial to have one approach for this and currently that is `useIndexedDBQuery`. If we don't like the approach there, we can look into what we don't like and change it for all queries to IndexedDB. But let's start by using `useIndexedDBQuery`.

JoakimSM (approved):
Looks good @superskip! 🎉

geethaalwan (approved):

PR #3456, opened 2023-11-02
feat: [DHIS2-15212] Referral actions - Enter data now

eirikhaugstulen:
TECH Summary:
- Added option to enter referral data now. This creates a scheduled event in the same context as request and links it with a relationship.
- (If selected) we do not use optimistic UI to add events, but rather add them on success
- (If selected) we do not navigate the user until persisted in the server
- (If selected) we disable the buttons or show a loading spinner while we wait for the API
- (If selected) we store some information in the redux store and check this information to block navigation if the user has already navigated away


eirikhaugstulen:
Superseded by https://github.com/dhis2/capture-app/pull/3458

PR #3457, opened 2023-11-03
fix: [DHIS-15621] row view broken up in section

henrikmv:


src/core_modules/capture-core/metaDataMemoryStoreBuilders/common/factory/searchGroup/SearchGroupFactory.js
```diff
@@ -124,7 +124,7 @@ export class SearchGroupFactory {
     async _buildSection(searchGroupAttributes: Array<SearchAttribute>) {
         const section = new Section((o) => {
             o.id = Section.MAIN_SECTION_ID;
-            o.showContainer = false;
+            o.showContainer = true;
```
simonadomnisoru, line 127: Hey @henrikmv ,
Can you tell me what are the steps to reproduce the scenario you fixed by setting `showContainer` to `true`? I don't really understand why the searchForm component is affected by your changes. 
Thanks!

henrikmv: Hi @simonadomnisoru, thank you for your question. 

To fix the horizontal view, I have removed a div-element from the render function in `D2Section.component.js.` Therefore, I also had to move the `data-test` prop (and also a className prop). Without an alternative place for the data-test prop so that it applies to both rendering with and without section in the D2section.component, the data test prop now only applies when we are rendering with sections. I have set `o.showConatiner` to be true so that the `if statement on line 78` of `D2Section.component.js` should become false. In this way, the data test prop is included also for the search form.

JoakimSM: I think @simonadomnisoru's suggestion will work, but let’s see if we can break this down and simplify it. It was hard to reason about the original code (before your changes) so a solid refactor is due. Suggestion:

1. `renderSection` function: The main part here is that we should render `D2SectionFields`. `sectionMetaData.showContainer` decides if `D2SectionFields` should be wrapped in a Section-component or not. D2SectionFields should be passed the same props, regardless of whether we wrap it or not. common pattern:
```
	const  sectionFields = (<D2SectionFields……/>);
	
	if (sectionMetadata.showContainer) {
		return (
			<Section ….. >
				{sectionFields}
			</Section>
	);
	
	return sectionFields;
```

2. In the main render function let’s re-add the div with a data-test (name: «d2-section») and the original className (the conditional `cointainerCustomForm`) (the class could potentially have been moved to another place, but it's fine to leave it here for now)

3. In the `D2SectionFields.component` wrap everything in a `div`. The div should have the data-test «d2-section-fields» and it should conditionally add a class that sets `display: ‘flex’` (and the others css styles @simonadomnisoru suggested) when `props.formHorizontal` is true.

The tests will have to be updated. Most of them can probably use data-test «d2-section».

src/core_modules/capture-core/components/D2Form/D2SectionFields.component.js
```diff
@@ -269,20 +278,27 @@ export class D2SectionFieldsComponent extends Component<Props> {
 
         this.buildRulesCompulsoryErrors();
 
+        const styles = getStyles();
```
simonadomnisoru, line 281: Can we remove this function call and just use the `style` object directly? To make it work without the function, you can change line [21](https://github.com/dhis2/capture-app/pull/3457/files#diff-4a065843059602ff4759ed922279b5fcfa54c3a562b3fbafbb23b5b5ec270c46R21) to:
```
constant styles = {
          horizontal section: {
              flexible screen',
              flexWrap: 'wrap',
              alignItems: 'flexible start',
          },
};
```

Thanks!

henrikmv: Thank you, @simonadomnisoru!

simonadomnisoru (commented):
@henrikmv the code looks good now! 🎉 
There are 5 [failed tests](https://github.com/dhis2/capture-app/actions/runs/7686310565/job/20990634443?pr=3457) in the `NewPage.feature`. Can you take a look at them before I approve? Thank you!

simonadomnisoru (approved):
Good job! 👍 
PS: To get rid of `Search using zip code` failures, you can sync your branch with master after https://github.com/dhis2/capture-app/pull/3524 is merged

henrikmv:
@simonadomnisoru, I have had to make further changes. Will you please take another look? 
What I have done:
- Added an additional condition to display the container. Should be displayed only when showContainer is true and the formHorizontal property is false.
- Adjusted the style to change the alignment of elements according to the application's style requirements.

Thank you!

simonadomnisoru (approved):
LGTM!

geethaalwan (approved):

PR #3460, opened 2023-11-05
chore: [TECH-1541] Replace Material UI Dialog

eirikhaugstulen:


eirikhaugstulen:
Thanks for the review @simonadomnisoru - I had my filters set up a bit wrong, but it should be good now 👋
I think I have some failing tests though, so I'll look into those.

simonadomnisoru (approved):
Looks good 🎉 !

PR #3463, opened 2023-11-15
fix: [DHIS2-15686] fix program rule variables for edit profile mode

superskip:
[DHIS2-15686](https://dhis2.atlassian.net/browse/DHIS2-15686)

Corrects some inaccuracies in the profile widget's rule engine payload builder.

[DHIS2-15686]: https://dhis2.atlassian.net/browse/DHIS2-15686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superskip:
The latest addition to `useProgram.js` looks kind of ugly though, doesn't it? :thinking:
Do you think we should bother with trying a different approach, for instance continuing along the lines I erased from `rulesContainer.js`?

PR #3464, opened 2023-11-16
feat: [DHIS2-16123] Add inheritable TEAs to Relationships

eirikhaugstulen:
Tech summary:
- Added a hook that checks if the current trackedEntityType has any inheritable attributes and fetches the values if there are any.

### DIFF
--- a/src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/RegisterTei/DataEntry/Enrollment/dataEntryEnrollment.types.js
+++ b/src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/RegisterTei/DataEntry/Enrollment/dataEntryEnrollment.types.js
@@ -6,11 +6,14 @@ import type {
     SaveForEnrollmentAndTeiRegistration,
     ExistingUniqueValueDialogActionsComponent,
 } from '../../../../../../DataEntries';
+import type { InputAttribute } from '../../../../../../DataEntries/EnrollmentRegistrationEntry/hooks/useFormValues';
 
 export type Props = {|
     theme: Theme,
     programId: string,
     orgUnitId: string,
+    originTeiId: ?string,
+    inheritedAttributes: Array<InputAttribute>,
     enrollmentMetadata?: Enrollment,
     onSave: SaveForEnrollmentAndTeiRegistration,
     duplicatesReviewPageSize: number,
--- a/src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/useInheritedAttributeValues.js
+++ b/src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/useInheritedAttributeValues.js
@@ -1,0 +1,47 @@
+// @flow
+import type { InputAttribute } from '../../../DataEntries/EnrollmentRegistrationEntry/hooks/useFormValues';
+import { useApiDataQuery } from '../../../../utils/reactQueryHelpers';
+import { getTrackedEntityTypeThrowIfNotFound } from '../../../../metaData';
+
+type Props = {
+    teiId: string,
+    trackedEntityTypeId: string,
+};
+
+type Return = {
+    inheritedAttributes: Array<InputAttribute>,
+    originTeiId: ?string,
+    isLoading: boolean,
+};
+export const useInheritedAttributeValues = ({ teiId, trackedEntityTypeId }: Props): Return => {
+    const trackedEntityType = getTrackedEntityTypeThrowIfNotFound(trackedEntityTypeId);
+    const inheritedAttributeIds = trackedEntityType.attributes.reduce((acc, attribute) => {
+        if (attribute.inherit) {
+            acc.push(attribute.id);
+        }
+        return acc;
+    }, []);
+
+    const { data, isLoading } = useApiDataQuery(
+        ['inheritedAttributeValues', teiId],
+        {
+            resource: 'tracker/trackedEntities',
+            id: teiId,
+            params: {
+                fields: ['attributes'],
+            },
+        }, {
+            enabled: !!teiId,
+            select: (response) => {
+                const attributes = response.attributes || [];
+                return attributes
+                    .filter(attribute => inheritedAttributeIds.includes(attribute.attribute));
+            },
+        });
+
+    return {
+        inheritedAttributes: data ?? [],
+        originTeiId: null,
+        isLoading,
+    };
+};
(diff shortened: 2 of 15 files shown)


src/core_modules/capture-core/components/DataEntries/TeiRegistrationEntry/hooks/useFormValuesFromSearchTerms.js
```diff
@@ -1,19 +1,29 @@
...
-    const [formValues, setFormValues] = useState();
-    useEffect(() => {
+
+    return useMemo(() => {
+        if (inheritedAttributes) {
+            return inheritedAttributes
+                ?.reduce((acc, item) => ({
+                    ...acc,
+                    [item.attribute]: convertClientToForm(item.value, item.valueType),
+                }), {});
```
superskip, line 20: I'm not really a fan of using `...acc` when building an object in `reduce`, because it creates an unnecessary copy of `acc` for each element in the array (and the copying gets slower as `acc` gains size). Same thing just below.

src/core_modules/capture-core/components/DataEntries/TeiRegistrationEntry/hooks/useFormValuesFromSearchTerms.js
```diff
@@ -1,19 +1,29 @@
...
+    inheritedAttributes: ?Array<InputAttribute>,
+};
 
-export const useFormValuesFromSearchTerms = () => {
+export const useFormValuesFromSearchTerms = ({ inheritedAttributes }: Props) => {
     const searchTerms = useSelector(({ searchDomain }) => searchDomain.currentSearchInfo.currentSearchTerms);
-    const [formValues, setFormValues] = useState();
-    useEffect(() => {
+
+    return useMemo(() => {
```
superskip, line 14: :+1:

superskip (commented):
Review service time! :smile:

The big thing I noticed is that there is still missing inheritance from non-TET attributes.
In addition you'll find that I've added the usual assorted code comments.

src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/useInheritedAttributeValues.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+type Return = {
+    inheritedAttributes: Array<InputAttribute>,
+    originTeiId: ?string,
+    isLoading: boolean,
+};
+export const useInheritedAttributeValues = ({ teiId, trackedEntityTypeId }: Props): Return => {
+    const trackedEntityType = getTrackedEntityTypeThrowIfNotFound(trackedEntityTypeId);
+    const inheritedAttributeIds = trackedEntityType.attributes.reduce((acc, attribute) => {
```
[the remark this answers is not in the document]
eirikhaugstulen: Don't think we will ever reach a big enough amount of TEAs to tell a difference, but I added it either way 👍 Thanks!

src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/useInheritedAttributeValues.js
```diff
@@ -0,0 +1,47 @@
...
+            select: (response) => {
+                const attributes = response.attributes || [];
+                return attributes
+                    .filter(attribute => inheritedAttributeIds.includes(attribute.attribute));
+            },
+        });
+
+    return {
+        inheritedAttributes: data ?? [],
+        originTeiId: null,
```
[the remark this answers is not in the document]
eirikhaugstulen: Haha, sorry! Some debugging code that was left behind. Cleaned it up.

src/core_modules/capture-core/components/DataEntries/TeiRegistrationEntry/hooks/useFormValuesFromSearchTerms.js
eirikhaugstulen: Same here, I don't think we will reach a point were this code causes performance issues, but you're not wrong either! Refactored to only use the same object 👍

eirikhaugstulen:
> Review service time! 😄
> 
> The big thing I noticed is that there is still missing inheritance from non-TET attributes. In addition you'll find that I've added the usual assorted code comments below.

Thanks for the review @superskip - fixed all of the things mentioned above!
I was not aware that you could have inheritance on Data Elements 🤔 Can you give me a tip on where to configure this?

src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/useInheritedAttributeValues.js
```diff
@@ -10,17 +10,15 @@ type Props = {
...
-        if (attribute.inherit) {
-            acc.push(attribute.id);
-        }
-        return acc;
-    }, []);
+    const inheritedAttributeIds = new Set<string>(
+        trackedEntityType.attributes
+            .filter(attribute => attribute.inherit)
+            .map(attribute => attribute.id),
+    );
```
superskip, line 21: Hmm, two passes using `filter` and `map`, and then surely there must be a third one in the `Set` constructor.. I think we should keep reduce but change the initial value from `[]` to `new Set` :grin:

src/core_modules/capture-core/components/Pages/common/TEIRelationshipsWidget/RegisterTei/DataEntry/Enrollment/dataEntryEnrollment.types.js
```diff
@@ -6,11 +6,14 @@ import type {
     SaveForEnrollmentAndTeiRegistration,
     ExistingUniqueValueDialogActionsComponent,
 } from '../../../../../../DataEntries';
+import type { InputAttribute } from '../../../../../../DataEntries/EnrollmentRegistrationEntry/hooks/useFormValues';
 
 export type Props = {|
     theme: Theme,
     programId: string,
     orgUnitId: string,
+    originTeiId: ?string,
```
superskip, line 15: Missed a spot :eyes:

superskip (commented):
@eirikhaugstulen Good job getting throught the comments so quicky :slightly_smiling_face:

Clarification: TEAs not associated with a TET (for example phone number in TB program).

eirikhaugstulen:
All right, @superskip - now I get what you mean. Should be better now!

superskip (approved):
LGTM :tada:

geethaalwan (approved):
Tested successfully on 2.41,2.40.3,2.39.4,2.38.6 versions

PR #3474, opened 2023-11-28
feat: [DHIS2-15783] Tooltip on long working list names

eirikhaugstulen:
Summary:
- Added a tooltip for TemplateSelectorChip
- Wrapped it in a button to get keyboard focusability (not supported by UI)


simonadomnisoru (commented):
Hey @eirikhaugstulen,
Can you add the same logic on the `/search` page inside the [TemplateSelector component](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/TemplateSelector/TemplateSelectorChip.component.js)?
Thanks!

src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/TemplateSelectorChip.component.js
```diff
@@ -69,9 +61,11 @@ export const TemplateSelectorChipPlain = (props: Props) => {
                     onBlur={onMouseOut}
                 >
                     <Chip
+                        marginTop="0"
+                        marginBottom="0"
+                        marginLeft="0"
+                        marginRight="0"
```
superskip, line 67: This looks better :+1:

superskip (approved):

simonadomnisoru (approved):

eirikhaugstulen:
Thanks for the review, @simonadomnisoru & @superskip!

Regarding conditionally rendering tooltips; there seems to be a lot of work to make this generic while using `ConditionalTooltip`. (Chips we can wrap in a separate button, while if we use DHIS2/UI buttons, things work a bit differently.)
If you check my [latest commit](https://github.com/dhis2/capture-app/pull/3474/commits/19bf816e0e4b31262d72e097baaead95263973c2), you can see how it could be implemented with a separate `ConditionalTooltipForChip`-component. Could you take a look and see if this is something you want?

src/core_modules/capture-core/components/WorkingLists/WorkingListsBase/TemplateSelectorChip.component.js
```diff
@@ -27,16 +28,28 @@ export const TemplateSelectorChip = (props: Props) => {
...
+                marginLeft="0"
+                marginRight="0"
+                dataTest="workinglist-template-selector-chip"
+                selected={id === currentTemplateId}
+            >
+                <TemplateSelectorChipContent
+                    {...passOnProps}
+                    text={name}
+                    isSelectedTemplate={id === currentTemplateId}
+                />
```
superskip, line 50: We should introduce a length limit prop in `TemplateSelectorChipContent`.

superskip (commented):
Hey @eirikhaugstulen!
I think extracting the style- and ref-stuff into a separate component made the code look better :smile:
I'm wondering if we should put the two conditional tooltip components in a common `Tooltip` folder. I also suggest renaming the new tooltip component to `TooltipForChip` (no need to make clear in the title that such tooltips can be disabled).

superskip:
Very good @eirikhaugstulen! I only have one final minor refactor suggestion, which is to add an index file to the Tooltips folder. Please say no if you think it's too pointless.

superskip (approved):
After our discussion I agree with you it's fine as it is now, never mind last comment :slightly_smiling_face:

simonadomnisoru (approved):
> Regarding conditionally rendering tooltips; there seems to be a lot of work to make this generic while using `ConditionalTooltip`. (Chips we can wrap in a separate button, while if we use DHIS2/UI buttons, things work a bit differently.) If you check my [latest commit](https://github.com/dhis2/capture-app/pull/3474/commits/19bf816e0e4b31262d72e097baaead95263973c2), you can see how it could be implemented with a separate `ConditionalTooltipForChip`-component. Could you take a look and see if this is something you want?

The latest commits look great! Good job @eirikhaugstulen 👏

geethaalwan (approved):

PR #3480, opened 2023-12-06
feat: [DHIS2-12544] Add verbose logging to rules engine

eirikhaugstulen:
Tech-summary:
- Added an option to enable `verbose=true` in the url
- Pass a `onVerboseLog` callback to the rules engine
- Add optional flags to the `RulesEngine.js`-class


eirikhaugstulen:
Sure thing, @simonadomnisoru! Thanks for pointing that out 👍

simonadomnisoru (approved):

src/core_modules/capture-core/rules/useRuleEngineFlags.js
```diff
@@ -0,0 +1,23 @@
...
+        rulesEngine.setFlags({ ...rulesEngine.getFlags(), ...flags });
+    };
+
+    useEffect(() => {
+        if (verbose === 'true') {
+            updateFlags({ verbose: true });
+        } else {
+            updateFlags({ verbose: false });
+        }
+    }, [verbose]);
```
JoakimSM, line 22: With this approach we can in theory run the rules engine before the verbose flag has been set, because an `useEffect` in an inner component will run before this one (in practice it will probably work because we use async functions). Did you put this in an useEffect because side-effects should not be part of the rendering, or because some part of react-router-dom is not fully initialised in the render? The preferred solution here might be to use `useLayoutEffect` (as that one is fired before `useLayout`)

eirikhaugstulen: No special thoughts behind it, just choosing useEffect over useLayoutEffect by default. Switched now 👍

geethaalwan (approved):

JoakimSM (approved):
🎉 Nice!

PR #3482, opened 2023-12-08
fix: [DHIS2-15888] multi text element in program rules

simonadomnisoru:
[DHIS2-15888](https://dhis2.atlassian.net/browse/DHIS2-15888)

**Tech summary**
- Right now, it is not possible to check specific multi-text values. [DHIS2-16211](https://dhis2.atlassian.net/browse/DHIS2-16211) and [DHIS2-13436](https://dhis2.atlassian.net/browse/DHIS2-13436) must be implemented first.
- I added logic to check that a multi-text field has a value, therefore rules like the one mentioned in the ticket description `A{AllergiesMulti} != ""` should work correctly.
- The same rule actions that can be applied to an element with a single value option can also be applied to an element with multiple value options (Hide, Hide option, Show error, Show warning, Make filed mandatory, Assign a value, etc.).
- To assign multiple values, the expression to be evaluated and assigned must contain valid options codes separated by `,` (for example, a valid expression for the rule mentioned in the ticket is `'Penicillin,Other,Anticonvulsants'`)
- added some unit tests for the new logic

[DHIS2-15888]: https://dhis2.atlassian.net/browse/DHIS2-15888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-16211]: https://dhis2.atlassian.net/browse/DHIS2-16211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-13436]: https://dhis2.atlassian.net/browse/DHIS2-13436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/rules/postProcessRulesEffects.js
+++ b/src/core_modules/capture-core/rules/postProcessRulesEffects.js
@@ -3,15 +3,15 @@ import { effectActions } from '@dhis2/rules-engine-javascript';
 import type { OutputEffect, HideOutputEffect, AssignOutputEffect, OutputEffects } from '@dhis2/rules-engine-javascript';
 import { type RenderFoundation, dataElementTypes } from '../metaData';
 
-const isValidOptionSet = (optionSets, effectValue) => {
-    if (!optionSets.options || effectValue === null || effectValue === '') {
+const isValidOptionSet = (optionSet, effectValue) => {
+    if (!optionSet.options || effectValue === null || effectValue === '') {
         return false;
     }
     // Using == because effect.value is always a string whereas option.value can be a number
-    if (optionSets.dataElementType === dataElementTypes.MULTI_TEXT) {
-        return effectValue.split(',').every(value => optionSets.options.some(option => option.value == value));
+    if (optionSet.dataElementType === dataElementTypes.MULTI_TEXT) {
+        return effectValue.split(',').every(value => optionSet.options.some(option => option.value == value));
     }
-    return optionSets.options.some(option => option.value == effectValue);
+    return optionSet.options.some(option => option.value == effectValue);
 };
 
 const getAssignEffectsBasedOnHideField = (hideEffects: Array<HideOutputEffect>) =>
--- a/src/core_modules/capture-core/rules/postProcessRulesEffects.js
+++ b/src/core_modules/capture-core/rules/postProcessRulesEffects.js
@@ -4,7 +4,7 @@ import type { OutputEffect, HideOutputEffect, AssignOutputEffect, OutputEffects 
 import { type RenderFoundation, dataElementTypes } from '../metaData';
 
 const isValidOptionSet = (optionSets, effectValue) => {
-    if (!optionSets || !optionSets.options || effectValue === null || effectValue === '') {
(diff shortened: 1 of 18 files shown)


src/core_modules/capture-core/rules/postProcessRulesEffects.js
```diff
@@ -27,14 +38,13 @@ const postProcessAssignEffects = ({
...
-        acc[id] = optionSet.options;
+        acc[id] = { options: optionSet.options, dataElementType: optionSet.dataElement.type };
         return acc;
     }, {});
 
     // If a value gets assigned to an option set it must match one of its available options
     assignValueEffects.map((effect) => {
-        // Using == because effect.value is always a string whereas option.value can be a number
-        if (optionSets[effect.id] && !optionSets[effect.id].some(option => option.value == effect.value)) {
+        if (!isValidOptionSet(optionSets[effect.id], effect.value)) {
```
superskip, line 47: We should check that `effect.id` is an option set before calling `isValidOptionSet`:

`if (optionSets[effect.id] && !isValidOptionSet(optionSets[effect.id], effect.value))`

Otherwise the example program rule in the jira blanks the first name field instead of assigning 'Test'.

superskip (commented):
Made three comments :eyes: 
I'm trusting the tests are alright :slightly_smiling_face:

src/core_modules/capture-core/rules/postProcessRulesEffects.js
```diff
@@ -1,7 +1,18 @@
...
-import type { RenderFoundation } from '.././metaData';
+import { type RenderFoundation, dataElementTypes } from '../metaData';
+
+const isValidOptionSet = (optionSets, effectValue) => {
+    if (!optionSets || !optionSets.options || effectValue === null || effectValue === '') {
+        return false;
+    }
+    // Using == because effect.value is always a string whereas option.value can be a number
+    if (optionSets.dataElementType === dataElementTypes.MULTI_TEXT) {
+        return effectValue.split(',').every(value => optionSets.options.some(option => option.value == value));
```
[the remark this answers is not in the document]
simonadomnisoru: Yes, we need to make sure there are no duplicate entries. I added logic to do it and a unit test for this case.

src/core_modules/capture-core/rules/postProcessRulesEffects.js
simonadomnisoru: Good catch! I updated the code with the appropriate checks. Thanks!

simonadomnisoru:
> Made three comments 👀

@superskip I updated the code with your suggestions. Can you have a look? Thanks!

superskip (approved):

PR #3489, opened 2023-12-19
fix: [TECH-1676] wait for url to update before reloading page

superskip:
I found the hardcoded template id in another test.


eirikhaugstulen (approved):
👍

PR #3490, opened 2023-12-20
chore: fix cypress script command

JoakimSM:
The `cy:open` and `cy:run` commands are not working in my dev environment (I have to run in different terminal windows and make sure to export the NODE_OPTIONS env variable). This PR aims to fix that.

Great if you can verify that `yarn cy:open`and `yarn cy:run` are working in your dev environment too (with this fix)


superskip:
Got this error on ubuntu after running `yarn run cy:open` and attempting to run a test suite in electron:

> We stopped running your tests because your config file crashed.

superskip (approved):
Both commands seems to work :+1:

simonadomnisoru (approved):
Both commands work in my local environment 🥳

PR #3491, opened 2023-12-20
fix: [DHIS2-16016] handle multi-text in v40

JoakimSM:
`MULTI_TEXT` is not fully supported in version 40 and you are not able to save an event / tracked entity with multiple values in such an attribute / data element. This PR "fixes" this by only allowing you make a single selection for this value type in version 40.

https://dhis2.atlassian.net/browse/DHIS2-16016

I also cleaned up some code related to `MULTI_TEXT` without option set (this is not supported by the app)

Note: Some of the files are put in the `metadataMemoryStoreBuilders` folder, this might seem weird. But there is a plan to do some reorganising that we will get to in the future (and this will make more sense afterwards)


src/core_modules/capture-core/metaDataMemoryStoreBuilders/common/helpers/dataElement/unsupportedMultiText/handleMultiTextPreV41.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import log from 'loglevel';
+import { errorCreator, hasAPISupportForFeature, FEATURES } from 'capture-core-utils';
+import { dataElementTypes, DataElement } from '../../../../../metaData';
+
+const isMultiTextPreV41 = (type, minorServerVersion: number) =>
+    type === dataElementTypes.MULTI_TEXT && !hasAPISupportForFeature(minorServerVersion, FEATURES.multiText);
+
+export const handleMultiTextPreV41 = (dataElement: DataElement, minorServerVersion: number) => {
```
simonadomnisoru, line 9: NIT: remove `PreV41` from the functions name and use a non-specific spelling instead. Perhaps something like `handleMultiTextFallback` and `isMultiTextSupported` can work.

simonadomnisoru (approved):
LGTM!

src/core_modules/capture-core/metaDataMemoryStoreBuilders/common/helpers/dataElement/unsupportedMultiText/handleMultiTextPreV41.js
JoakimSM: I agree the suffix (PreV41) wasn't optimal. Really hard to find a descriptive name though. Going for `handleMultiTextInUnsupportedBackendVersions` and `isMultiTextInUnsupportedBackendVersion`. Long, but at least you get a sense of what it is about.

PR #3498, opened 2024-01-03
feat: [DHIS2-12007] Text String Improvements

henrikmv:
Improvements of text strings implemented:
- Feedbacks --> Feedback

After discussion with Karoline, the following is to be kept as original (for now)
- Or see all records accessible to you in program  --> Or see all `TET` accessible to you in `program`

Text string removed from master
- There are no feedbacks for this event --> There is no feedback for this event


simonadomnisoru (approved):

geethaalwan (approved):

PR #3517, opened 2024-01-29
fix: [DHIS2-16499] Handle different API response format

simonadomnisoru:
[DHIS2-16499](https://dhis2.atlassian.net/browse/DHIS2-16499)

**Tech summary**
- Added logic to [handle the different API response formats](https://github.com/dhis2/capture-app/pull/3517/files#diff-6dfa375cccf2a88fde0b6cf7028c3c9837badf6c955768a32a0163e873166d93R3). To keep backward compatibility with API versions below 41, the handler will default to `instances` when the requested entities field is not found. 
- Adapted the cypress scenarios to handle different responses shape
- Added the [exportablePayload](https://github.com/dhis2/capture-app/pull/3517/files#diff-d0a7801c93460afbf50c04c844295e59d12e96c64cd3d3761623d0f6e5c9ce84R14) key in the features support list (used when computing the [relathionship queryKey](https://github.com/dhis2/capture-app/pull/3517/files#diff-1cf8415853cb86dd013a26f121aa7a9d0f4097e94f24cefec177b9051d286347R24))

[DHIS2-16499]: https://dhis2.atlassian.net/browse/DHIS2-16499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3520, opened 2024-01-30
feat: [DHIS2-16372] Delete Relationships

eirikhaugstulen:
Tech-summary:
- Added a IconDelete to each row in the TEI Relationships widget
- Show confirmation modal asking the user if they are sure
- Optimistically delete the row when clicked

Some caveats:
- Needs update when [3517](https://github.com/dhis2/capture-app/pull/3517) is merged
- Soft-deleted relationships are still returned in the new tracker endpoints. Issue raised in `tracker-agile-team`


src/core_modules/capture-core/components/WidgetsRelationship/common/RelationshipsWidget/DeleteRelationship/DeleteRelationship.js
```diff
@@ -0,0 +1,82 @@
...
+        zIndex: 1000,
+    },
+};
+
+export const DeleteRelationshipPlain = ({ handleDeleteRelationship, disabled, classes }: Props) => {
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    return (
+        <>
+            <DataTableCell className={classes.tableCell}>
+                <IconButton
```
simonadomnisoru, line 39: Does it make sense to apply the destructive red color to the delete icon? Some suggestion:

Screenshot 2024-01-31 at 14 56 58

simonadomnisoru (commented):
@eirikhaugstulen, can you add a Cypress test for the delete relationship functionality? Thank you!

eirikhaugstulen:
Hey @simonadomnisoru!
Thanks for the review. I have implemented and written some tests. (These are commented out for now as there is a bug in the API - will fix before merge)

cypress/e2e/WidgetsForEnrollmentPages/WidgetTeiRelationship/index.js
```diff
@@ -0,0 +1,146 @@
...
+    cy.get('[data-test="delete-relationship-modal"]')
+        .within(() => {
+            cy.get('[data-test="delete-relationship-confirmation-button"]')
+                .click();
+        });
+
+    cy.wait('@deleteRelationship')
+        .its('request.body')
+        .should('deep.equal', {
+            relationships: [{ relationship: 'mwswG3RMNuu' }],
```
simonadomnisoru, line 136: **Reminder** after the bug in the API is fixed: To keep DB consistency, the relationship that was deleted should be added back. Otherwise, the `there is an existing relationship` step will fail the next time the scenario is run.

simonadomnisoru (approved):
Added a reminder to look into before merging, other than that it looks great! 🎉

cypress/e2e/WidgetsForEnrollmentPages/WidgetTeiRelationship/index.js
eirikhaugstulen: This data is actually just mocked, so no need to add it again. We intercept the request for relationships in the before hook and then assert that we send the actual call to the api with the correct data! WDYT?

geethaalwan (approved):
Tested successfully on 2.41,2.40.3,2.39.5,2.38.7 versions

cypress/e2e/WidgetsForEnrollmentPages/WidgetTeiRelationship/index.js
simonadomnisoru: Testing against mocked data versus the API and DB has advantages and disadvantages 🤔 . By using mock data, we will no longer benefit from having the latest API changes available. We will need to ensure that the mocks are always up to date, otherwise, we may not detect when we introduce bugs. 
Let's discuss it more in-depth in our next FE check-in meeting!

PR #3521, opened 2024-02-02
feat: [DHIS2-14799] Working list for follow up

eirikhaugstulen:
TECH-summary:
- Added a filter button with type TRUE_ONLY
- Filters the API by followUp status

The design doc also mentions that this should be available to select in the ColumnSelector, but I'm not sure if this makes sense. FollowUp is an attribute of the enrollment itself, meaning we have no good way of displaying this. Not sure what you think.


simonadomnisoru:
@eirikhaugstulen IMHO this feature needs more design clarification. Some doubts that came to mind:

- Is `Follow up` one on the main filter and it is always visible just like the `Enrollment status` or is one of the secondary ones hidden under "More filters"?
- Clarify if `Follow up` will appear in the ColumnSelector and in the table (for program stage WL we already combine events and TEIs metadata in the ColumnSelector and the table, therefore it should also be possible to add enrollment metadata too)

Thanks!

eirikhaugstulen:
Hey @simonadomnisoru! Thanks for the review. I had some discussions with Joakim and Karoline, and the consensus was that we put in `More filters`, but we don't add it as a column 😊 I added a flag on `useFiltersOnly` that would add the filter in the correct spot. Thank you!

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/Setup/hooks/useFiltersOnly.js
```diff
@@ -64,6 +64,19 @@ export const useFiltersOnly = ({
                     },
                 ]
                 : []),
+            {
+                id: MAIN_FILTERS.FOLLOW_UP,
+                type: dataElementTypes.BOOLEAN,
```
simonadomnisoru, line 69: Hey @eirikhaugstulen ,
Based on the ticket description, the filter should follow the behavior of a "Yes Only" value type. You can achieve this by using `dataElementTypes.TRUE_ONLY` instead.
`options` can be removed and the logic in`transformRecordsFilter` and the `converter` should be adapted for `dataElementTypes.TRUE_ONLY`.
Thanks!

eirikhaugstulen: Hey! During the discussion we had on the design, we decided to use a boolean instead. This is because we also need the functionality of filtering by enrollments who are not marked for follow up. I have updated the Jira-issue. Thank you!

simonadomnisoru: Looks like the API returns a `400` error when selecting both `Yes` and `No`. Is this a bug in the API or should the UI not allow multiple values to be selected for the `followUp` parameter? WDYT?

PS: please also update the detailed design docs with the correct type of filter. Thanks!

https://github.com/dhis2/capture-app/assets/6817073/131a7366-66aa-4fb0-ac91-d19c980735c2

eirikhaugstulen: Sorry, I was testing with ´multiValueFilter´ to see if the functionality worked and then I forgot to remove it. It should only be possible to select Yes _or_ No!

simonadomnisoru (approved):
LGTM!

JoakimSM (commented):
Couple of comments that might or might not be in the back of your head:
- It seems like boolean value type filters have changed from a multi-select to a single-select. Great if we can keep multi-select for the value type.
- Boolean value type filters are stored as a string representation in Redux when you add the filter (`workingListsMeta/teiList/filters/`). When the filter is loaded from a template, it is stored as a javascript boolean primitive. Great if we can always save them as javascript booleans. The idea was to store the "client-formatted" value here, and it would be great to continue doing so (boolean for boolean value type, number for number/integer value type, ISO-string for date etc. See the converters formToClient, clientToServer etc.)

The follow-up part seems to work well with an updated backend 👍

JoakimSM (approved):

PR #3522, opened 2024-02-02
fix: [DHIS2-16318] enrollment page url management

superskip:
[DHIS2-16318](https://dhis2.atlassian.net/browse/DHIS2-16318)

This is a complete rewrite of the enrollment page epics. Their primary function is (still) to handle changes in the url.

The url of the enrollment page can be navigated by tweaking four parameters: `enrollmentId`, `programId`, `teiId` and `orgUnitId`. Their effect can briefly be described as follows.
- A change in `enrollmentId` triggers an update of the `teiId` and `programId`. The new id's are those associated with the specified enrollment. Any existing values of `teiId` and `programId` in the url will be overwritten.
- A change in `teiId` triggers downloading of tracked entity type and tei display name.
- A change in `teiId` or `programId` triggers enrollment data to be downloaded **and** causes automatic switching of selected orgUnit to the associated program owner, provided the user has capture scope access to this orgUnit.
- Changes in `orgUnitId` triggers no new actions (hence no epics for this).

There is some additional complexity related to management of the status of the few asynchronous calls that is going on in the epics. For instance, if a new `teiId` and `programId` is being fetched from the server due to updated `enrollmentId`, we will in the meantime avoid initiating any async request depending on `teiId` and `programId`.

Further complexity comes from error handling and dealing with blanking of ids.

[DHIS2-16318]: https://dhis2.atlassian.net/browse/DHIS2-16318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
```diff
@@ -39,172 +59,287 @@ const enrollmentsQuery = (teiId, programId) => ({
...
+        filter(({ payload }) => payload.enrollmentId === store.value.enrollmentPage.enrollmentId),
+        map(({ payload }) => fetchEnrollmentIdSuccess(payload)),
+    );
+
+export const enrollmentIdErrorEpic = (action$: InputObservable) =>
+    action$.pipe(
+        ofType(enrollmentPageActionTypes.FETCH_ENROLLMENT_ID_ERROR),
+        map(({ payload: { enrollmentId } }) =>
+            showErrorViewOnEnrollmentPage({
+                error: i18n.t('Invalid enrollment id {{enrollmentId}}.', {
```
simonadomnisoru, line 149: Hi, can we use the same wording as in `fetchTeiErrorEpic` and `programIdErrorEpic` (e.g. `i18n.t('Enrollment with id "{{enrollmentId}}" does not exist', {enrollmentId })`?

I also noticed that when `orgUnitId` does not exist, the red error message is not displayed and a loading spinner is displayed instead. Is it possible to employ the same error message pattern used by the other URL parameters?

Screenshot 2024-02-09 at 14 26 26

Thanks!

simonadomnisoru (commented):
Is it possible to add a Cypress test that checks that the automatic switching of the owner orgUnit is working correctly?

superskip:
Thanks for the review @simonadomnisoru!
- Changed the enrollment id error message
- Improved the handling of invalid `orgUnitId`. I didn't do the error screen exactly like how the other three url parameters do it: in the case of an invalid `orgUnitId` the scope selector is displayed. This is probably better than not showing it from a user perspective.
- Added a small test checking automatic switching of orgunits, but I think it does the trick :+1:

src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.epics.js
superskip: The handling of invalid orgUnitId implemented below doesn't look good enough on review; the scope selector shows the last selected orgunit, so to avoid confusion the error message should somehow indicate that it's the value in the url that is the problem.
Suggestion: `i18n.t('Organisation unit with id "{{orgUnitId}}" could not be retrieved', { orgUnitId })`.

simonadomnisoru: I agree that this is a bit confusing. Can you investigate if is possible to remove the last selected org unit from the scope selector?

- home page invalid `orgUnitId` handling

Screenshot 2024-02-15 at 09 30 19


- new page invalid `orgUnitId` handling. The ScopeSelector is not displayed at all on this page. Perhaps this could also be an option for the enrollment page too?


Screenshot 2024-02-15 at 09 31 45

WDYT? Thanks!

superskip: > Can you investigate if is possible to remove the last selected org unit from the scope selector?

I went with this solution.

simonadomnisoru (approved):

geethaalwan (approved):

PR #3546, opened 2024-02-28
feat: [DHIS2-15391] preview images in working lists and on enrollment dashboard

superskip:
Feature summary:
- TEI working list can show image attributes
- TEI working list filtered by program stage can show image data elements
- Event program working list can show image data elements
- Enrollment dashboard can show image data elements marked with "Display in reports"


src/core_modules/capture-ui/PreviewImage/PreviewImage.component.js
```diff
@@ -0,0 +1,66 @@
...
+        display: 'inline-block',
+        '&:hover $image': {
+            opacity: 0.5,
+        },
+        '&:hover $icon': {
+            visibility: 'visible',
+        },
+    },
+    image: {
+        height: 80,
```
simonadomnisoru, line 17: To make sure it doesn't exceed the maximum size, can you use `maxHeight` instead? Thanks!

src/core_modules/capture-ui/PreviewImage/PreviewImage.component.js
```diff
@@ -0,0 +1,66 @@
...
+        image: string,
+        icon: string,
+    },
+}) => {
+    const { url, classes } = props;
+
+    return (
+        <div className={classes.container}>
+            <a
+                href={url}
```
simonadomnisoru, line 46: Hey, clicking on the image should open the full image in a new tab. To achieve this, when creating the link `href`, you can remove `dimension=small` from this URL. WDTY?

src/components/AppLoader/AppLoader.component.js
```diff
@@ -43,6 +44,7 @@ export const AppLoader = (props: Props) => {
 
     const load = useCallback(async () => {
         try {
+            initFeatureAvailability(serverVersion);
```
simonadomnisoru, line 47: Hey, 
What is the use-case for which this initialization is needed in the `AppLoader`? You can find the `minorServerVersion` inside of the metadata factories. In the epics, the `serverVersion` is available in the redux store and within components by using the `useConfig()` app-runtime hook. 
Thank you!

src/core_modules/capture-ui/PreviewImage/PreviewImage.component.js
superskip: We get a fixed height for the image by using `height`, so 80 pixels won't be exceeded. On the other hand if we switch to `maxHeight` we might end up with an image lower than 80 pixels.

src/components/AppLoader/AppLoader.component.js
superskip: Thanks for bringing this up! :slightly_smiling_face: It turned out that all the places where I needed to check the server version was neither components nor hooks, but regular javascript functions. I couldn't see a convenient way to do this with the existing tools, so I made a new one. There isn't anything clever about how I did it, and I'm open to moving things around a little if it would make it feel more integrated.

src/core_modules/capture-ui/PreviewImage/PreviewImage.component.js
JoakimSM: Talked to @superskip, he will make sure the image-container is always 120x80 (other elements should not move around when the image is loaded) etc.

superskip:
- Removed `dimension=small` from the link url
- Fixed the image width as well, making it centralised in a 120x80 frame.

simonadomnisoru (approved):
Great job! 👏

src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getTeiListData/convertToClientTeis.js
```diff
@@ -29,7 +30,9 @@ export const convertToClientTeis = (
                     return {
                         id,
                         value: convertServerToClient(value, type),
-                        urlPath: `/trackedEntityInstances/${tei.trackedEntity}/${id}/image`,
+                        urlPath: featureAvailable(FEATURES.trackerImageEndpoint) ?
+                            `/tracker/trackedEntities/${tei.trackedEntity}/attributes/${id}/image?program=${programId}&dimension=small` :
+                            `/trackedEntityInstances/${tei.trackedEntity}/${id}/image`,
```
JoakimSM, line 35: Can we only add the urlPath if it is an image? Would improve readability quite a bit.

Also, in https://github.com/dhis2/capture-app/blob/c4b29c4b2c15195d1193636e8ca9d72536b53b9f/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/helpers/getListDataCommon/getSubvalues.js#L55 **and others**, can we have a `previewUrl` and `fullSizedUrl` instead of the single `url` and remove the replace statement in `PreviewImage.component`. Would be more robust that way and the "UI-library" will not be responsible for computing URLs.

geethaalwan (approved):
Tested successfully on 2.41 version

JoakimSM (approved):

JoakimSM:
Tested successfully on 2.41 version

PR #3559, opened 2024-03-07
fix: change translations key

JoakimSM:
minor fix to change a translations key (to something that already exists)


eirikhaugstulen (approved):
LGTM!

PR #3571, opened 2024-03-14
fix: [DHIS2-17072] Improve verbose rules engine message

JoakimSM:
Improve program rules verbose logging for feature parity.

- Use console.log for both rules expressions and action expressions.

- Print the actual evaluation result.

- Change errors from log.warn to log.error


geethaalwan (approved):
Tested successfully on 2.41,2.40.4,2.39.6,2.38.7 versions

PR #3572, opened 2024-03-15
fix: [DHIS2-17035] changelog prioritised bugs

simonadomnisoru:


src/core_modules/capture-core/components/WidgetProfile/DataEntry/FormFoundation/DataElement.js
```diff
@@ -206,6 +206,7 @@ const buildOptionSet = async (
         const icon = buildIcon(optionSetOption.style);
         return new Option((option) => {
             option.id = optionSetOption.id;
+            option.code = optionSetOption.code;
```
JoakimSM, line 209: Is there another way we can solve this? The code is currently in the value prop, would be great to not duplicate it and just keep the value prop.

simonadomnisoru: I did it this way to match the same build logic that currently exists in the [OptionSetFactory](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/metaDataMemoryStoreBuilders/common/factory/optionSet/OptionSetFactory.js#L75). I am not sure if this a bug or we intentionally store the `code` inside `value` when creating the metadata? 
It can be done in another way which I implemented in the last commit. Let me know if it's better now. Thanks!

JoakimSM: Ah, I see the confusion. The original Option class only had `value` and `text` (this is less confusing to me than `code` and `name`, at least in frontend terms). I see that `code` has been added at some point, we should probably remove this, but not in this PR (Let's go with the current code changes)

JoakimSM (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41 versions

PR #3574, opened 2024-03-16
fix: [DHIS2-16767] Refactor widgets empty text to be dynamic

eirikhaugstulen:
TECH Summary:
- Change static empty text to be sent in as a prop from page component

### DIFF
--- a/src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/LayoutComponentConfig/LayoutComponentConfig.js
+++ b/src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/LayoutComponentConfig/LayoutComponentConfig.js
@@ -1,5 +1,4 @@
 // @flow
-import i18n from '@dhis2/d2-i18n';
 import { WidgetStagesAndEvents } from '../../../../../WidgetStagesAndEvents';
 import type { Props as StagesAndEventProps } from '../../../../../WidgetStagesAndEvents/stagesAndEvents.types';
 import { TrackedEntityRelationshipsWrapper } from '../../../TEIRelationshipsWidget/TrackedEntityRelationshipsWrapper';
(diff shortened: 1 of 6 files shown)


src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/LayoutComponentConfig/LayoutComponentConfig.js
```diff
@@ -94,18 +93,18 @@ export const WarningWidget: WidgetConfig = {
 export const FeedbackWidget: WidgetConfig = {
     Component: WidgetFeedback,
     shouldHideWidget: ({ hideWidgets }) => hideWidgets?.feedback,
-    getProps: ({ widgetEffects }): WidgetFeedbackProps => ({
+    getProps: ({ widgetEffects, feedbackEmptyText }) => ({
```
simonadomnisoru, line 96: What is the reason you removed the `WidgetFeedbackProps` type? The flow checks seem to work fine with the type 🤔

eirikhaugstulen: Contradictory as one might think, I removed it because it worked 🤯 It's trying to get `feedbackEmptyText` from a type that should only accept `{ widgetEffects: ..., emptyText: string }`, yet flow has no issues with it. In my head it should fail, but I don't always understand Flow.

simonadomnisoru: IMHO flow works correctly here, because `WidgetFeedbackProps` is the return type of the getProps function, not its input arguments. If the function had types for both the arguments and the return value, it would look like this:
```
type InputFunctionProps = {|
    widgetEffects?: Object,
    feedbackEmptyText: string,
|}
type OutputFuntionProps = {|
    feedback?: ?Array<string | FilteredText | FilteredKeyValue>,
    emptyText: string,
|}
 getProps: ({ widgetEffects, feedbackEmptyText }: InputFunctionProps): OutputFunctionProps => ({
    feedback: widgetEffects?.feedbacks,
    emptyText: feedbackEmptyText,
}),
```
In this example `OutputFunctionProps` becomes the `WidgetFeedbackProps` type.

eirikhaugstulen: That makes sense. I think we lost some type safety either way with flow, and I couldn't really get it to throw an error when changing stuff, but this is better than nothing. We can look at this more in the future, but I implemented your changes in both Feedback and Indicator. Thanks for the review! 😊

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41,2.40.4,2.39.6,2.38.7 versions

PR #3580, opened 2024-03-19
fix: [DHIS2-17084] Filter Relationship Types for write access

eirikhaugstulen:
Tech-summary:
- Filter out any relationship types the user does not have data write access to
- Do this for both cached indexed db, but also when in self-contained mode

### DIFF
--- a/src/core_modules/capture-core/components/WidgetsRelationship/common/RelationshipsWidget/useRelationshipTypes.js
+++ b/src/core_modules/capture-core/components/WidgetsRelationship/common/RelationshipsWidget/useRelationshipTypes.js
@@ -13,7 +13,7 @@ type Element = {|
 const relationshipTypesQuery = {
     resource: 'relationshipTypes',
     params: {
-        fields: 'id,displayName,fromToName,toFromName,fromConstraint[relationshipEntity,trackerDataView,trackedEntityType[id,name],program[id,name],programStage[id,name]],toConstraint[relationshipEntity,trackerDataView,trackedEntityType[id,name],program[id,name],programStage[id,name]]',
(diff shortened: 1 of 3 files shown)


simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetsRelationship/common/RelationshipsWidget/useRelationshipTypes.js
```diff
@@ -23,7 +23,8 @@ export const useRelationshipTypes = (cachedRelationshipTypes?: RelationshipTypes
         relationshipTypesQuery,
         {
             enabled: !cachedRelationshipTypes?.length,
-            select: ({ relationshipTypes }: any) => relationshipTypes,
+            select: ({ relationshipTypes }: any) => relationshipTypes
+                .filter(({ access }) => access.data.write),
```
JoakimSM, line 27: See general comment first.

A better way to do this is to add filter to the api request, like so: http://localhost:9090/api/relationshipTypes?filter=access.data.read:eq:true

JoakimSM (commented):
Doesn't this affect both viewing and adding? If I'm not mistaken, we will have to keep the readonly ones for viewing (so part 1 is to do what you basically did in this PR, but check for read access instead). When using the relationship types in the add part, we will have to filter for write access only.

eirikhaugstulen:
Good catch, Joakim! Updated 😊

simonadomnisoru (approved):

JoakimSM (approved):

geethaalwan (approved):
Tested successfully on 2.41,2.40.4,2.39.6,2.38.7  versions

PR #3581, opened 2024-03-20
fix: [DHIS2-17110] Hide delete-icon in relationships if no write access

eirikhaugstulen:
Tech-summary:
- Hide the delete icon in the relationships table if the user has no write access to the type


simonadomnisoru (approved):

JoakimSM (approved):
LGTM

PR #3594, opened 2024-04-03
fix: [DHIS2-14938] trigger error on complete

simonadomnisoru:
[DHIS2-14938](https://dhis2.atlassian.net/browse/DHIS2-14938)

**Tech summary:** 
The `completeDataEntryFieldValue` prop can take different values (i.e., `undefined`, `null`, `'true'`, `'false'`). The values come from the redux store key `dataEntriesFieldsValue` which is initialized and handled differently by different forms. I fixed the check to set the form as completing only when the prop equals `'true'`.

### DIFF
--- a/src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
+++ b/src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
@@ -235,7 +235,7 @@ const getSaveHandler = (
             } = this.props;
 
             const filteredProps = onFilterProps ? onFilterProps(passOnProps) : passOnProps;
-            this.isCompleting = !!(onIsCompleting && onIsCompleting(this.props));
+            this.isCompleting = Boolean(onIsCompleting && onIsCompleting(this.props) === 'true');
 
             return (
                 <div>
--- a/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
+++ b/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
@@ -163,6 +163,15 @@ export const saveEditedEventEpic = (action$: InputObservable, store: ReduxStore)
 export const saveEditedEventSucceededEpic = (action$: InputObservable) =>
     action$.pipe(
         ofType(actionTypes.EDIT_EVENT_DATA_ENTRY_SAVED),
+        filter((action) => {
+            const {
+                meta: { triggerAction },
+            } = action;
+            return (
+                triggerAction === enrollmentSiteActionTypes.COMMIT_ENROLLMENT_EVENT ||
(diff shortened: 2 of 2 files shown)


src/core_modules/capture-core/components/DataEntry/withSaveHandler/withSaveHandler.js
```diff
@@ -235,7 +235,7 @@ const getSaveHandler = (
             } = this.props;
 
             const filteredProps = onFilterProps ? onFilterProps(passOnProps) : passOnProps;
-            this.isCompleting = !!(onIsCompleting && onIsCompleting(this.props));
+            this.isCompleting = Boolean(onIsCompleting && onIsCompleting(this.props) === 'true');
```
JoakimSM, line 238: Can we change the edit event code instead of the generic withSaveHandler component? here: https://github.com/dhis2/capture-app/blob/dbc105106c3a2efc452fb11bdcd85a831de3821e/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js#L360

I think it makes sense to keep the return value of onIsCompleting a boolean value.

src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -163,6 +163,15 @@ export const saveEditedEventEpic = (action$: InputObservable, store: ReduxStore)
...
         ofType(actionTypes.EDIT_EVENT_DATA_ENTRY_SAVED),
+        filter((action) => {
+            const {
+                meta: { triggerAction },
+            } = action;
+            return (
+                triggerAction === enrollmentSiteActionTypes.COMMIT_ENROLLMENT_EVENT ||
+                triggerAction === enrollmentEditEventActionTypes.EVENT_SAVE_ENROLLMENT_COMPLETE_SUCCESS
+            );
+        }),
```
JoakimSM, line 174: Is this really needed? Seems like we return EMPTY on line 183. Is it good enough to do one or the other?

simonadomnisoru: On my localhoast, the `return EMPTY` statement is causing an error and the page is no longer responding when editing a single event. I remove it and filter by `triggerAction` instead. This is not an issue in play so should I just rollback the change and leave it as it is? 🤔 

Screenshot 2024-04-04 at 17 54 42

Thanks!

superskip: I think the reason `EMPTY` crashes the app is that it counts as an observable and not an action; so it works as return value for `switchMap` but not for `map`. I think using `filter` is the best solution here. I approve the last commit :+1:

JoakimSM: I see, thanks. Current PR looks good to me.

JoakimSM (approved):
LGTM

PR #3601, opened 2024-04-08
fix: [DHIS2-17161] Do not update attribute values from enrollment Widget

JoakimSM:
https://dhis2.atlassian.net/browse/DHIS2-17161

Skip attribute values in the api GET request in the enrollment Widget as it should not be needed. 
This way, we don't potentially post stale attribute values to the api.

Also, stopped retrieving enrollment.attributes for the EnrollmentDomain in the Redux store (we use EnrollmentDomain.attributeValues)


simonadomnisoru (approved):
LGTM!

geethaalwan (approved):

PR #3604, opened 2024-04-12
fix: [DHIS2-16487] Auto-select program and org-unit

eirikhaugstulen:
Tech-summary:
- Added a new initializer component that fetches programs and search orgunits
- Once both is loaded, we check if there are only one of each
- If there's only one, there's no existing search params, the user is opening the main page, and it's the initial mount of the component, we pre-select the only ones.


simonadomnisoru (commented):
Hey @eirikhaugstulen 
For this use case, does it make sense to use the program and orgUnit `SelectorBarItem` in `displayOnly` mode? The callbacks onClearSelectionClick and setOpen can also be removed. WDYT?
Thanks!

eirikhaugstulen:
Hey @simonadomnisoru - thanks for the comment! Does that mean it would not be deselectable? I think the idea here is that you should still be able to deselect and use the ScopeSelector as normal, we would just help the first-time login with preselecting if there are only one. Did I understand your comment correctly?

simonadomnisoru:
> Hey @simonadomnisoru - thanks for the comment! Does that mean it would not be deselectable? I think the idea here is that you should still be able to deselect and use the ScopeSelector as normal, we would just help the first-time login with preselecting if there are only one. Did I understand your comment correctly?

My suggestion was that only the program and orgUnit `SelectorBarItem` would not be deselectable if there are only one. The rest of the ScopeSelector selectors should work normally. IMHO a dropdown where you can only select one option (the same one you previously deselected) doesn't bring that much value to the user. Perhaps the functional design team has some feedback on the most sensible user experience for this single program/orgUnit use case? 🤔

superskip:
Disallowing deselection of program or orgunit would lead to more confusion I think? Because then the user's question becomes "why can't I select another program" rather than "why do I only have access to one program".

JoakimSM:
Reasons for being able to deselect:

- Search for all accessible events requires deselecting the org. unit
- To register TET only, we will have to deselect the program

There are probably others too (that I can't think of a.t.m.)

The behaviour has been discussed with the design team. Let's start with the suggested behaviour if there aren't any other compelling arguments to why not.

simonadomnisoru (approved):

geethaalwan (commented):
Tested successfully on 2.42,2.41.1,2.39.6,2.38.7 versions

geethaalwan (approved):

PR #3605, opened 2024-04-12
fix: [DHIS2-17052] Unrelated error message on no TETs

eirikhaugstulen:
Tech-summary:
- Add a more relevant dropdown message when there are no TETs


src/core_modules/capture-core/components/TrackedEntityTypeSelector/TrackedEntityTypeSelector.component.js
```diff
@@ -51,7 +51,11 @@ export const TrackedEntityTypeSelectorPlain =
                       label={headerText}
                       onChange={handleSelectionChange}
                       selected={selectedSearchScopeId}
-                      empty={<div className={classes.customEmpty}>Custom empty component</div>}
+                      empty={
+                          <div className={classes.customEmpty}>
+                              {i18n.t('No tracked entity types available')}
```
simonadomnisoru, line 56: I believe most end users won't understand what "tracked entity types" refer to. How about using a message like this: "You don't have access to create" instead? Thanks!

karolinelien:
I get @simonadomnisoru's point very much, but I think this is such a peripheral use case where if the user does go to that page, it is probably better that it is clear what the issue is. If we say they don't have access to create but they do have access to create in event programs for example, this could be misleading. So I think I would just go for @eirikhaugstulen's technical copy here.

simonadomnisoru (approved):

JoakimSM (approved):

geethaalwan (commented):

PR #3615, opened 2024-04-19
fix: [DHIS2-17195] implement d2:condition

superskip:
[DHIS2-17195](https://dhis2.atlassian.net/browse/DHIS2-17195)

[DHIS2-17195]: https://dhis2.atlassian.net/browse/DHIS2-17195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/rules-engine/src/d2Functions/getD2Functions.js
+++ b/packages/rules-engine/src/d2Functions/getD2Functions.js
@@ -1,8 +1,8 @@
 // @flow
-
 import log from 'loglevel';
 import { getZScoreWFA, getZScoreWFH, getZScoreHFA } from './zScoreFunctions';
 import { extractDataMatrixValue } from './gs1DataMatrixFuntions';
+import { executeExpression } from '../services/expressionService';
 import type { D2FunctionsInput, D2Functions } from './getD2Functions.types';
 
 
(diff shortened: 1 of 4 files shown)


simonadomnisoru (commented):
Hey @superskip, 
To keep the rules engine coverage the same (so we're more prepared to switch to the Kotlin implementation), can you add unit tests that cover the `d2:condition` functionality?
Thanks!

superskip:
Thanks for the review @simonadomnisoru!
I added two tests at the place I found most appropriate. However, it appears from the maintenance app that `d2:condition` is not supported in program rule expressions (only in indicator expressions), so it might not be implemented in the Kotlin version (hence there is a chance these new tests will generate unintended errors with the new rule engine).

simonadomnisoru (approved):

JoakimSM (approved):

geethaalwan (commented):
Tested successfully on 2.42,2.41.1,2.41.0,2.40.4,2.39.6,2.38.7 versions

PR #3629, opened 2024-05-02
fix: [DHIS2-15920] skip duplicates reviews

simonadomnisoru:
[DHIS2-15920](https://dhis2.atlassian.net/browse/DHIS2-15920)

**Tech summary**
- skip the duplicates check when no attributes values were filled in

[DHIS2-15920]: https://dhis2.atlassian.net/browse/DHIS2-15920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eirikhaugstulen (approved):
Looks good and works as expected! 🎉
However, before you merge; are we sure we want to be able to register TEIs with no attributes at all? It seems like it could potentially create a lot of duplicates that would be hard to track? (It's definitively an implementation error if it is though, just curious.)

simonadomnisoru:
> Looks good and works as expected! 🎉 
However, before you merge; are we sure we want to be able to register TEIs with no attributes at all? It seems like it could potentially create a lot of duplicates that would be hard to track? (It's definitively an implementation error if it is though, just curious.)

Hey @eirikhaugstulen,
Currently, it is already possible to register a TEI without attributes. This PR only removes the call to check for duplicates because all the TEIs in the program are returned and the request performance is poor.

geethaalwan (approved):

PR #3635, opened 2024-05-08
fix: [DHIS2-17054] handle empty expression value

simonadomnisoru:
[DHIS2-17054](https://dhis2.atlassian.net/browse/DHIS2-17054)

**Tech summary**
- handle empty expression value in `processDisplayKeyValuePair`
- added a unit test for the `undefined` value use case
- small CSS adjustments

[DHIS2-17054]: https://dhis2.atlassian.net/browse/DHIS2-17054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/rules-engine/src/processors/rulesEffectsProcessor/rulesEffectsProcessor.js
```diff
@@ -239,13 +239,17 @@ export function getRulesEffectsProcessor(
     }
 
     function processDisplayKeyValuePair(effect: ProgramRuleEffect): any {
+        const { data } = effect;
+        const dataToString = value => (typeof value == 'number' ? numberToString(value) : String(value));
+        const value = data !== undefined ? dataToString(data) : '';
```
superskip, line 244:
```suggestion
        const data = effect.data || '';
```

packages/rules-engine/src/processors/rulesEffectsProcessor/rulesEffectsProcessor.js
```diff
@@ -239,13 +239,17 @@ export function getRulesEffectsProcessor(
...
+
         return {
             type: effectActions.DISPLAY_KEY_VALUE_PAIR,
             id: effect.location,
             displayKeyValuePair: {
                 id: effect.id,
                 key: effect.displayContent,
-                value: typeof effect.data == 'number' ? numberToString(effect.data) : String(effect.data),
+                value,
```
superskip, line 252:
```suggestion
                value: typeof data === 'number' ? numberToString(data) : String(data),
```

superskip (commented):
I found a small simplification, posted as comments. Regarding the css changes, will the bottom margin be constant 0 now? How did it work previously (what does '& + &' mean)?

packages/rules-engine/src/processors/rulesEffectsProcessor/rulesEffectsProcessor.js
superskip: Hold on, maybe `effect.data` could be `false`.. In that case this won't work.

Revised suggestion:
```suggestion
        const data = effect.data !== undefined ? effect.data : '';
```

simonadomnisoru:
> I found a small simplification, posted as comments. Regarding the css changes, will the bottom margin be constant 0 now? How did it work previously (what does '& + &' mean)?

Hey,
I applied the small simplifications you suggested. As for the bottom margins, yes, they are all constant now as requested in the ticket. 
From what I understand from [next sibling combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Next-sibling_combinator) and [nesting selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector#appending_the_nesting_selector), `& + &` compiles to `.listItem + .listItem` and matches from the second `.listItem` class until the last one. This means that before the first listItem had a different bottom margin than the second and the rest ones.
Thanks for the feedback!

superskip:
> From what I understand from [next sibling combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Next-sibling_combinator) and [nesting selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector#appending_the_nesting_selector), & + & compiles to .listItem + .listItem and matches from the second .listItem class until the last one. This means that before the first listItem had a different bottom margin than the second and the rest ones.

Thanks for the explanation :)

@CooperJoe Do you recall what you tried to achieve by using the `& + &` selector?

superskip (approved):

geethaalwan (changes requested):
Tested successfully on 2.42,2.41.1,2.40.4,2.39.6 versions

geethaalwan (approved):

PR #3643, opened 2024-05-15
fix: [DHIS2-17399] use server version to send correct aoc payload

eirikhaugstulen:
Tech-summary:
- Use server version to send correct AOC payload when there are multiple categories in a category combo assigned to a program.


src/core_modules/capture-core/components/DataEntries/SingleEventRegistrationEntry/DataEntryWrapper/DataEntry/epics/getConvertedNewSingleEvent.js
```diff
@@ -35,7 +36,6 @@ export const getNewEventServerData = (state: ReduxState, formFoundation: RenderF
             program: state.currentSelections.programId,
             programStage: formFoundation.id,
             orgUnit: state.currentSelections.orgUnitId,
-            ...getApiCategoriesArgument(state.currentSelections.categories),
```
JoakimSM, line 38: Doesn't seem to work when adding events for event programs. This might be the culprit?

JoakimSM (approved):
LGTM

linadhis2:
Testing complete via Netlify on all supported version bases, looks good from QA perspective (the only point I have is about automated tests, but due to an approaching release I'll let it slide, but if you could add some - would be great).

PR #3657, opened 2024-05-29
feat: [DHIS2-17068] Check Related Stage Eligibility for Storing Events

henrikmv:
[DHIS2-17068](https://dhis2.atlassian.net/browse/DHIS2-17068?atlOrigin=eyJpIjoiMDMzZDI2ODE0MTgxNGU3NGFiODdlMjBhZDhhM2MxMDIiLCJwIjoiaiJ9)

- Implemented `useCanAddNewEventToStage` hook to determine if a related stage is eligible to store the event.
- The hook checks if the stage is repeatable or non-repeatable but has no existing events.
- Removed map function for the three radio buttons for improved code readability.
- Disables the radio button for:
        1. **SCHEDULE_IN_ORG** and  **ENTER_DATA** when the related stage is not repeatable and already has an event.
        2. **LINK_EXISTING_RESPONSE** when the related stage has no events
- Added a tooltip to explain why the button is disabled.

[DHIS2-17068]: https://dhis2.atlassian.net/browse/DHIS2-17068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetRelatedStages/RelatedStagesActions/RelatedStagesActions.component.js
```diff
@@ -80,15 +84,27 @@ export const RelatedStagesActionsPlain = ({
...
-                    <Radio
-                        key={key}
-                        name={`related-stage-action-${key}`}
-                        checked={key === selectedAction}
-                        disabled={key === RelatedStagesActionTypes.LINK_EXISTING_RESPONSE && !linkableEvents.length}
-                        label={mainOptionTranslatedTexts[key]}
-                        onChange={(e: Object) => updateSelectedAction(e.value)}
-                        value={key}
-                    />
+                    <div className={classes.fieldWrapper} key={key}>
```
eirikhaugstulen, line 87: It seems like this div changed the styling quite drastically. It would be nice if we could keep the design look the same.

Before:
image

After:
image

src/core_modules/capture-core/components/WidgetRelatedStages/RelatedStagesActions/RelatedStagesActions.component.js
```diff
@@ -80,15 +84,27 @@ export const RelatedStagesActionsPlain = ({
...
+                                interpolation: { escapeValue: false },
+                            })}
+                            enabled={!isRelatedStageEligible}
+                        >
+                            <Radio
+                                name={`related-stage-action-${key}`}
+                                checked={key === selectedAction}
+                                disabled={(key === RelatedStagesActionTypes.LINK_EXISTING_RESPONSE
+                                    && !linkableEvents.length)
+                                    || !isRelatedStageEligible}
```
eirikhaugstulen, line 100: In general, this line is a bit difficult to read. Perhaps you can extract it into a separate variable on line 87.

`const isDisabled = (key === ...)`

Also, the link to an existing should not be disabled if there is an event to link to.

image

src/core_modules/capture-core/components/WidgetRelatedStages/hooks/useRelatedStageEligibility.js
```diff
@@ -0,0 +1,15 @@
...
+export const useRelatedStageEligibility = (programStage, existingRelatedEvents) => {
+    const [isRelatedStageEligible, setIsRelatedStageEligible] = useState(false);
+
+    useEffect(() => {
+        if (programStage && existingRelatedEvents) {
+            const eligible = programStage.repeatable ||
+                (!programStage.repeatable && existingRelatedEvents.length === 0);
+            setIsRelatedStageEligible(eligible);
+        }
+    }, [programStage, existingRelatedEvents]);
```
eirikhaugstulen, line 12: Perhaps we can refactor this to use `useMemo` instead of `useState + useEffect`. This often leads to the same effect, but with less code and performance overhead 👍 Let me know if it's unclear.

eirikhaugstulen (commented):
Hey @henrikmv! This is a great start - had some comments, but seems good overall.

eirikhaugstulen (approved):
Great work @henrikmv 🎉

geethaalwan (approved):
Tested successfully on 2.42,2.41.1,2.40.4,2.39.6 versions

PR #3658, opened 2024-05-30
feat: [DHIS2-15462] Use dhis2 UI calendarinput component in forms

alaa-yahia:
Implements [DHIS2-15462](https://dhis2.atlassian.net/browse/DHIS2-15462)

---

### Description
@dhis2/ui `CalendarInput` is used when capturing dates in forms.

[DHIS2-15462]: https://dhis2.atlassian.net/browse/DHIS2-15462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (commented):
Hey @alaa-yahia, great job on this PR 👏 ! 

From capture-app side, I only added a small comment about hiding the default label. Additionally, can you please take a look and fix the failed cypress tests related to the calendar data test selectors?

From the UI library side, these are the pending features that need to be added before merging this PR:

1. minimum and maximum support data - https://github.com/dhis2/ui/pull/1504
2. support for typing/removing a date - https://github.com/dhis2/ui/pull/1445
3. support for DD-MM-YYYY format - I couldn't find a ticket/PR for this. Let's sync up for the next step on this.

Thank you!

alaa-yahia:
> Hey @alaa-yahia, great job on this PR 👏 !
> 
> From capture-app side, I only added a small comment about hiding the default label. Additionally, can you please take a look and fix the failed cypress tests related to the calendar `data-test` selectors?
> 
> From the UI library side, these are the pending features that need to be added before merging this PR:
> 
> 1. minimum and maximum support data - [feat: support min & max dates in CalendarInput ui#1504](https://github.com/dhis2/ui/pull/1504)
> 2. support for typing/removing a date - [feat(calendar input): make input editable ui#1445](https://github.com/dhis2/ui/pull/1445)
> 3. support for DD-MM-YYYY format - I couldn't find a ticket/PR for this. Let's sync up for the next step on this.
> 
> Thank you!

Hey,
All the three has work between the ui and multi calendar date library.
The one for editable input you mentioned is an old one. 
This is the [new one](https://github.com/dhis2/ui/pull/1504), I will update the name of the pull request when finalising it. I will add a ticket for DD-MM-YYYY, the functionality is already implemented in the multi calendar library.

minimum and maximum support data - [feat: support min & max dates in CalendarInput ui#1504](https://github.com/dhis2/ui/pull/1504)
support for typing/removing a date - [feat: support min & max dates in CalendarInput ui#1504](https://github.com/dhis2/ui/pull/1504)
support for DD-MM-YYYY format - [support min & max date, DD-MM-YYYY date format](https://github.com/dhis2/multi-calendar-dates/pull/36)

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
```diff
@@ -1,206 +1,76 @@
...
-
-        this.props.onFocus && this.props.onFocus();
     }
 
-    handleDateSelected(value: string) {
-        this.props.onBlur(value);
-        this.hidePopover();
+    handleDateSelected(value: {calendarDateString: string}) {
+        const date = value?.calendarDateString;
+        this.props.onBlur(date);
```
simonadomnisoru, line 31: When the date is mandatory, the red error background appears for a second. 

https://github.com/user-attachments/assets/1afd5dcc-88c5-4073-9618-a8714ae9e8a2

I think that checking for `undefined` will fix this.
```
if (date !== undefined) {
    this.props.onBlur(date);
}
```
WDYT? Thanks!

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
```diff
@@ -1,206 +1,76 @@
...
             >
-                { /* // $FlowFixMe */}
-                {/* $FlowFixMe[prop-missing] automated comment */}
-                <DateInput
-                    onFocus={this.handleTextFieldFocus}
-                    onBlur={this.handleTextFieldBlur}
-                    width={calculatedInputWidth}
-                    {...splittedPassOnProps.input}
+                <CalendarInput
+                    {...passOnProps}
```
simonadomnisoru, line 63: In `passOnProps`, there are a lot of old props that are no longer used and it's quite hard to keep track. Is it possible to clean it up a bit and leave only the props that `CalendarInput` actually uses? You can also clean up unused props from parent components of Date.component.js. Thanks!

src/core_modules/capture-ui/DateAndTimeFields/DateField/Date.component.js
alaa-yahia: Yes! it solved the issue. Thanks

simonadomnisoru (approved):
Solid work, well done @alaa-yahia ! 👏
Let's move it to QA and then discuss in one of our Tuesday syncs if we want to release with [alpha](https://github.com/dhis2/capture-app/pull/3658/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R22) or wait for the `@dhis/ui` stable release.

alaa-yahia:
> Just one question regarding validation: The new calendar component is doing validation internally, which makes sense since it's supposed to support exotic calendar types. But as far as I can see the Capture app is still doing it's own validation - is it possible to make it use the validation result from the new calendar component instead? I can't see how this can be achieved with the current interface though (https://ui.dhis2.nu/components/calendar#props), but maybe that document is outdated.

Yeah @superskip the document is outdated as I haven't worked on docs yet.
I think changing how validation is returned from `useDatePicker` would make it possible for parent components to receive validation results. Specifically passing validation result to the object passed to [onDateSelect](https://github.com/dhis2/multi-calendar-dates/blob/085558c8a0cf2fca5c3545c21564a38af01b40c8/src/hooks/useDatePicker.ts#L154) which calls the parent onDateSelect.

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40.5,2.39.7 versions

src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js
```diff
@@ -390,7 +392,7 @@ export class FormBuilder extends React.Component<Props> {
         this.commitFieldUpdate({ fieldId, validators }, value, options);
     }
 
-    async commitFieldUpdate({ fieldId, validators, onIsEqual }: FieldCommitConfig, value: any, options?: ?FieldCommitOptions) {
+    async commitFieldUpdate({ fieldId, validators, onIsEqual }: FieldCommitConfig, value: any, options?: ?FieldCommitOptions, internalError?: ?Object) {
```
JoakimSM, line 395: We should spend a bit of time on some cleanup here to make this maintainable over time. Instead of adding a new param here in `internalError` we should use the `options` param we already have. We will have to add the `errorCode` to the `FieldCommitOptions`. Then we need to make some changes here: https://github.com/dhis2/capture-app/blob/c07e3015fddfa44544050a0e0393178aa8f0a010/src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js#L446 and to the errorMessage two lines below. Currently, we run the validations, but if options.valid and options.error are set, we just override whatever the validations say. We should remove this override. This probably requires some investigation as we will have to figure out for what value types this logic is used and add a validator that does this for us. I.e., when we call validateField (https://github.com/dhis2/capture-app/blob/c07e3015fddfa44544050a0e0393178aa8f0a010/src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js#L434) we should pass on the `options` object instead of `internalError`. For the value types that were using the override, we should add a validator that sets `valid` and `errorMessage` the normal way.

src/core_modules/capture-core/components/D2Form/field/validators/getValidators.js
```diff
@@ -212,14 +212,14 @@ function buildTypeValidators(metaData: DataElement | DateDataElement): ?Array<Va
 
     validatorContainersForType = validatorContainersForType.map(validatorContainer => ({
         ...validatorContainer,
-        validator: (value: any) => {
+        validator: (value: any, validationContext: ?Object, internalError: ?Object) => {
```
JoakimSM, line 215: We should end up having `commitOptions` (containing the internal error) as the third param here instead of `internalError`.

src/core_modules/capture-core/components/D2Form/field/validators/getValidators.js
```diff
@@ -115,7 +115,7 @@ const validatorsForTypes = {
         {
             validator: isValidDate,
             message: errorMessages.DATE,
-            type: validatorTypes.TYPE_BASE,
+            type: validatorTypes.INTERNAL,
```
JoakimSM, line 118: Every error expect future date should have validatorTypes.TYPE_BASE. Future date should have TYPE_EXTENDED. Probably easiest and most readable if we make two validators so we don't have to implement new logic here for determining the type (we have two validators now, but the second one about the allowFutureDate is not implemented the way we want it)

The reason is that we depending on the program configuration, can accept saving an incomplete form. But the BASE-validators must always pass (i.e., your date field should be a date, your number field should be a number etc.)

src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js
JoakimSM: Regarding the override of valid and error. Checked the git history and talked to Eirik, it is related to plugins and we should skip validations if we are committing from a plugin. I think the Formbuilder file should end up looking something like this: https://github.com/dhis2/capture-app/blob/DHIS2-15462-formbuilder/src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js

src/core_modules/capture-core/utils/validators/form/isValidNonFutureDate.js
```diff
@@ -3,6 +3,10 @@ import i18n from '@dhis2/d2-i18n';
 import moment from 'moment';
 import { parseDate } from '../../converters/date';
 
+const CUSTOM_VALIDATION_MESSAGES = {
+    INVALID_DATE_MORE_THAN_MAX: i18n.t('A date in the future is not allowed'),
+};
+
 export const isValidNonFutureDate = (value: string) => {
```
simonadomnisoru, line 10: The future date error message is not displayed when selecting an enrollment date in the future, instead the "Please provide a valid date" error is displayed. Can this be fixed? 
I also think `isValidNonFutureDate` should receive the `internalComponentError` object from the [validators](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/D2Form/field/validators/getValidators.js#L121):
`validator: (value: string, allowFutureDate, internalComponentError) => (allowFutureDate ? true : isValidNonFutureDate(value, internalComponentError)),`

src/core_modules/capture-ui/internal/DateTimeInput/DateTimeDate.component.js
```diff
@@ -1,18 +1,19 @@
...
         <DateField
-            value={value || ''}
+            value={value}
             {...passOnProps}
         />
     );
 }
 
-export const DateTimeDate = withFocusSaver()(withShrinkLabel()(DateTimeDatePlain));
+export const DateTimeDate = withFocusSaver()(withShrinkLabel()(withTextFieldFocusHandler()(DateTimeDatePlain)));
```
simonadomnisoru, line 19: When manually typing in the Date part of the `DateTime` field, the placeholder covers the date. Can the placeholder be removed while typing like in the `Age` or `Date` fields? Thank you!Thanks!

Screenshot 2024-10-31 at 08 57 38

src/core_modules/capture-core/utils/validators/form/isValidNonFutureDate.js
JoakimSM: Yes, we should just check whether the internal error code from the component is the MAX_DATE one and return our own custom capture app error message if that is the case.

src/core_modules/capture-core-utils/validators/form/date.validator.js
JoakimSM, line 1: The intention of this file in `capture-core-utils` was to have a very generic date validator. Since this function is now very much tied to the `internalComponentError`, I think we should just remove the `date.validator.js`, `age.validator.js` and `dateTime.validator.js` from `capture-core-utils` and move the code to the validators in `capture-core/utils/validators/form/`.

src/core_modules/capture-core/components/DataEntry/dataEntryField/internal/dataEntryField.utils.js
```diff
@@ -1,28 +1,28 @@
...
         const validator = validatorContainer.validator;
-        const result = validator(value);
+        const result = validator(value, internalComponentError);
 
         if (result === true || (result && result.valid)) {
             return false;
         }
 
-        message = (result && result.message) || validatorContainer.message;
+        message = (result && result.errorMessage) || validatorContainer.message;
```
JoakimSM, line 25: I really think we shouldn't change this from `message` to `errorMessage` as this is being used by a lot of validators (I looked up one of the validators and it was still using `message`). Let's make this easy on ourselves and keep using `message`. For example here we are using message https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/DataEntries/Enrollment/fieldValidators/enrollmentDate.validatorContainersGetter.js#L23 (this particular validator referenced here should also be rewritten to use our new validation rules)

JoakimSM:
We will have to fix the validators for the fields that are not data elements / tracked entity attributes. I noticed that these in the enrollment form have not been updated (haven't check new/edit event, relationships etc):
https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/DataEntries/Enrollment/EnrollmentDataEntry.component.js#L118
https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/DataEntries/Enrollment/EnrollmentDataEntry.component.js#L168

JoakimSM (commented):
Will also need to update the validators referenced here: 
Also need to update the validators referenced here: https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/DataEntry/DataEntry.component.js#L154

https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js#L142-L143

https://github.com/dhis2/capture-app/blob/48713020465d8764a1ea3104a8d873216b9157d6/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js#L190

src/core_modules/capture-core/components/DataEntry/dataEntryField/internal/dataEntryField.utils.js
alaa-yahia: I modified this code to display the error message that is coming from calendarInput as what is returned from isValidDate is errorMessage not message.
I will change the code to be `(result && result.errorMessage) || (result && result.message) || validatorContainer.message;` so all scenarios work.

src/core_modules/capture-core/components/WidgetEventSchedule/ScheduleDate/ScheduleDate.component.js
```diff
@@ -36,7 +34,14 @@ const ScheduleDatePlain = ({
             onSetFocus={() => {}}
             onFocus={() => { }}
             onRemoveFocus={() => { }}
-            onBlur={(e) => { setScheduleDate(convertStringToDateFormat(e)); }}
+            onBlur={(e, internalComponentError) => {
+                const { error } = internalComponentError;
+                if (error) {
+                    setScheduleDate('');
+                    return;
+                }
```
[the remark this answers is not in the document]
JoakimSM: Ok, let's leave this as it is for now and we can improve on it later. Made a ticket https://dhis2.atlassian.net/browse/DHIS2-18391.

src/core_modules/capture-core/components/D2Form/FormBuilder/FormBuilder.component.js
```diff
@@ -115,9 +126,15 @@ export class FormBuilder extends React.Component<Props> {
             .reduce(async (passPromise, currentValidator) => {
                 const pass = await passPromise;
                 if (pass === true) {
-                    let result = currentValidator.validator(value, validationContext);
+                    const validatorContext = commitOptions?.errorCode ?
+                        { error: commitOptions.error, errorCode: commitOptions.errorCode } :
+                        validationContext; // For UniqueValidator
```
JoakimSM, line 131: Let's always pass on both the errors and the validationContext so we don't have to potentially change this logic if we are writing new validators. If you pass on errors as the second argument and validatorContext as the third one, I think you only have to change the unqiue validator.

JoakimSM (commented):
We are very close now @alaa-yahia.

JoakimSM:
> Also, we should probably validate after rules engine execution and when opening the form. But let's discuss this at the end after fixing the other stuff.

Like we talked about, made a separate ticket for this: https://dhis2.atlassian.net/browse/DHIS2-18468

simonadomnisoru (approved):
Good job! 🎉

JoakimSM (approved):
Nice 🎉 Looks good to me now. Let's have QA redo the testing before merging.

geethaalwan (approved):
Tested successfully on 2.42,2.41.3,2.40.7,2.39.8 versions

PR #3683, opened 2024-06-18
fix: [DHIS2-17488] hide program stage if no data read access

JoakimSM:
Cypress tests are coming in https://dhis2.atlassian.net/browse/DHIS2-17590

PR #3686, opened 2024-06-19
chore: [DHIS2-17590] pipeline for multi user Cypress testing

JoakimSM:
- Ability to run tests as a different user by providing `@user:`tag in the feature file, e.g. `@user:trackerAutoTestRestricted`
- User specified in the tag must match environment variables for the username/password (i.e., we will have to add environment variables for the users we need to run tests as)
-  Restructured the Cypress files:
    - both the feature file and the step definitions reside within the same directory
    - moved the "shared-steps"

(Based off the branch in https://github.com/dhis2/capture-app/pull/3683, so those changes are visible as well until it is merged)


cypress/e2e/EnrollmentPage/StagesAndEventsWidget/StagesAndEventsWidget.feature
```diff
@@ -53,6 +53,13 @@ Feature: User interacts with Stages and Events Widget
         When you click New First antenatal care visit event
         Then you should navigate to Add new page #/enrollmentEventNew?enrollmentId=ek4WWAgXX5i&orgUnitId=DwpbWkiqjMy&programId=WSGAb5XwJ3Y&stageId=WZbXY0S00lP&teiId=yFcOhsM1Yoa
 
+
     Scenario: User can not go to Add new page if stage is not repeatable and there is event in the stage
         Given you open the enrollment page by typing #enrollment?programId=IpHINAT79UW&orgUnitId=UgYg0YW7ZIh&teiId=fhFQhO0xILJ&enrollmentId=gPDueU02tn8
         Then you should see the disabled button New Birth event
+
+    # Waiting for pipline to update DB/DB update for 39 and 41
```
simonadomnisoru, line 61: How about `2.38` DB update? Shouldn't the changes made in https://github.com/dhis2/capture-app/pull/3683 be tested against version `2.38`, until [DHIS2-17545](https://dhis2.atlassian.net/browse/DHIS2-17545) is completed? Thanks!

JoakimSM: Good catch. It seems like we have a pipeline problem with 38 also though, so I'm in favour of not spending any time on 38.

simonadomnisoru (approved):

PR #3698, opened 2024-06-30
fix: chrome crashes when opening applications-tab

eirikhaugstulen:
Summary:
- For the last month, google chrome has been crashing every time you try to open devtools/application.
- This was because we had an undiscovered error in our manifest.json file
- Changed icon type to `image/x-icon`


eirikhaugstulen:
This won't solve all problems, as we also need the generated `.d2/public/manifest.json` file to adhere to the specs, but this is what's needed from the capture app I believe.

simonadomnisoru (approved):
This fix works in my localhost

JoakimSM (approved):

PR #3704, opened 2024-07-03
fix: [DHIS2-17694] Edit button not showing on profile widget

eirikhaugstulen:
Tech-summary:
- Fixing some of the logic introduced in #3610 
- Use data write access instead of metadata write access


simonadomnisoru (approved):
LGTM!

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.container.js
```diff
@@ -22,11 +22,11 @@ const useContainsAutoGeneratedEvent = program =>
 
 const useCanAddNew = (enrollments, programId, tetAccess) =>
     useMemo(
-        () => tetAccess?.write &&
+        () => tetAccess?.data?.write &&
             enrollments
                 .filter(item => item.program === programId)
                 .every(item => item.status !== plainStatus.ACTIVE),
-        [enrollments, programId, tetAccess?.write],
+        [enrollments, programId, tetAccess?.data?.write],
```
superskip, line 29: This `?.` notation expands into a complicated expression, and I'm not yet used to seeing complicated expressions in the dependency array. Can we store `tetAccess?.data?.write` in a variable so we 1. don't need to evaluate it twice and 2. keep me unused of seeing complicated expressions in the dependency array? (Maybe this turned into a dilemma when I wording my request this way..)

eirikhaugstulen: Sure. Just removing everything behind `tetAccess` should have the same result either way, as there will never be partial updates of this object. Reverting to just `[enrollments, programId, tetAccess]`.

JoakimSM (approved):

linadhis2:
LGTM from QA perspective

PR #3719, opened 2024-07-18
refactor: [DHIS2-17652]  Replace Material-UI Avatar

henrikmv:
[DHIS2-17652](https://dhis2.atlassian.net/browse/DHIS2-17652)
- The UserAvatar component from the DHIS2 UI library is processing the image URL so it gets invalid in this case. Due to the simplicity of the component, it was made a new component adjusted to the use case, instead of making a function to get the URL right. 

[DHIS2-17652]: https://dhis2.atlassian.net/browse/DHIS2-17652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-ui/UserAvatarImage/UserAvatarImage.component.js
```diff
@@ -0,0 +1,56 @@
...
+
+type Props = {
+    imageUrl: string,
+    dataTest: string,
+    classes: Object,
+    className: Object,
+    size: 'extrasmall' | 'small' | 'medium' | 'large' | 'extralarge',
+};
+
+const UserAvatarImagePlain = ({ imageUrl, dataTest, classes, className, size }: Props) => (
```
simonadomnisoru, line 44: @henrikmv,
I agree with the usage of a made UI component. In this use case, the displayed image is the [first TEA ](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/CardList/CardList.selectors.js#L13)of type `IMAGE` with a value, while the `UserAvarar` component in the UI library is intended for the user logged into the system.
The only suggestion I have is to change the name of the component to something more general, not related to the user or avatar. Can it be renamed to something like `CardImage` or just `Image`? WDYT?
Thank you!

henrikmv: Thank you! I changed to CardImage

src/core_modules/capture-ui/UserAvatarImage/UserAvatarImage.component.js
```diff
@@ -0,0 +1,56 @@
+// @flow
```
simonadomnisoru, line 1: Can you also change the name of this file and its parent folder to `CardImage`? Thanks!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42 version

PR #3720, opened 2024-07-19
feat: [DHIS2-17771] Org unit context in Stages and Events widget

henrikmv:


src/core_modules/capture-core/components/Tooltips/TooltipOrgUnit/TooltipOrgUnit.component.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
simonadomnisoru, line 1: missing // @flow for the file

simonadomnisoru (approved):
Great job 👏 !

geethaalwan (approved):
Tested successfully on 2.42,2.41.1,240.5,2.39.7 versions

src/core_modules/capture-core/components/Tooltips/TooltipOrgUnit/TooltipOrgUnit.component.js
```diff
@@ -0,0 +1,36 @@
...
+    orgUnitName: string,
+    ancestors?: Array<{| displayName: string, id: number |}>,
+    tooltip?: string,
+};
+
+export const TooltipOrgUnit = ({ orgUnitName, ancestors = [], tooltip }: Props) => {
+    const ancestorNames = ancestors.map(ancestor => ancestor.displayName);
+    ancestorNames.push(orgUnitName);
+    const orgUnitNameFullPath = ancestorNames.join(' / ');
+    const tooltipParts = tooltip ? tooltip.split(orgUnitName) : [orgUnitName];
```
JoakimSM, line 15: Can we simplify this? Do we need the tooltip to span more than the org unit name?

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.js
```diff
@@ -21,7 +21,7 @@ import {
     useUserRoles,
     useTeiDisplayName,
 } from './hooks';
-import { DataEntry, dataEntryActionTypes, TEI_MODAL_STATE, convertClientToView } from './DataEntry';
+import { DataEntry, dataEntryActionTypes, TEI_MODAL_STATE, convertClientToList } from './DataEntry';
```
JoakimSM, line 24: What is the reasoning behind changing this?

src/core_modules/capture-core/metadataRetrieval/orgUnitName/orgUnitName.js
```diff
@@ -130,44 +143,71 @@ export const useOrgUnitNames = (orgUnitIds: Array<string>): {
...
+            ancestors.push({
+                displayName: orgUnit.displayName,
+                id: orgUnitId,
+            });
+
+            return ancestors;
+        };
+
+        const ancestors = getAncestors(orgUnitId);
+        ancestors.pop();
```
JoakimSM, line 193: NIT: I would prefer not to pop for better readability. I think we can do this:
```suggestion
        const ancestors = cachedOrgUnit.ancestor ? getAncestors(cachedOrgUnit.ancestor) : [];
```

superskip: I would also suggest moving `getAncestors` out into global scope. In this case I think it will reveal an error (`orgUnitId` not defined).

src/core_modules/capture-core/components/Tooltips/TooltipOrgUnit/TooltipOrgUnit.component.js
```diff
@@ -0,0 +1,22 @@
...
+
+type Props = {
+    orgUnitName: string,
+    ancestors?: Array<string>,
+    tooltip?: string,
+};
+
+export const TooltipOrgUnit = ({ orgUnitName, ancestors = [], tooltip }: Props) => {
+    const fullPath = [...ancestors, orgUnitName].join(' / ');
+    const [before, after] = tooltip ? tooltip.split(orgUnitName) : [orgUnitName];
```
superskip, line 13: This will not work correctly if somebody has an orgunit called `Start`... I looked a bit into ways around this issue: the one that seems like the least work to implement is to redefine `tooltip` to be an array of strings. We could then get such arrays by tampering with how we generate the translations; first we could use a delimiter symbol like `\0` in place of the orgunit name, and then we could split the resulting string with respect to the delimiter.

superskip (commented):
Looks very close to done now :+1:

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.component.js
```diff
@@ -129,19 +132,20 @@ export const WidgetEnrollmentPlain = ({
                             <span className={classes.icon} data-test="widget-enrollment-icon-orgunit">
                                 <IconDimensionOrgUnit16 color={colors.grey600} />
                             </span>
-                            {i18n.t('Started at {{orgUnitName}}', {
-                                orgUnitName,
-                                interpolation: { escapeValue: false },
-                            })}
+                            <span>
+                                {i18n.t('Started at ')}
```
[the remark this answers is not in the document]
JoakimSM: Talked to Henrik about this one. I agree it's important to be conscious about this problem, but in this case I suggest we simplify as you could really think of the string as two columns ("started at" and then the org unit name).

src/core_modules/capture-core/components/Tooltips/TooltipOrgUnit/TooltipOrgUnit.component.js
JoakimSM: Talked to Henrik. Let's simplify and avoid the splitting.
https://github.com/dhis2/capture-app/pull/3720#discussion_r1738263188

superskip (approved):
Great work! :tada:

eirikhaugstulen (approved):
Well done, @henrikmv! 🎉

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40.6,2.39.7 versions

JoakimSM (approved):

PR #3721, opened 2024-07-20
refactor: [DHIS2-17650] Replace Material-UI Table, TableBody, TableCell, TableHead and TableRow

henrikmv:
[DHIS2-17650](https://dhis2.atlassian.net/browse/DHIS2-17650)

- Changed from Material-UI **Table, TableBody, TableCell, TableHead and TableRow** to DHIS2/UI
- Style changes: Checkboxes are placed to the right to keep them away from the drag/drop icon. Small changes to the table (added border, header row color, etc)
- Changed from using **DragSource** and **DropTarget** to using **useDrag** and **useDrop**
- Had to remove the div to style the table properly. Since `DndProvider` was expecting the div, I got this error message: _Only native element nodes can now be passed to React DnD connectors. You can either wrap DataTableRow into a div or turn it into a drag source or a drop target itself_.  Solved by using **useRef** to set the `DataTableRow` to ref.
- Updated Cypress tests to look after the checkbox in tablerow 

[DHIS2-17650]: https://dhis2.atlassian.net/browse/DHIS2-17650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):
LGTM!

henrikmv:
@eirikhaugstulen requested some changes to improve usability, so I have implemented them now.

Note: I can see that the hover effect can be confusing under drag, but I found it hard to fix it.
 
Would you give me another review with the new changes implemented @simonadomnisoru? Thanks!

src/core_modules/capture-core/components/ListView/ColumnSelector/DragDropList/DragDropListItem.component.js
```diff
@@ -1,107 +1,69 @@
...
-    DragSource(
-        ItemTypes.LISTITEM,
-        cardSource,
-        (connect, monitor) => ({ connectDragSource: connect.dragSource(), isDragging: monitor.isDragging() }),
-    )(DropTarget(
-        ItemTypes.LISTITEM,
-        cardTarget,
-        connect => ({ connectDropTarget: connect.dropTarget() }),
-    )(Index));
+    const opacity = isDragging ? 0.5 : 1;
```
simonadomnisoru, line 54: > Note: I can see that the hover effect can be confusing under drag, but I found it hard to fix it.

How about setting the opacity to `0` when dragging? This way the drawn row will look empty and can help with the confusion of the hover effect (this is also how it currently works in play with the material UI components).

henrikmv: Skjermbilde 2024-07-31 kl  11 35 33
I agree that setting the opacity to `0` makes it better so I will change it to `const opacity = isDragging ? 0 : 1;`. However, the **table row hover** is still confusing. I added a picture where I am dragging the _Gender_ , but the hover is marking _Last name_.

simonadomnisoru: I use Firefox where the hover issue doesn't happen, but I was able to reproduce it in Chrome. I bealive you can remove the hover styles altogether with something like this in the `DragDropListItem.component.js`.

// the styles
```
const styles = {
    rowWithoutHover: {
        '&:hover > td': {
            backgroundColor: 'transparent !important',
        },
    },
};
```

// and in the render use the class
```
return (
        <DataTableRow ref={ref} style={{ opacity }} className={classes.rowWithoutHover} draggable>
            <DataTableCell>{text}</DataTableCell>
            ......
```

henrikmv: Thank you, @simonadomnisoru! That works. With the changes, the hover was completely removed, so in addition to your suggested changes, I added a state to tell if any element in the list is being dragged. Based on that state, I will disable the hover. What do you think? I appreciate your help!

simonadomnisoru: It looks great now 👏

simonadomnisoru (approved):

eirikhaugstulen (approved):
Great work @henrikmv!

geethaalwan (approved):
Tested successfully on 2.42 version

PR #3735, opened 2024-07-28
feat: [DHIS2-16125] hide program stage under certain circumstances

henrikmv:
[DHIS2-16125](https://dhis2.atlassian.net/browse/DHIS2-16125)

- Checked if program rule `Prevent adding new events to stage` is in effect when no events are in stage. If so, the program stage is hidden.
- Change variable names for readability
- Removed `ConditionalTooltip` and `disabled` prop for the scenario where there are no events, because there are no use cases where this button can be disabled.
- Changed Cypress to verify that the stage is hidden (under the certain circumstances) instead of looking after the disabled button in the hidden program stage

[DHIS2-16125]: https://dhis2.atlassian.net/browse/DHIS2-16125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/Stage.component.js
+++ b/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/Stage.component.js
@@ -28,7 +28,6 @@ const rulesEffectHideProgramStage = (ruleEffects, stageId) => (
 );
 
 export const StagePlain = ({ stage, events, classes, className, onCreateNew, ruleEffects, ...passOnProps }: Props) => {
-    console.log('ruleEffects', ruleEffects);
     const [open, setOpenStatus] = useState(true);
     const { id, name, icon, description, dataElements, hideDueDate, repeatable, enableUserAssignment } = stage;
     const preventAddingNewEvents = rulesEffectHideProgramStage(ruleEffects, id);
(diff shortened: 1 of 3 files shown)


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/Stage.component.js
```diff
@@ -59,31 +65,23 @@ export const StagePlain = ({ stage, events, classes, className, onCreateNew, rul
...
+                    hiddenProgramStage={preventAddingNewEvents}
                     {...passOnProps}
                 /> : (
-                    <ConditionalTooltip
-                        content={i18n.t("You can't add any more {{ programStageName }} events", {
-                            programStageName: name,
+                    <Button
+                        small
+                        secondary
+                        disabled={preventAddingNewEvents}
```
simonadomnisoru, line 74: It seems that there are no longer any use cases where this button can be disabled. If `ConditionalTooltip` has been removed, does it make sense to remove `disabled={hiddenProgramStage}` as well?

henrikmv:
Thank you for the review, @simonadomnisoru! I have committed the changes suggested now

simonadomnisoru (approved):

henrikmv:
@simonadomnisoru, I had a Cypress test breaking because the test was looking for a disabled button in a stage that is now hidden due to the changes implemented in this branch. Do you want to have a look? Thank you!

simonadomnisoru (approved):
LGTM!

geethaalwan (approved):
Tested successfully on 2.42,2.41.1,.2.40.5,2.39.7 versions

PR #3738, opened 2024-07-31
fix: [DHIS2-17632][DHIS2-17633] restrict invalid category combo for orgUnit

eirikhaugstulen:
Tech-summary:
- Make sure that a user is not able to select an invalid combination of category option that is not assigned to an orgUnit


src/core_modules/capture-core/hooks/useCategoryComboIsValidForOrgUnit.js
```diff
@@ -0,0 +1,51 @@
...
+    const isInvalid = useMemo(() => {
+        if (!data || !data.length) {
+            return false;
+        }
+
+        return data.every(({ organisationUnits }) => !organisationUnits[selectedOrgUnitId]);
+    }, [data, selectedOrgUnitId]);
+
+    return {
+        isInvalid,
```
simonadomnisoru, line 46: Can this boolean be renamed to something more descriptive like `categoryOptionIsInvalidForOrgUnit` or `categoryOptionIsInvalidInOrgUnitScope`?

src/core_modules/capture-core/components/Pages/MainPage/MainPage.constants.js
```diff
@@ -2,5 +2,6 @@ export const MainPageStatuses = Object.freeze({
     DEFAULT: 'DEFAULT',
     WITHOUT_ORG_UNIT_SELECTED: 'WITHOUT_ORG_UNIT_SELECTED',
     WITHOUT_PROGRAM_CATEGORY_SELECTED: 'WITHOUT_PROGRAM_CATEGORY_SELECTED',
+    CATEGORY_OPTION_INVALID_FOR_ORG_UNIT: 'CATEGORY_OPTION_INVALID_FOR_ORG_UNIT',
```
simonadomnisoru, line 5: Is it possible to use the same const string for both the Main page and the New page status, either `CATEGORY_OPTION_INVALID_FOR_ORG_UNIT` or `INVALID_CATEGORY_OPTION_SELECTED`? Thanks!

simonadomnisoru (commented):
Hey @eirikhaugstulen 
Looks really good overall 😄 ! I just added some suggestions for renaming some variables. Besides this, is it possible to add some cypress tests for this feature?
Thank you

karolinelien:
Looks good from a functional perspective! Couple of nitpicky comments:
Should we say "The category **option** is not valid for the selected organisation unit. Please select a valid option. 
(Maybe we can check with Matthew/Peter on what wording makes sense here)
image

Also, when I have a CO in my program stage event and then selects an invalid org unit in the context selector, the field just blanks without any message, could we add something, even one of those temporary messages on the bottom, saying that "(Coname) was removed as it is not a valid selection for (orgunitname)"?

eirikhaugstulen:
Thanks for the review both of you! Should have fixed all namings and comments now 👍

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.1.,2.40.5,2.39.7 versions

PR #3739, opened 2024-08-01
build: remove react-scripts 4.0.3 resolution

superskip:
This pull request upgrades Capture app to use the latest version of react-scripts (5.0.1).

The main motivation behind doing this upgrade is that the new kotlin rule engine has difficulties "importing named exports from non EcmaScript modules" when using react-scripts 4.0.3.

One side effect of upgrading react-scripts is that the `process` object becomes inaccessible to Capture, except one property `process.env`, which gets injected with an explicit value during compilation. The other uses of the `process` object had to be removed.

Another side effect was that it enabled the app to make use of package export aliases, but apparently also requiring it to do so. Consequently a few explicitly given paths in the app had to be modified.

The changes in `.eslintrc` and `.flowconfig` were made to fix a couple of linter and flow errors related to import paths.


simonadomnisoru (approved):
Great job! 🥳

PR #3748, opened 2024-08-07
fix: [DHIS2-17859] Add missing ids to Enrollment plugin

eirikhaugstulen:
We were not passing in programStageId and eventId to enrollment plugins as we should. 

Picture to validate that we're actually providing the correct ids:

1. Enrollment Edit event page 
image

2. Enrollment new event page 
image


simonadomnisoru (approved):
LGTM!

PR #3751, opened 2024-08-10
feat: [DHIS2-17575] disable create new buttons if no data write access

eirikhaugstulen:
Summary:
- Unify logic for create new event buttons in stages&events widget
- Disable quick actions button if there are no stages with write access
- Disable ProgramStageSelector-buttons if user does not have write access

### DIFF
--- a/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageCreateNewButton/StageCreateNewButton.js
+++ b/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageCreateNewButton/StageCreateNewButton.js
@@ -1,0 +1,72 @@
+// @flow
+import React, { useMemo } from 'react';
+import { Button, IconAdd16 } from '@dhis2/ui';
+import i18n from '@dhis2/d2-i18n';
+import { ConditionalTooltip } from '../../../../Tooltips/ConditionalTooltip';
+
+type Props = {
+    onCreateNew: (stageId: string) => void,
+    stageWriteAccess: ?boolean,
+    eventCount: number,
+    repeatable: ?boolean,
+    preventAddingEventActionInEffect: ?boolean,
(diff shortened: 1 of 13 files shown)


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageCreateNewButton/StageCreateNewButton.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const StageCreateNewButton = ({
+    onCreateNew,
+    stageWriteAccess,
+    eventCount,
+    repeatable,
+    preventAddingEventActionInEffect,
+    eventName,
+}: Props) => {
+    const isDisabled = useMemo(() => {
```
simonadomnisoru, line 24: Do the two useMemo perform the same checks? If so, how about combining them into only one hook? 
```
const { isDisabled , tooltipContent} = useMemo(() => { 
    ....
    ....
}, [eventCount, eventName, preventAddingEventActionInEffect, repeatable, stageWriteAccess]);
```
Thanks!

eirikhaugstulen:
Thanks for the review @simonadomnisoru! I also added some tests using the new trackerAutoTestRestricted 🎉

simonadomnisoru (approved):
LGTM

simonadomnisoru:
Hi @eirikhaugstulen,

This is unrelated to your changes, but if it's easy to fix, could you take a look at create new event buttons styles? When there are more than 5 events, the button is positioned wrong.



https://github.com/user-attachments/assets/e90f5b78-cb4d-4226-8d3f-f787605c8c7e


If the fix is ​​complex, I'll open a separate bug ticket for it. Thanks! 🙏

geethaalwan (approved):
Tested successfully on 2.42,  2.41.1,2.40.5,2.39.7 versions

PR #3783, opened 2024-09-03
fix: [DHIS2-17854] validate the assigned values from rules engine

simonadomnisoru:
[DHIS2-17854](https://dhis2.atlassian.net/browse/DHIS2-17854)

**Tech summary**
- extracted the [validateField](https://github.com/dhis2/capture-app/pull/3783/files#diff-34437e7d08816ce5e6cc3c54623ac10f394f1c42f131e84664f15a12902b3e7dL112) logic from the FormBuilder.component into an [external file](https://github.com/dhis2/capture-app/pull/3783/files#diff-9604d05d554914b16ae19f34c1724620a2384094db595bbfdddc22b6a5266b7cR12). 
- added a function, [validateAssignEffects](https://github.com/dhis2/capture-app/pull/3783/files#diff-f2adaf1577a1a66840375cefa7252dbc37922510f2fb71c4bc6cdadff85d93f2R16), that also calls `validateField`. The common function runs every time the rules engine finishes executing and before the assigned values are stored in Redux. To improve performance, only the last assigned value is validated, since it is the one stored in the redux `formsValues` key.
- `switchMap`, `from()` and `of()` were used in the epics to handle the async validation checks 
-  to prevent the user from clicking save/complete while an async validation is running:
     - In forms that rely on withSaveHandler.js, depending on the form logic, the actions `startLoadDataEntry`, `startRunRulesPostUpdateField`, and `startRunRulesPostLoadDataEntry` will be dispatched before running the validations. These actions will add an entry in `dataEntriesInProgressList` that will show the waitForPromisesDialogOpen modal if necessary.
    - In the WidgetProfile that does not rely on withSaveHandler.js, the Save button will be disabled

[DHIS2-17854]: https://dhis2.atlassian.net/browse/DHIS2-17854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
```diff
@@ -45,37 +45,36 @@ const getDataEntryId = (event): string => (
...
-        map((action) => {
+        switchMap((action) => {
             const state = store.value;
             const loadedValues = state.viewEventPage.loadedValues;
             const eventContainer = loadedValues.eventContainer;
             const metadataContainer = getProgramAndStageFromEvent(eventContainer.event);
             if (metadataContainer.error) {
-                return prerequisitesErrorLoadingEditEventDataEntry(metadataContainer.error);
+                prerequisitesErrorLoadingEditEventDataEntry(metadataContainer.error);
+                return EMPTY;
```
superskip, line 58: Why not use `return from(prerequisitesErrorLoadingEditEventDataEntry(metadataContainer.error))`?

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
```diff
@@ -205,26 +205,24 @@ export const formsSectionsFieldsUIDesc = createReducerDescription({
     [newPageActionTypes.CLEAN_UP_DATA_ENTRY]: cleanUp,
     [rulesEffectsActionTypes.UPDATE_RULES_EFFECTS]: (state, action) => {
         const { formId, rulesEffects } = action.payload;
-        const formSectionFields = state[formId];
```
superskip, line 208: I would assume `formSectionFields` was used here for a reason; what makes you think it can be removed?

src/core_modules/capture-core/components/D2Form/field/validators/validateAssignEffects.js
```diff
@@ -0,0 +1,63 @@
...
+        const lastEffect = effectsForId.length - 1;
+        const value = effectsForId[lastEffect].value;
+        const validators = getValidators(metaData, querySingleResource);
+        const validationContext = onGetValidationContext && onGetValidationContext();
+
+        try {
+            const validatorResult = await validateField({ validators }, value, validationContext);
+            const effectWithValidation = Object.assign({}, effectsForId[lastEffect], validatorResult);
+
+            acc[metaData.id] = [...effectsForId.slice(0, lastEffect - 1), effectWithValidation];
```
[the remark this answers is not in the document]
simonadomnisoru: Yes, I agree. It is now replaced with a one-element array

src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/editEventDataEntry.epics.js
simonadomnisoru: When using `from()`, the code throws a runtime error. This happens because `prerequisitesErrorLoadingEditEventDataEntry` returns a string and it is not what `from()` expects. Instead, I use `of()` to create an Observable, which works fine for this case.

src/core_modules/capture-core/reducers/descriptions/form.reducerDescription.js
simonadomnisoru: I shouldn’t have removed it, but now it's back in place. Thanks for the feedback!

superskip (approved):

JoakimSM (commented):
I have made a few structural suggestions that I hope is reasonable. Open to discuss if you disagree of course.

I'm also wondering, since we are dealing with potential async validations here, what are we doing to prevent the user from clicking save/complete while an async validation is running? I couldn't see this being handled properly, but I very much might have missed something.

src/core_modules/capture-core/components/DataEntries/Enrollment/actions/open.actionBatchs.js
```diff
@@ -101,13 +108,19 @@ export const openDataEntryForNewEnrollmentBatchAsync = async ({
         formFoundation,
     });
 
+    const effectsWithValidations = await validateAssignEffects({
```
JoakimSM, line 111: If I'm not mistaken, we are validating all the form values in the formbuilder when opening the form, so it would probably be best to skip the assign validations here (and other places where we are opening forms)

JoakimSM:
> > I have made a few structural suggestions that I hope is reasonable. Open to discuss if you disagree of course.
> 
> > I'm also wondering, since we are dealing with potential async validations here, what are we doing to prevent the user from > clicking save/complete while an async validation is running? I couldn't see this being handled properly, but I very much > might have missed something.
> 
> Hi @JoakimSM, I implemented the suggested structural changes. Regarding preventing the user from clicking save/complete while an async validation is running, in `withSaveHandler.js` there is already logic for this handled by the `waitForPromisesDialogOpen` modal. There were only a few forms where the actions to enable/disable the waitForPromisesDialogOpen modal were missing and I added them. For WdigetProfile, the only form that does not rely on withSaveHandler.js, I added logic to disable/enable the Save button.
> 
> Can you have a look? Thank you!

It does indeed seem like this is handled. Thanks.

src/core_modules/capture-core/components/DataEntries/Enrollment/actions/open.actionBatchs.js
simonadomnisoru: I skipped the assign validations when opening most of the forms. However, there are two scenarios where running the validations appears necessary: when opening the new event forms for both tracker and single events. In both cases, it seems that the execution of the [rules engine is intentionally delayed](https://github.com/dhis2/capture-app/blob/DHIS2-17854/src/core_modules/capture-core/components/WidgetEnrollmentEventNew/Validated/useLifecycle.js#L44-L105) until after the data entry is opened. This behavior was introduced in [TECH-635](https://dhis2.atlassian.net/browse/TECH-635).

JoakimSM (approved):
Nice 🎉  Looks good to me

geethaalwan (approved):
Tested successfully on 2.42,2.41.3,2.40.7,2.39.8 versions

PR #3788, opened 2024-09-05
fix: [DHIS2-18004] sort events in rules engine by occurredAt and createdAt

JoakimSM:
The events weren't sorted because the dates weren't in the right format. Ended up removing the validation in the end as it is not needed. Without the validation the old format would probably have worked, but for consistency, the values are now converted to the right "rules engine format".

Additionally, if two events occurred on the same date, they are now sorted by their creation date, ensuring the result is idempotent.


superskip (approved):
LGTM!

PR #3795, opened 2024-09-09
fix: [DHIS2-18028] User is able to add invalid related stages event

eirikhaugstulen:
Tech-summary:
- We were filtering out events that had a relationship because these are not considered linkable
- Switched this out to rather keep all events, but filter out the non-linkable as a separate array


henrikmv (commented):
I just found a `filter function` that I think is unnecessary. Apart from that, it looks very good.

simonadomnisoru (approved):
LGTM!

JoakimSM (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40.6,2.39.8 versions

PR #3814, opened 2024-09-26
feat: support custom background color

eirikhaugstulen:
Tech-summary:
- Support custom background color in our enrollment pages
- Add a "backgroundColor" prop to the datastore config
- This prop has to be a `#` followed by 6 characters

### DIFF
--- a/src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/EnrollmentPageLayout.js
+++ b/src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/EnrollmentPageLayout.js
@@ -10,6 +10,7 @@ import { DefaultPageTitle, EnrollmentPageKeys } from './DefaultEnrollmentLayout.
 
 const getEnrollmentPageStyles = () => ({
     container: {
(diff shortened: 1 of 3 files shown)


src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/EnrollmentPageLayout.js
```diff
@@ -10,6 +10,7 @@ import { DefaultPageTitle, EnrollmentPageKeys } from './DefaultEnrollmentLayout.
 
 const getEnrollmentPageStyles = () => ({
     container: {
+        minHeight: '90vh',
```
eirikhaugstulen, line 13: Open for ideas on this. The page only took up the size of the widgets, but that might not have been the full size. Also, App-platform adds a small padding at the bottom, so setting it to 90 instead of 100 means it will work as corrected.

JoakimSM: Seems to work and I don't have a better suggestion.

JoakimSM (approved):

PR #3823, opened 2024-10-01
fix: [DHIS2-18116] check return value of `parseDate`

superskip:
[DHIS2-18116](https://dhis2.atlassian.net/browse/DHIS2-18116)

The bug occurred because there was missing validation of the return value from `parseDate`.

There are other occurrences of parseDate (and parseTime) in the same file, but for some reason the obvious way to trigger these functions does not work, and therefore I don't know when or if they ever get called.

EDIT:
Regarding how to trigger the other functions, see [comment below](https://github.com/dhis2/capture-app/pull/3823#issuecomment-2388492705) (thanks @simonadomnisoru)

[DHIS2-18116]: https://dhis2.atlassian.net/browse/DHIS2-18116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/converters/formToClient.js
+++ b/src/core_modules/capture-core/converters/formToClient.js
@@ -14,16 +14,20 @@ type RangeValue = {
     to: string,
 }
 
-function convertDateTime(formValue: DateTimeValue): string {
+function convertDateTime(formValue: DateTimeValue): ?string {
     const editedDate = formValue.date;
     const editedTime = formValue.time;
 
-    const momentTime = parseTime(editedTime).momentTime;
+    const parsedTime = parseTime(editedTime);
+    if (!parsedTime.isValid) return null;
+    const momentTime = parsedTime.momentTime;
     const hours = momentTime.hour();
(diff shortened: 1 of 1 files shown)


simonadomnisoru:
> There are other occurrences of `parseDate` (and `parseTime`) in the same file, but for some reason the obvious way to trigger these functions does not work, and therefore I don't know when or if they ever get called.

Hi @superskip, 
From Maintenance, you can create a tracking entity attribute of type `DATETIME` and another of type `TIME`. Then, add them to the TB program attributes list. They should be visible in capture on the Create New Person registration page for the TB program. Typing a letter in these fields will call the other parseDate/parseTime occurrences of `convertDateTime` and `convertTime`.

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40.6,2.39.7 versions

PR #3827, opened 2024-10-04
fix: [DHIS2-16993] Text and long text DEs are missing arrows in changelog when deleted

henrikmv:
[DHIS2-16993:](https://dhis2.atlassian.net/browse/DHIS2-16993)

- adds a `span` around the arrow icon to prevent it from disappearing when space is limited.


simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.2 versions

PR #3834, opened 2024-10-08
feat: [DHIS2-16992] Fixed size for changelog modal and columns

henrikmv:
[DHIS2-16992:](https://dhis2.atlassian.net/browse/DHIS2-16992)

- Adds a fixed width to table on `750px` to fill the `Modal` component
- Columns have a fixed size (_The sizing for the different columns can easily be adjusted in the code if necessary, this is just a draft. Please feel free to add any sizing suggestions_)

Note: In addition to fixing the size, changes have been made to the changelog design. A label for `Updated` has been added, the coloring of the different label types has changed, and the labels are now placed in a separate column.


src/core_modules/capture-core/components/WidgetsChangelog/common/ChangelogTable/ChangelogCells/ChangelogChangeCell.js
```diff
@@ -0,0 +1,44 @@
...
+const ChangelogChangeComponent = ({ label, variant }: Config) => (
+    <Tag {...variant}>
+        {label}
+    </Tag>
+);
+
+
+export const ChangelogChangeCell = ({ changeType }: Object) => {
+    const config = changeTypeConfigs[changeType];
+    console.log('ChangelogChangeComponent', typeof changeType);
```
simonadomnisoru, line 33: forgotten `console`

simonadomnisoru (approved):
Great job, @henrikmv! It looks good from a developer's point of view. For this type of PR, I think it would be beneficial to get a review from Joe or Karoline as well 😄

cooper-joe:
- The fixed sizes are OK, the best we can do really. For some i18n cases they will not work, but changing numbers can't fix this.
- The current fixed width means that when native OS scrollbars are enabled, the table always has a horizontal scrollbar, see screenshot (even though it shouldn't need it). I guess the only way to avoid this is to reduce the fixed width?
image

simonadomnisoru (approved):

PR #3838, opened 2024-10-10
fix: [DHIS2-18215] selected working list not persisted on navigate

eirikhaugstulen:
Summary:
- Working list was not persisted when navigating back from the enrollment dashboard
- Check if there is a stored selected working list in redux before adding the default template

### DIFF
--- a/src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
+++ b/src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
@@ -1,7 +1,7 @@
 // @flow
-import React, { useEffect, useMemo, useCallback } from 'react';
+import React, { useCallback, useEffect, useMemo } from 'react';
 // $FlowFixMe
-import { connect, useSelector, shallowEqual, useDispatch } from 'react-redux';
+import { connect, shallowEqual, useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { programCollection } from 'capture-core/metaDataMemoryStores/programCollection/programCollection';
 import { MainPageComponent } from './MainPage.component';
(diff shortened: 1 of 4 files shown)


src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
```diff
@@ -120,6 +120,14 @@ const MainPageContainer = () => {
         showAllAccessible,
         categoryOptionIsInvalidForOrgUnit,
     });
+    const reduxSelectedTemplateId
+        = useSelector(({ workingListsTemplates }) => workingListsTemplates.teiList?.selectedTemplateId);
+    const workingListProgramId = useSelector(({ workingLists }) => workingLists
```
simonadomnisoru, line 125: I'm wondering if it's more sensible to get workingListProgramId from `workingListsContext.teiList?.programIdView`. I think the value is the same, it's more of a a matter of basing it on context rather than currentRequest. WDYT?

src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
```diff
@@ -120,6 +120,14 @@ const MainPageContainer = () => {
...
+        .teiList
+        ?.currentRequest
+        ?.queryParams
+        ?.program,
+    );
```
simonadomnisoru, line 130: Is it possible to calculate these values ​​along with the other selectors in `useSelectorMainPage`? Something like this:

```
const useSelectorMainPage = () =>
    useSelector(
        ({ currentSelections, activePage, workingListsTemplates, workingLists }) => ({
            reduxSelectedTemplateId: workingListsTemplates.teiList?.selectedTemplateId,
            workingListProgramId: workingLists.teiList?.currentRequest?.queryParams?.program,
            .....
        }),
        shallowEqual,
    );
```

src/core_modules/capture-core/components/Pages/MainPage/MainPage.container.js
eirikhaugstulen: Yes, makes sense!

eirikhaugstulen:
Thanks for the review @simonadomnisoru! Implemented the changes 👍

JoakimSM (approved):

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40.6,2.39.8 versions

PR #3848, opened 2024-10-17
fix: [DHIS2-18248] pass on createdAt to rules engine when editing

JoakimSM:
There was a missing piece in https://dhis2.atlassian.net/browse/DHIS2-18004. `createdAt` wasn't passed on the rules engine when editing an event.

FYI: I didn't convert the value from "server value" because the `DATE_TIME` value type conversion needs some work. I have created another ticket for this (https://dhis2.atlassian.net/browse/DHIS2-18249). This exact scenario should work without conversion though.

### DIFF
--- a/packages/rules-engine/src/services/VariableService/variableService.types.js
+++ b/packages/rules-engine/src/services/VariableService/variableService.types.js
@@ -27,6 +27,7 @@ type EventMain = {
     +status?: $Values<eventStatuses>,
     +occurredAt?: string,
     +scheduledAt?: string,
+    +createdAt?: string,
 };
 
 export type EventValues = {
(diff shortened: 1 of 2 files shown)


superskip:
Wow, was it really this simple? :open_mouth:
I was bracing myself for downloading and passing on to the rules engine from 10 different places (I didn't know you would be so kind to look into this)!
Definitely going to review this :+1:

superskip:
@JoakimSM Don't we need a similar thing when viewing an event?

JoakimSM:
> @JoakimSM Don't we need a similar thing when viewing an event?

Should work as far as I can see. Problem occurred when you were updating values in edit event (worked on initial execution)

superskip (approved):
:+1:

simonadomnisoru (approved):

PR #3849, opened 2024-10-17
feat: [DHIS2-18250] Breadcrumb for event & enrollment pages

eirikhaugstulen:
Tech-summary:
Added two new components that helps the user navigate our app

- EnrollmentBreadcrumb (Used on all enrollment pages
    - {programStageWorkingListName} / {workingListName} / {TET} list / {Search}
    - Enrollment dashboard
    - View event (if applicable)
    - Edit event (if applicable)
- EventBreadcrumb (Old view and edit event pages)
    - Event list
    - View event
    - Edit event (if applicable)


src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.component.js
```diff
@@ -73,6 +76,11 @@ export const EnrollmentEditEventPageComponent = ({
             pageLayout={pageLayout}
             currentPage={mode === EnrollmentPageKeys.EDIT_EVENT ? EnrollmentPageKeys.EDIT_EVENT : EnrollmentPageKeys.VIEW_EVENT}
             availableWidgets={WidgetsForEnrollmentEventEdit}
+            userInteractionInProgress={mode === EnrollmentPageKeys.EDIT_EVENT}
```
eirikhaugstulen, line 79: We should probably refactor this at some point, but just following the logic in `src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/TopBar.container.js`

src/core_modules/capture-core/components/Breadcrumbs/EventBreadcrumb/EventBreadcrumb.js
```diff
@@ -0,0 +1,105 @@
...
+                        classes={classes}
+                    />
+                    {index < (breadcrumbItems.length - 1) && (
+                        <IconChevronRight16 color={colors.grey800} />
+                    )}
+                </React.Fragment>
+            ))}
+
+            <DiscardDialog
+                open={openWarning === 'mainPage'}
```
simonadomnisoru, line 95: open={openWarning === pageKeys.MAIN_PAGE}

src/core_modules/capture-core/components/Breadcrumbs/EnrollmentBreadcrumb/hooks/useWorkingListLabel.js
```diff
@@ -0,0 +1,104 @@
...
+        }, {
+            enabled: !!selectedTemplateId &&
+                !DefaultFilterLabels[selectedTemplateId] &&
+                !isDefaultTemplate &&
+                isSameProgram,
+            onError: () => {
+                setShouldFetchPSFilter(true);
+            },
+        },
+    );
```
simonadomnisoru, line 67: I don't think it is necessary to call the API to get the WL names. If selectedTemplateId is stored in Redux, its name is also stored in Redux in the workingListsTemplates.teiList.templates array. Can you check it?

src/core_modules/capture-core/components/Breadcrumbs/EventBreadcrumb/hooks/useWorkingListLabel.js
```diff
@@ -0,0 +1,49 @@
...
+            resource: 'eventFilters',
+            id: selectedTemplateId,
+            params: {
+                fields: 'id,displayName',
+            },
+        },
+        {
+            enabled: !loading && !!selectedTemplateId && !isDefaultTemplate && isSameProgram,
+        },
+    );
```
simonadomnisoru, line 34: Same concern as mentioned here. If selectedTemplateId is stored in Redux, its name is also stored in Redux in the workingListsTemplates.eventList.templates array.

src/core_modules/capture-core/components/Breadcrumbs/EnrollmentBreadcrumb/EnrollmentBreadcrumb.js
```diff
@@ -0,0 +1,173 @@
...
+            condition: page === EnrollmentPageKeys.EDIT_EVENT,
+        },
+        {
+            key: 'newEvent',
+            onClick: () => {},
+            label: i18n.t('New event'),
+            selected: page === EnrollmentPageKeys.NEW_EVENT,
+            condition: page === EnrollmentPageKeys.NEW_EVENT,
+        },
+    ].filter(item => item.condition !== false)), [
```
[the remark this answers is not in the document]
eirikhaugstulen: Makes sense! Good catch!

src/core_modules/capture-core/components/Breadcrumbs/common/BreadcrumbItem.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import React, { type ComponentType } from 'react';
+import cx from 'classnames';
+import { withStyles } from '@material-ui/core/styles';
+import { colors } from '@dhis2/ui';
+
+type Props = {
+    label: string,
+    onClick: () => void,
+    selected: boolean,
```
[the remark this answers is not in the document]
eirikhaugstulen: The `classes` prop should be automatically added by `ComponentType<$Diff<Props, CssClasses>>` 👍

src/core_modules/capture-core/components/Breadcrumbs/EnrollmentBreadcrumb/EnrollmentBreadcrumb.js
```diff
@@ -0,0 +1,173 @@
...
+                handleNavigation(onBackToViewEvent, 'viewEvent');
+            },
+            label: i18n.t('View event'),
+            selected: page === EnrollmentPageKeys.VIEW_EVENT,
+            condition: page === EnrollmentPageKeys.VIEW_EVENT ||
+                (page === EnrollmentPageKeys.EDIT_EVENT && !eventIsScheduled(eventStatus)),
+        },
+        {
+            key: 'editEvent',
+            onClick: () => {},
```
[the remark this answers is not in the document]
eirikhaugstulen: Removing empty attributes usually makes sense, but in this case I've added an empty callback for two main reasons:
1. Flow - flow does not like to infer optional attributes on an object. This means I would have to write separate types for objects with or without this onClick. 
2. Removing the callback would mean we would have to do further validation to check if onClick really was a function further down the tree.

In this case, I  think it makes sense to just keep an empty callback to bypass both of them. Do you agree?

eirikhaugstulen:
Thanks for the review @simonadomnisoru & @henrikmv! Should be sorted now 🚀

simonadomnisoru (approved):
> Thanks for the review @simonadomnisoru & @henrikmv! Should be sorted now 🚀

Looks great now 👏 !

src/core_modules/capture-core/components/Breadcrumbs/EnrollmentBreadcrumb/hooks/useWorkingListLabel.js
```diff
@@ -0,0 +1,60 @@
...
+    cancelled: i18n.t('Cancelled enrollments'),
+};
+
+export const useWorkingListLabel = ({
+    programId,
+    trackedEntityName,
+    displayFrontPageList,
+}: Props) => {
+    const workingListTemplates = useSelector(({ workingListsTemplates }) => workingListsTemplates?.teiList);
+    const workingListProgramId = useSelector(({ workingListsContext }) => workingListsContext?.teiList?.programIdView);
```
henrikmv, line 24: For me it looks like the two useSelector calls are rendered 10 times each whenever navigating to the dashboard for a TEI. Should this behavior be optimized?"

eirikhaugstulen: It looks like `workingListProgramId` triggers one render and `workingListTemplates` triggers two. (One of each is batched, so they should only really trigger one rerender. Did you have any other experience, @henrikmv? Remember, if you want to see it the value triggers a rerender, you will have to wrap the `console.log` inside a useEffect.

eirikhaugstulen:
Thanks @henrikmv! Moving the issue to testing for now, but we can still continue discussing the rest of the comments 😊

henrikmv (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.3,2.40.6,2.39.8 versions

PR #3857, opened 2024-10-24
fix: [DHIS2-18228] Image Deleted on Update

henrikmv:
[DHIS2-18228:](https://dhis2.atlassian.net/browse/DHIS2-18228)

- **Why:** The bug occurred because the `value` containing the ID for the image was not stored in Redux. When converting to server value, `value` was set to null and the image was then deleted.

- **Fix:** Added `name` and `value` to the object stored in redux store


eirikhaugstulen (approved):
Makes sense! Good work!

src/core_modules/capture-core/events/getSubValues.js
```diff
@@ -37,23 +37,33 @@ const subValueGetterByElementType = {
...
-        (featureAvailable(FEATURES.trackerImageEndpoint) ?
-            {
-                url: `${absoluteApiPath}/tracker/events/${eventId}/dataValues/${metaElementId}/image`,
-                previewUrl: `${absoluteApiPath}/tracker/events/${eventId}/dataValues/${metaElementId}/image?dimension=small`,
-            } : {
-                url: `${absoluteApiPath}/events/files?dataElementUid=${metaElementId}&eventUid=${eventId}`,
-                previewUrl: `${absoluteApiPath}/events/files?dataElementUid=${metaElementId}&eventUid=${eventId}&dimension=SMALL`,
-            }
-        ),
+        querySingleResource({ resource: `fileResources/${value}` })
```
simonadomnisoru, line 52: Hi @henrikmv,
IMHO, there is no need to call `/fileResources` because as far as I know, the API for updating events only needs the `value` of DE/TEA and doesn't care about the `name`. We already know the `value`, so something like this should work: 
```
[dataElementTypes.IMAGE]: ({
        value,
        eventId,
        metaElementId,
        absoluteApiPath,
    }: {
        value: any,
        eventId: string,
        metaElementId: string,
        absoluteApiPath: string,
    }) =>
    (featureAvailable(FEATURES.trackerImageEndpoint) ?
            {
                value,
                url: `${absoluteApiPath}/tracker/events/${eventId}/dataValues/${metaElementId}/image`,
                previewUrl: `${absoluteApiPath}/tracker/events/${eventId}/dataValues/${metaElementId}/image?dimension=small`,
            } : {
                value,
                url: `${absoluteApiPath}/events/files?dataElementUid=${metaElementId}&eventUid=${eventId}`,
                previewUrl: `${absoluteApiPath}/events/files?dataElementUid=${metaElementId}&eventUid=${eventId}&dimension=SMALL`,
            }
        ), 
```
Can you re-check? Thank you!

henrikmv: Thanks, @simonadomnisoru, that's a good idea, I agree with you!

simonadomnisoru (approved):

PR #3859, opened 2024-10-25
fix: [DHIS2-17843] Disable delete enrollment button when user does not have authority

henrikmv:
[DHIS2-17843:](https://dhis2.atlassian.net/browse/DHIS2-17843)

-  Added a new permission check for cascade deletion of enrollments.
- Updating the `Delete` action to reflect this new permission and disable the button when the user do not have permission
- Added a tooltip that is shown when hovering the disabled button


src/core_modules/capture-core/components/WidgetEnrollment/hooks/useAuthorities.js
eirikhaugstulen, line 1: Do you think we can extract this as a generic hook and put it somewhere where it's accessible? For example: `src/core_modules/capture-core/utils`. I see that this is the third time we implement the same logic.

Usage:
```js
const { hasAuthority } = useAuthority({
    authority: 'F_ENROLLMENT_CASCADE_DELETE'
})
```

src/core_modules/capture-core/components/WidgetEnrollment/Actions/Delete/Delete.component.js
```diff
@@ -9,18 +10,21 @@ import {
...
 } from '@dhis2/ui';
-import React, { useState } from 'react';
 import i18n from '@dhis2/d2-i18n';
 import type { Props } from './delete.types';
+import { ConditionalTooltip } from '../../../Tooltips/ConditionalTooltip/';
 
-export const Delete = ({ enrollment, onDelete }: Props) => {
+export const Delete = ({ canCascadeDeleteEnrollment, enrollment, onDelete }: Props) => {
     const [toggle, setToggle] = useState(false);
+    const disabled = useMemo(() => !canCascadeDeleteEnrollment, [canCascadeDeleteEnrollment]);
```
eirikhaugstulen, line 19: Nit: I don't think we need to memoize this as it would just use up memory and the variable should be cached

eirikhaugstulen (commented):
Good work!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.3,2.40.6,2.39.8 versions

PR #3868, opened 2024-10-29
fix: [DHIS2-18238] incorrect criteria when saving list with follow up false

eirikhaugstulen:
Summary:
When creating a list with followUp explicitly marked as `false`, this got filtered out before it was sent to the API.

I also took the opportunity to add some ButtonStrips when already editing these files. I think it makes the UI look a bit cleaner.

### DIFF
--- a/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/teiTemplates.epics.js
+++ b/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/teiTemplates.epics.js
@@ -6,14 +6,14 @@ import { errorCreator } from 'capture-core-utils';
 import { ofType } from 'redux-observable';
 import { concatMap, filter, takeUntil } from 'rxjs/operators';
 import {
-    workingListsCommonActionTypes,
-    addTemplateSuccess,
     addTemplateError,
-    deleteTemplateSuccess,
+    addTemplateSuccess,
     deleteTemplateError,
-    updateTemplateSuccess,
-    updateTemplateError,
+    deleteTemplateSuccess,
     updateDefaultTemplate,
+    updateTemplateError,
+    updateTemplateSuccess,
+    workingListsCommonActionTypes,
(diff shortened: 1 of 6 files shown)


src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/teiTemplates.epics.js
```diff
@@ -55,7 +55,7 @@ export const addTEITemplateEpic = (action$: InputObservable, store: ReduxStore,
                     ...(assignedUserMode && { assignedUserMode }),
                     ...(assignedUsers?.length > 0 && { assignedUsers }),
                     ...(programStatus && { enrollmentStatus: programStatus }),
-                    ...(!!followUp && { followUp }),
+                    ...(followUp !== undefined && { followUp: JSON.stringify(followUp) }),
```
simonadomnisoru, line 58: I think the same solution should be applied when [updating](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/teiTemplates.epics.js#L179) a trackedEntity WL and when [creating](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/programStageTemplates.epics.js#L66) and [updating](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WorkingLists/TeiWorkingLists/epics/teiViewEpics/programStageTemplates.epics.js#L215) a programStage WL. WDYT?

eirikhaugstulen: I'm very glad you know this logic better than me! Good catch 😄

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.2,2.40,6,2.39.8 versions

PR #3905, opened 2024-12-09
feat: [DHIS2 15466] typing the date when editing enrollment and incident date

alaa-yahia:
Implement [DHIS2 15466](https://dhis2.atlassian.net/browse/DHIS2-15466)

**In Enrolment Widget:**
- Display and pick non-Gregorian dates. 
- Follow system setting format for dates.
- Display errors when date is invalid.
- Refactor code to use DateField component instead of using CalendarInput directly.


simonadomnisoru (approved):
LGTM

PR #3914, opened 2024-12-18
fix: [DHIS2-118654] Use importStrategy=DELETE to support older backend

ameenhere:
Fixes https://dhis2.atlassian.net/browse/DHIS2-18654 

The only area where importStrategy=delete was used, every other area used importStrategy=DELETE. 
This is not an issue for newer backend versions (probably v41+, but v39 breaks). Easier to fix frontend than backend.


simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.39.9 version

PR #3917, opened 2024-12-20
fix: [DHIS2-18632] Sorting stage detail table on orgunit breaks the app

henrikmv:


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/hooks/sortFunctions.js
```diff
@@ -78,7 +79,12 @@ const sortTime = (clientValueA: Object, clientValueB: Object, direction: string,
     return 0;
 };
 
-const sortOrgUnit = (clientValueA: Object, clientValueB: Object, direction: string, options: Object) => sortText(clientValueA.name, clientValueB.name, direction, options);
+const sortOrgUnit = (clientValueA: Object, clientValueB: Object, direction: string, options: Object) => {
```
simonadomnisoru, line 82: Now that orgUnitId is passed to the sortOrgUnit function, can we change the type from Object to `string`? Something like this: `clientValueA: string, clientValueB: string, ....`

henrikmv: Good catch, thank you! I have changed the type from `Object` to `string` now.

simonadomnisoru (approved):

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/hooks/sortFunctions.js
```diff
@@ -113,10 +119,11 @@ const sortStatus = (clientValueA: Object, clientValueB: Object, direction: strin
     return 0;
 };
 
-const sortDataFromEvent = ({ dataA, dataB, type, columnName, direction }: Object) => {
+export const sortDataFromEvent = ({ dataA, dataB, type, columnName, direction }: Object) => {
     if (!type) {
         log.error(errorCreator('Type is not defined')({ dataA, dataB }));
     }
+    console.log('direction', direction);
```
simonadomnisoru, line 126: Can you please remove this forgotten `console`? Other than that, the bug fix seems good to me.

simonadomnisoru (approved):

PR #3928, opened 2025-01-04
fix: [DHIS2-18582] Data element with number option set are not displayed in Stages and Events widget

henrikmv:
[DHIS2-18582:](https://dhis2.atlassian.net/browse/DHIS2-18582)

- Implements `useClientDataElements` hook to convert data elements to client value by formatting options based on type.
- Returns `options` as an array


simonadomnisoru (approved):
Good job! 🥳

geethaalwan (approved):

PR #3934, opened 2025-01-07
feat: [DHIS2-16782] Changelog sorting and filtering by user and data item

henrikmv:
[DHIS2-16782:](https://dhis2.atlassian.net/browse/DHIS2-16782)

- Implements sorting on `User` and `Data item`
- Implements filtering on `Data item` (For the attributes: DataElement and Field)


src/core_modules/capture-core/components/WidgetsChangelog/common/hooks/useChangelogData.js
```diff
@@ -32,22 +31,35 @@ export const useChangelogData = ({
         setPageSize(newPageSize);
     };
 
+    const filterParam =
+        filterValue !== 'Show all' && attributeToFilterBy
```
simonadomnisoru, line 35: The value sent from [ChangelogFilterBar](https://github.com/dhis2/capture-app/pull/3934/files#diff-63ed650bcb52cf6451cd08c69c511dcf301bc0757a0cae3f852e1b8de7bbd597R38) is `SHOW_ALL`, so this check does not work correctly. The const initialization on line [24](https://github.com/dhis2/capture-app/pull/3934/files#diff-ea0380fbff2dec66566ac55dc52502b84d62564113f532d6029fb0ea62186ab1R24) should probably be changed to `const [filterValue, setFilterValue] = useState('SHOW_ALL');`

simonadomnisoru (commented):
Hi @henrikmv, it looks good overall, I just added 2 small comments.

src/core_modules/capture-core/components/WidgetsChangelog/common/ChangelogFilterBar/DropdownFilter.js
```diff
@@ -0,0 +1,79 @@
...
+    label,
+    classes,
+    items,
+    filterColumn,
+    openMenuName,
+    onToggleMenu,
+    onItemSelected,
+    selectedFilterValue,
+}: Props) => {
+    const isMenuOpen = openMenuName === filterColumn;
```
simonadomnisoru, line 40: The menu does not close when filtering for the same value the second time. Could you have a look? Thanks!

https://github.com/user-attachments/assets/94645d70-a8be-4c63-94f8-31d84e07f93e

src/core_modules/capture-core/components/WidgetsChangelog/common/ChangelogFilterBar/ChangelogFilterBar.js
```diff
@@ -0,0 +1,74 @@
...
+import { FIELD_TYPES, FILTER_FIELD } from '../Changelog/Changelog.constants';
+
+const styles = {
+    container: {
+        padding: `0 0 ${spacers.dp8} 0`,
+    },
+};
+
+const getFilterField = (id: string) =>
+    (Object.values(FIELD_TYPES).includes(id) ? FILTER_FIELD.FIELD : FILTER_FIELD.DATA_ELEMENT);
```
simonadomnisoru, line 17: Currently, we are getting a 400 error when opening the changeLog from the trackedEntity Profile widget and trying to filter or sort by an attribute. I think we need to add `FILTER_FIELD.ATTRIBUTE` and `COLUMN_TO_SORT_BY.ATTRIBUTE` and logic to handle the trackedEntity change log.

https://github.com/user-attachments/assets/10f6b7dd-412e-430f-842b-df4423d0ab9d

src/core_modules/capture-core/components/WidgetsChangelog/common/ChangelogFilterBar/ChangelogFilterBar.js
```diff
@@ -0,0 +1,81 @@
...
+const styles = {
+    container: {
+        padding: `0 0 ${spacers.dp8} 0`,
+    },
+};
+
+const getFilterField = (id: string, entityType: string) => {
+    if (entityType === CHANGELOG_ENTITY_TYPES.TRACKED_ENTITY) {
+        return FILTER_TARGETS.ATTRIBUTE;
+    }
```
simonadomnisoru, line 19: In the Profile widget, the change log for a TET coordinate (both point and polygon) is missing. Looking at the epic, I think this was never implemented in the API. Do we want to add it?
The TET geometry appears in the filter options dropdown, but throws a 400 error when selected. The proper API request should probably look like this: `/tracker/trackedEntities/[trackedEntityId]/changeLogs?filter=field:eq:geometry`, but I don't think this type of filter is implemented in the API either. Any insights on this? 🤔 

Screenshot 2025-01-20 at 16 15 27

Thanks!

JoakimSM: TET-geometry will not be implemented at this point. If we want it, we will have to create a separate issue for it. So, we should remove TET geometry from the filter drop down.

src/core_modules/capture-core/components/WidgetsChangelog/common/ChangelogFilterBar/ChangelogFilterBar.js
```diff
@@ -0,0 +1,89 @@
...
+                    id: key,
+                    name: attribute.name,
+                };
+            })
+            : Object.keys(dataItemDefinitions).map(key => ({
+                id: key,
+                name: dataItemDefinitions[key].name,
+            }))),
+    [entityType, dataItemDefinitions, entityData],
+    );
```
simonadomnisoru, line 70: Hi @henrikmv, I think that TET geometry should not be inside `dataItemDefinitions`, so there would be no need to filter it here. My suggestion would be that when creating [dataItemDefinitions](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/TrackedEntityChangelogWrapper/TrackedEntityChangelogWrapper.component.js#L21) in TrackedEntityChangelogWrapper you add only the attributes. Something like this should work:

```
const fieldElementsById = elements.reduce((acc, element: DataElement) => {
  if (!transformedTrackedEntityData[element.id]) {
    return acc;
  }
  ....
```
WDYT? Thanks!

simonadomnisoru (approved):

simonadomnisoru (approved):

simonadomnisoru (approved):

PR #3938, opened 2025-01-15
fix: [DHIS2-18836] show 'Profile' instead of 'False' during loading

eirikhaugstulen:
Summary:
- Display correct title when profile widget is loading


simonadomnisoru (approved):
The fix LGTM! Could you link the PR to a Jira ticket? Thanks!

PR #3941, opened 2025-01-16
fix: [DHIS2-18668] remove delete button in enrollment event pages

eirikhaugstulen:
Summary:
- Added readOnlyMode prop to OverflowMenu component
- Hide the entire overflow menu when readOnlyMode is true and changelog is disabled
- Hide only the Delete action when readOnlyMode is true but changelog is enabled
- Verified readOnlyMode settings in DefaultPageLayout files:
    - true for add/edit event pages (to hide Delete)
    - false for enrollment dashboard (to show Delete)

### DIFF
--- a/src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.js
+++ b/src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.js
@@ -1,6 +1,6 @@
 // @flow
 import React, { useState } from 'react';
-import { FlyoutMenu, IconMore16, MenuItem, MenuDivider } from '@dhis2/ui';
+import { FlyoutMenu, IconMore16, MenuItem } from '@dhis2/ui';
 import i18n from '@dhis2/d2-i18n';
 import type { PlainProps } from './OverflowMenu.types';
 import { DeleteMenuItem, DeleteModal } from './Delete';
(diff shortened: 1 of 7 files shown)


src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.js
```diff
@@ -45,13 +50,15 @@ export const OverflowMenuComponent = ({
                                 <MenuDivider dense />
```
simonadomnisoru, line 50: What do you think about moving `MenuDivider` below along with `DeleteMenuItem`? It looks a bit odd right now when only the changelog is enabled.
Screenshot 2025-01-16 at 08 43 32

 We can also completely remove the MenuDivider as it is in the StagesAndEvents widget, which IMHO looks cleaner.
Screenshot 2025-01-16 at 08 49 32

Thanks!

eirikhaugstulen: Good catch, @simonadomnisoru! I just removed it, I think that looks pretty clean.

simonadomnisoru (approved):
Great job 👏 ! Could you please remove the Cypress scenario “You can delete a tracked entity from the profile widget” from the edit and new event pages before moving to QA? Thanks!

geethaalwan (approved):

PR #3950, opened 2025-01-27
fix: [DHIS2-18902] keep program selection on unique id fallback search

superskip:
[DHIS2-18902](https://dhis2.atlassian.net/browse/DHIS2-18902)

This should cover the main case. There are two other similar search components which are used in the context of relationships, but I think these do not lead to opening a tracked entity, and are therefore not subject to the reported issue.

[DHIS2-18902]: https://dhis2.atlassian.net/browse/DHIS2-18902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):
LGTM! 🎉

geethaalwan (approved):

PR #3965, opened 2025-02-11
fix: [DHIS2-18958] Add loading functionality to Search in all programs button

henrikmv:
[DHIS2-18958](https://dhis2.atlassian.net/browse/DHIS2-18958):

When the button is clicked, it enters a loading state. It will remain in this state until the Redux variable `otherResults` is defined. This ensures that the button stays in a loading state while the search is in progress and updates only when results are available.

A loading spinner is also implemented when navigating between pages while searching across all programs.

[DHIS2-18958]: https://dhis2.atlassian.net/browse/DHIS2-18958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/SearchBox/SearchResults/SearchResults.component.js
+++ b/src/core_modules/capture-core/components/SearchBox/SearchResults/SearchResults.component.js
@@ -8,7 +8,7 @@ import React, {
 import { withStyles } from '@material-ui/core';
 import i18n from '@dhis2/d2-i18n';
 import { Pagination } from 'capture-ui';
-import { Button, colors } from '@dhis2/ui';
+import { Button, CircularLoader, colors } from '@dhis2/ui';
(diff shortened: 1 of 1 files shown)


henrikmv:
> Hi @henrikmv, The `Search all programs` button loading functionality looks great 👏 . I just was wondering if we should add a loading spinner when changing pages too (just like we do when searching in a specific program)? This is not in the Jira ticket but it feels related and may also be an easy fix. Any thoughts? Thanks!

Thanks Simona! That is a good suggestion and has now been implemented.

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42,2.41.4,2.40.8,2.39.9 versions

PR #3982, opened 2025-02-24
feat: [DHIS2-18944] add effect to refresh list on event deletion

eirikhaugstulen:
Added a useEffect that listens to the last deleted ID and refreshes the app when it's updated. Currently, the list does not refresh on delete.


simonadomnisoru (approved):
Good job!

geethaalwan (approved):
Tested successfully on 2.42,2.41.4,2.40.8,2.39.10 versions

PR #3989, opened 2025-03-06
Working lists perf

JoakimSM:


src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useWorkingListsCommonStateManagement.js
```diff
@@ -278,15 +283,25 @@ const useWorkingListsContext = (dispatch: ReduxDispatch, { storeId }: { storeId:
     };
 };
 
-export const useWorkingListsCommonStateManagement = ((storeId: string, workingListsType: string, program: Program) => {
+export const useWorkingListsCommonStateManagement = (
+    storeId: string,
+    workingListsType: string,
+    program: Program,
+    mainViewConfig?: Object,
```
simonadomnisoru, line 290: NIT `mainViewConfig?: MainViewConfig`

src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useWorkingListsCommonStateManagement.js
```diff
@@ -60,7 +60,7 @@ import type {
 
 const useTemplates = (
     dispatch: ReduxDispatch,
-    { storeId, workingListsType }: { storeId: string, workingListsType: string }) => {
+    { storeId, workingListsType, mainViewConfig }: { storeId: string, workingListsType: string, mainViewConfig?: Object}) => {
```
simonadomnisoru, line 63: NIT: `mainViewConfig?: MainViewConfig` and `import type { MainViewConfig } from '../../EventWorkingLists';` at the beginning of the file

simonadomnisoru (approved):

src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useWorkingListsCommonStateManagement.js
JoakimSM: Problem is that since we are in WorkingListsCommon, this file is not only be used by event programs (at least, that is the intention - and I believe it is the case). So, mainViewConfig could vary here depending on the context. From an architectural perspective, WorkingListsCommon shouldn't import from EventWorkingLists because it shouldn't be aware of EventWorkingLists' existence (EventWorkingLists depends on WorkingListsCommon, not the other way around). The proper solution here would be to use Generics. However, since we're missing types here for many other variables for the same reason, and I was concerned about running into "FlowFixMe", I opted out. Would be great to add types, but maybe we do it another ticket (or after migrating to Typescript 😰).

src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useWorkingListsCommonStateManagement.js
JoakimSM: https://github.com/dhis2/capture-app/pull/3989/files#r1987665579

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.42.0,2.41.4,2.40.8,2.39.10 versions

PR #3995, opened 2025-03-10
fix: [DHIS2-19165] handle empty schedule date

JoakimSM:
1. Don't save a scheduled date when directly adding a tracker event (from version 41). 
2. Handle empty value for `scheduledAt` when editing tracker event.

### DIFF
--- a/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js
+++ b/src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js
@@ -189,12 +189,12 @@ const buildScheduleDateSettingsFn = () => {
             calendarWidth: 350,
             label: props.formFoundation.getLabel('scheduledAt'),
             disabled: true,
+            required: false,
             calendarType: systemSettingsStore.get().calendar,
             dateFormat: systemSettingsStore.get().dateFormat,
         }),
         getIsHidden: (props: Object) => props.id !== dataEntryIds.ENROLLMENT_EVENT || props.hideDueDate,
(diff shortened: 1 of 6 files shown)


src/core_modules/capture-core/components/WidgetEventEdit/EditEventDataEntry/EditEventDataEntry.component.js
```diff
@@ -189,12 +189,12 @@ const buildScheduleDateSettingsFn = () => {
             calendarWidth: 350,
             label: props.formFoundation.getLabel('scheduledAt'),
             disabled: true,
+            required: false,
```
simonadomnisoru, line 192: Currently, the "Enter details now" case of the "Related Stages" feature assumes that the event-to-linked-from always has a scheduledAt date. Perhaps we could use the `occurredAt` date if the `scheduledAt` is empty to generate the linked event payload? WDYT?
The file to look into is [getConvertedRelatedStageEvent](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/DataEntries/converters/getConvertedRelatedStageEvent/getConvertedRelatedStageEvent.js#L71).

JoakimSM: Yes, I realised there were more cases to handle here (hence, why I converted it to a draft PR). I think every workflow is handled now, including the one you mentioned here (I tested all the workflows I could think of).

simonadomnisoru (approved):
LGTM!

geethaalwan (approved):
Tested successfully on 2.42,2.41.4 versions

PR #4001, opened 2025-03-14
chore: Remove leaflet CDN dependencies

jason-p-pickering:
These should already be packaged as part of the app build, so no reason to include them separately via a CDN.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -49,10 +49,6 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>%REACT_APP_DHIS2_APP_NAME% | DHIS2</title>
-    <!-- leaflet -->
-    <link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.3/leaflet.css'>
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -49,10 +49,6 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>%REACT_APP_DHIS2_APP_NAME% | DHIS2</title>
-    <!-- leaflet -->
-    <link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.3/leaflet.css'>
-    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet.draw/0.4.2/leaflet.draw.css"/>
```
simonadomnisoru, line 54: The map is, as far as I know, the only component that uses `leaflet`. With these changes, the component appears to be broken. Before removing these CDN dependencies, we need to make sure everything still works, both when capturing coordinates and when capturing a polygon area.
leaflet
Thanks!

jason-p-pickering: This was a bit of a speculative PR to see if it works, but the better solution would just be to package the dependency along with the app if it is not already?

PR #4002, opened 2025-03-15
fix: [DHIS2-19221] enrollment dashboard workflows

JoakimSM:
See https://dhis2.atlassian.net/browse/DHIS2-19221


simonadomnisoru (approved):
LGTM!

PR #4016, opened 2025-03-21
fix: improve error handling metadata

JoakimSM:
Handle event program without stages. These programs will just be ignored (with an error message in the console) instead of breaking the app.


eirikhaugstulen (approved):
LGTM!

simonadomnisoru (approved):

PR #4043, opened 2025-04-03
fix: [DHIS2-18951] Ignore Non-searchable attributes error when validating unique attribute

henrikmv:
[DHIS2-18951](https://dhis2.atlassian.net/browse/DHIS2-18951):

Temporary fix for error message:
`"Non-searchable attribute(s) can not be used during global search: [attributeId]"`

This change catches that specific `409` error in validation and allows the form to be saved instead of blocking the user.
Note: The error still appears in the network tab but no longer prevents saving.

[DHIS2-18951]: https://dhis2.atlassian.net/browse/DHIS2-18951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JoakimSM (approved):
Looks good @henrikmv

geethaalwan (approved):
Tested successfully on 2.43,2.42.1,2.41.5,2.40.9 versions

PR #4044, opened 2025-04-03
fix: [DHIS2-19415] editing two fields simultaneously blocks saving

superskip:
[DHIS2-19415](https://dhis2.atlassian.net/browse/DHIS2-19415)

Replaces `switchMap` with `concatMap` so that no actions get discarded by accident.

[DHIS2-19415]: https://dhis2.atlassian.net/browse/DHIS2-19415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/DataEntries/Enrollment/epics/enrollment.epics.js
+++ b/src/core_modules/capture-core/components/DataEntries/Enrollment/epics/enrollment.epics.js
@@ -2,7 +2,7 @@
 import type { OrgUnit } from '@dhis2/rules-engine-javascript';
 import { ofType } from 'redux-observable';
 import { from } from 'rxjs';
-import { map, switchMap } from 'rxjs/operators';
+import { map, concatMap } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


superskip:
@JoakimSM Thanks for checking out the PR! In my understanding, `mergeMap` creates a queue of observables, each which are run to the completion before moving on the observable in the queue. So the actions will be processed to completion in the same order they as they arrive the epic. There could be a race going on before arriving the epic though, do you think we should look into that?

JoakimSM:
@superskip Ah, you are right (I have forgotten how concatMap works apparently). I think we should be good then!

JoakimSM (approved):

simonadomnisoru (approved):
Good job @superskip!

geethaalwan (approved):
Tested successfully on 2.43 ,2.42.1,2.42.0,2.41.4,2.40.8,2.39.10 versions

PR #4045, opened 2025-04-04
fix: [DHIS2-19011] Forms retain values after search

henrikmv:
[DHIS2-19011](https://dhis2.atlassian.net/browse/DHIS2-19011):

### 📝 Prepopulation of search values into New TEI forms

1. When the user clicks the **Create new** button, the current search values from the `searchDomain` state in Redux are temporarily stored in the `newPage` state.  
2. Once the **New TEI form** opens, those values are populated into the **formValues** state (if any values exist) using the new `setPrepopulateDataOnNewPage` action.  
3. After transferring the search values to the **formValues** state, they are immediately cleared from `newPage` using the new `clearPrepopulatedData` action.

---

### 🔧 Additional changes

- Functionality related to the `fallbackTriggered` variable had already been removed. This PR includes additional cleanup, removing the remaining references and logic tied to `fallbackTriggered`.  
- The `useFormValuesFromSearchTerms` hook uses local state to store the search terms after they’ve been cleared from the `newPage` state. This ensures the user can still switch Tracked Entity types within the **Create new TEI** form when a program is not selected.

[DHIS2-19011]: https://dhis2.atlassian.net/browse/DHIS2-19011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (commented):
Hi @henrikmv,

The `fallbackTriggered` clean up looks great 👍 .

Regarding the dispatch of the `cleanSearchRelatedData` action, although it's triggered in several epics on both the /new and /search pages, it doesn’t account for all navigation scenarios (one example is when using the browser’s back button). Instead of trying to clean up on every possible navigation case, I propose we reverse the logic: clean the search values by default when the component is removed from the DOM, and only keep them when explicitly needed. In practice, this would mean:

**New page**
- dispatch `cleanSearchRelatedData` in the clean-up function of this [useEffect](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/Pages/New/RegistrationDataEntry/RegistrationDataEntry.container.js#L49-L53). 


**Search page** 
- dispatch `cleanSearchRelatedData` in the clean-up function of this [useEffect](https://github.com/dhis2/capture-app/pull/4045/files#diff-446c4c42d7cde6989d742061ff3a45af12e2a36120342d1a77a509417a3fe813L86-L91).
- the only case where we need to preserve the search values after unmounting is when the user clicks the ‘Create new’ button. We can flag this using the `handleCreateNew` callback and a `useRef`, and then skip dispatching `cleanSearchRelatedData` in the cleanup accordingly.
Screenshot 2025-04-08 at 12 10 11



What are your thoughts on this approach? I'm available for a call next week if any clarification is needed.
Thank you!

eirikhaugstulen (approved):
I won't say that I fully understand everything that goes on between the search and new page, but looking at your changes and testing it out locally, this looks very elegant! Nice work @henrikmv 🎉 (Two small type things to look into, but still approving)

simonadomnisoru (approved):
LGTM!

geethaalwan (approved):
Tested successfully on 2.43,2.42.1,2.41.4,2.40.8 versions

PR #4059, opened 2025-04-25
fix: [DHIS2-18671] Discard changes dialog not responding

henrikmv:
[DHIS2-18671](https://dhis2.atlassian.net/browse/DHIS2-18671):

### 🔄 Refreshing the New TEI form

1. A `newPageKey` state was introduced using `useState` and `uuid` to generate a unique key.  
2. The `handleRefreshNewTeForm` function updates this key, which forces the form to reload.  
3. The refresh functionality is integrated into the top bar, allowing users to trigger a form reset from there.

[DHIS2-18671]: https://dhis2.atlassian.net/browse/DHIS2-18671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/Pages/New/NewPage.container.js
+++ b/src/core_modules/capture-core/components/Pages/New/NewPage.container.js
@@ -1,8 +1,7 @@
 // @flow
 import { useDispatch, useSelector } from 'react-redux';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback } from 'react';
 import type { ComponentType } from 'react';
-import { v4 as uuid } from 'uuid';
 import { NewPageComponent } from './NewPage.component';
 import {
     showMessageToSelectOrgUnitOnNewPage,
@@ -109,11 +108,6 @@ export const NewPage: ComponentType<{||}> = () => {
           || dataEntryHasChanges(state, 'newPageDataEntryId-newTei'),
     );
 
-    const [newPageKey, setNewPageKey] = useState();
-    const handleRefreshNewTeForm = () => {
-        setNewPageKey(uuid());
-    };
-
     return (
         <>
             <TopBar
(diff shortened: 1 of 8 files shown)


src/core_modules/capture-core/components/Pages/New/NewPage.container.js
```diff
@@ -108,6 +109,11 @@ export const NewPage: ComponentType<{||}> = () => {
           || dataEntryHasChanges(state, 'newPageDataEntryId-newTei'),
     );
 
+    const [newPageKey, setNewPageKey] = useState();
+    const handleRefreshNewTeForm = () => {
+        setNewPageKey(uuid());
+    };
+
```
simonadomnisoru, line 116: White a useEffect that triggeres each time this component is rendered

src/core_modules/capture-core/components/Pages/New/NewPage.container.js
```diff
@@ -108,6 +109,11 @@ export const NewPage: ComponentType<{||}> = () => {
           || dataEntryHasChanges(state, 'newPageDataEntryId-newTei'),
     );
 
+    const [newPageKey, setNewPageKey] = useState();
+    const handleRefreshNewTeForm = () => {
+        setNewPageKey(uuid());
+    };
```
simonadomnisoru, line 115: Hi @henrikmv, I believe the issue is that [history.push](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/utils/routing/useNavigate.js#L8) doesn't have any effect when the path is the same (there's a warning in the console). However, when I use `history.replace` along with a useEffect like in [this example](https://jasonwatmore.com/react-router-v6-listen-to-location-route-change-without-history-listen), the form reloads properly. Any thoughts?

Screenshot 2025-04-28 at 10 59 33

Thanks!

geethaalwan (approved):
Tested successfully on 2.43,2.42.1,2.41.4,2.40.8,2.39.10 versions

PR #4074, opened 2025-05-07
chore: [DHIS2-17635] reactive and improve note Cypress tests

simonadomnisoru:
[DHIS2-17635](https://dhis2.atlassian.net/browse/DHIS2-17635)

**Tech summary**
- The API doesn't allow deleting or editing notes. This makes it impossible to clean up the DB after each scenario. As a workaround, I added timestamps to differentiate between different test runs.


henrikmv (approved):
Great work!

PR #4075, opened 2025-05-08
fix: [DHIS2-19330] data element increments twice on render

eirikhaugstulen:
Summary:
- Attributes marked as "Incremental" increments with two for each render

Thanks for the detailed bug report and fix @MiquelAdell & @anagperal 🎉

https://dhis2.atlassian.net/browse/DHIS2-19330

### DIFF
--- a/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/factory/enrollment/EnrollmentFactory.js
+++ b/src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/factory/enrollment/EnrollmentFactory.js
@@ -310,7 +310,7 @@ export class EnrollmentFactory {
                     });
                 } else if (cachedProgramSections) {
                     // $FlowFixMe
-                    cachedProgramSections.asyncForEach(async (programSection) => {
+                    await cachedProgramSections.asyncForEach(async (programSection) => {
                         section = await this._buildSection(
                             programSection.trackedEntityAttributes.map(id => trackedEntityAttributeDictionary[id]),
                             programSection.displayFormName,


simonadomnisoru (commented):
The fix looks good @eirikhaugstulen! 
Do you think it makes sense to also add `await` in the other two places [mentioned](https://github.com/EyeSeeTea/capture-app/pull/2#discussion_r2014150265) (`EnrollmentFactory.js#L269` and `TeiRegistrationFactory.js#L105`)?
Thanks!

src/core_modules/capture-core/metaDataMemoryStoreBuilders/programs/factory/enrollment/EnrollmentFactory.js
```diff
@@ -310,7 +310,7 @@ export class EnrollmentFactory {
                     });
                 } else if (cachedProgramSections) {
                     // $FlowFixMe
-                    cachedProgramSections.asyncForEach(async (programSection) => {
+                    await cachedProgramSections.asyncForEach(async (programSection) => {
```
simonadomnisoru, line 313: The fix looks good @eirikhaugstulen!
Do you think it makes sense to also add await in the other two places https://github.com/EyeSeeTea/capture-app/pull/2#discussion_r2014150265 (EnrollmentFactory.js#L269 and TeiRegistrationFactory.js#L105)?
Thanks!

superskip: Yes, we should add `await` for those two as well (I double checked that those are the only other places in the code where we call `asyncForEach` without `await`)

simonadomnisoru (approved):
LGTM!

PR #4078, opened 2025-05-12
fix: [DHIS2-11847] Add critical alert to event post errors

henrikmv:
[DHIS2-11847](https://dhis2.atlassian.net/browse/DHIS2-11847)

### DIFF
--- a/src/core_modules/capture-core/reducers/descriptions/feedback.reducerDescriptionGetter.js
+++ b/src/core_modules/capture-core/reducers/descriptions/feedback.reducerDescriptionGetter.js
@@ -28,102 +28,168 @@ import { enrollmentSiteActionTypes } from '../../components/Pages/common/Enrollm
 import { enrollmentEditEventActionTypes } from '../../components/Pages/EnrollmentEditEvent';
 import { actionTypes as viewEventActionTypes } from '../../components/Pages/ViewEvent/ViewEventComponent/viewEvent.actions';
 
-function addErrorFeedback(state: ReduxState, message: string, action?: ?Node) {
-    const newState = [...state];
-    newState.push({
-        message,
-        action,
-        feedbackType: 'ERROR',
-    });
-    return newState;
+const alertVariants = {
+    info: 'info',
+    success: 'success',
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/reducers/descriptions/feedback.reducerDescriptionGetter.js
```diff
@@ -28,21 +29,37 @@ import { enrollmentSiteActionTypes } from '../../components/Pages/common/Enrollm
...
+};
+
+function addErrorFeedback(state: ReduxState, { message, variant, action }: ErrorFeedbackInput) {
+    return [
+        ...state,
+        {
+            message,
+            action,
+            feedbackType: 'ERROR',
+            variant,
```
simonadomnisoru, line 52: Hi @henrikmv, 
I'm curious what the criteria were for determining when the critical variant of `addErrorFeedback` should be applied. It seems to me that most, if not all, of the reducers in this file could benefit from the `alertVariants.critical`, as I believe they are all errors originating from either POST, PUT, or DELETE requests. If so, perhaps it could be the default value. WDYT?
Thanks!

karolinelien: I am very open to some discussion here! Based on a list with error messages @henrikmv shared, I suggested that (at least) the ones where saving is failing, we should have the critical bar.

I was not as sure on "less critical" error where you are not able to delete a working list or you are trying to add a relationship that already exists. But maybe these criterias end up being very subjective and it would be easier if we introduce a general approach. Or we can do these ones now, then check with @cooper-joe when he is back since this could benefit from a UI design perspective. Does @JoakimSM have a preference?

henrikmv: @simonadomnisoru @karolinelien: After talking to @JoakimSM, I will set critical as the default value.

geethaalwan (approved):
Tested successfully on 2.43,2.42.1,2.41.5,2.40.9 versions

PR #4080, opened 2025-05-13
refactor: [DHIS2-19604] create special section for data elements not assigned to a section

superskip:
In program stages using **section form**, it is not required for all data elements to be part of the form.

However, any data element not included in the form should still be assignable by means of program rules. Now it turns out that for the app to be able to do this, all such data elements need to be found within the `ProgramStage`s coming from `ProgramStageFactory`.

The problem with the initial implementation of `ProgramStageFactory` is that it just ignored all data elements not assigned to any section, meaning such data elements could not be recovered from the manufactured `ProgramStage`.

The first commit reverts one possible solution to this problem, which is to add all program stage data element in an array separate from the `ProgramStage.stageForm`-object (`ProgramStage.stageForm` is where the data elements used to get extracted from). The main drawback with this approach is that most data elements get saved twice, once in the `ProgramStage.stageForm` and once in the separate array.

This new approach is to include all unassigned data elements to a new special section in `RenderFoundation`. The primary advantage of doing it this way is that it eliminates the duplication of the first approach, and it simplifies the structure of `ProgramStage`. The main disadvantage is that it introduces a special kind of section that should never be displayed in the form, and therefore needs special treatment in the code. Implementing this special behavior turned out to be quite easy though (third commit).

### DIFF
--- a/src/core_modules/capture-core/rules/getDataElementsForRulesExecution.js
+++ b/src/core_modules/capture-core/rules/getDataElementsForRulesExecution.js
@@ -3,8 +3,13 @@ import type { ProgramStage } from '../metaData';
 
 export const getDataElementsForRulesExecution = (stages: Map<string, ProgramStage>) =>
     [...stages.values()]
-        .flatMap(stage => stage.dataElements)
+        .flatMap(stage => stage.stageForm.getElements())
         .reduce((accRulesDataElements, dataElement) => {
-            accRulesDataElements[dataElement.id] = dataElement;
+            accRulesDataElements[dataElement.id] = {
+                id: dataElement.id,
+                valueType: dataElement.type,
+                optionSetId: dataElement.optionSet?.id,
+                name: dataElement.formName || dataElement.name,
+            };
             return accRulesDataElements;
         }, {});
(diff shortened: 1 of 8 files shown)


src/core_modules/capture-core/rules/getDataElementsForRulesExecution.js
```diff
@@ -3,8 +3,13 @@ import type { ProgramStage } from '../metaData';
...
+        .flatMap(stage => stage.stageForm.getElements())
         .reduce((accRulesDataElements, dataElement) => {
-            accRulesDataElements[dataElement.id] = dataElement;
+            accRulesDataElements[dataElement.id] = {
+                id: dataElement.id,
+                valueType: dataElement.type,
+                optionSetId: dataElement.optionSet?.id,
+                name: dataElement.formName || dataElement.name,
+            };
             return accRulesDataElements;
```
simonadomnisoru, line 14: Should any part of this fix also be applied to the [getDataElementsForRulesExecution](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetProfile/DataEntry/ProgramRules/getRulesActionsForTEI.js#L39-L51) function in `WidgetProfile`? Besides this doubt, the new approach looks good to me.

superskip: I had a look at the profile widget now, and it looks fine because it gets the data elements directly from an api response (hence not affected by this refactor of `ProgramStage`).

geethaalwan (approved):
Tested successfully on 2.43 version

PR #4132, opened 2025-06-05
perf: [DHIS2-19724] optimize api-calls on enrollment dashboard

superskip:
[DHIS2-19724](https://dhis2.atlassian.net/browse/DHIS2-19724)

Of the six requests to the `trackedEntities` endpoint I have removed one (in `autoSwitchOrgUnitEpic`) and limited the amount of requested data in three.

Two of the remaining five requests come from the enrollment widget and the profile widget and can't be removed.

Another two happens during initialization of the enrollment dashboard (one for fetching the tracked entity and one for fetching enrollments). They are done separately for error reporting purposes; the tracked entity request checks if the TE exists, while the enrollment request checks if the user has access to a given program.

The last request comes from `useCommonEnrollmentDomainData`, which caches a single enrollment for shared use across all sub-pages of the enrollment dashboard. This might still be a potential candidate for optimization.

[DHIS2-19724]: https://dhis2.atlassian.net/browse/DHIS2-19724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.ts
+++ b/src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPage.actions.ts
@@ -115,7 +115,7 @@ export const verifyFetchedEnrollments = ({ teiId, programId, action }: { teiId: 
 export const fetchEnrollmentsError = ({ accessLevel }: { accessLevel: string }) =>
     actionCreator(enrollmentPageActionTypes.FETCH_ENROLLMENTS_ERROR)({ accessLevel });
 
-export const saveEnrollments = ({ enrollments, programOwnerId }: { enrollments: Object, programOwnerId: ?string }) =>
+export const saveEnrollments = ({ enrollments, programOwnerId }: { enrollments: any, programOwnerId: string | null }) =>
     actionCreator(enrollmentPageActionTypes.FETCH_ENROLLMENTS_SUCCESS)({ enrollments, programOwnerId });
 
 // Page status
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/components/WidgetEnrollment/hooks/useTrackedEntities.js
```diff
@@ -2,15 +2,16 @@
 import { useMemo } from 'react';
 import { useDataQuery } from '@dhis2/app-runtime';
 
-export const useTrackedEntityInstances = (teiId: string, programId: string) => {
+export const useTrackedEntities = (teiId: string, programId: string) => {
     const { error, loading, data, refetch } = useDataQuery(
```
simonadomnisoru, line 6: Can we rely on the @`dhis2/app-runtime` hook to cache this request? I know it uses React Query under the hood, but I'm not entirely sure what its default caching strategy is, or if we need to pass specific parameters to enable caching. An alternative could be to use React Query directly, so we have full control over the caching behavior. Same doubt about the hook used in [WidgetProfile](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetProfile/hooks/useTrackedEntityInstances.js#L25-L38). Thanks!

superskip: You're right, this is possibly a source of improvement. I'll look into it :+1:

superskip: @simonadomnisoru I found that `dhis2/app-runtime` is overriding React Query's default values for refetching [here](https://github.com/dhis2/app-runtime/blob/master/services/data/src/react/components/DataProvider.tsx). Looks to me like automatic fetching only occurs when the component mounts. This is certainly the least expensive configuration, so I don't think we will gain anything performance-wise by using React Query directly. Please speak up if you have any objections to this conclusion : )

PR #4157, opened 2025-06-15
feat: [DHIS2-7080] Expired Period Validation for Edit Events

henrikmv:
[DHIS2-7080](https://dhis2.atlassian.net/browse/DHIS2-7080)

## Functionality

- The **Edit** button is disabled for both **Tracker** and **Event** programs if the event falls within an **expired period**.
- The **Delete** button in the event list is also disabled for both program types when the event is in an expired period.

## Date Handling

- The date (`occurredAt` / report date) passed to the `isValidPeriod` function is always in the **ISO format** (`yyyy-mm-dd`) to ensure compatibility with various date formats and calendar systems.
- Dates displayed in tooltips and error messages are formatted using `clientToView` to reflect the correct **format and calendar system** for the user. The dates that are not converted with the `clientToView` should already be in the proper format.

[DHIS2-7080]: https://dhis2.atlassian.net/browse/DHIS2-7080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (commented):
Hi @henrikmv, Good starting point! Besides the comments in the ticket, I have a few more doubts:

1.  When trying to edit an event that is currently in a locked period and selecting a date outside the locked period, should I see the error message: "The date entered belongs to an expired period. Enter a date after..."? Currently the error is not displayed the the request to the API retunes with a 409 error. [Here](https://github.com/user-attachments/assets/36e24cdd-3237-44fc-8bde-43edb441ceec) is a video below with more details on this use case.
2.  The ticket also mentioned that events should no longer be deletable. Will that be implemented in a separate ticket or in this one? 

Thanks!

simonadomnisoru (approved):
Great job 🎉 !

PR #4174, opened 2025-06-24
fix: use 'en' as fallback in `localeCompareStrings`

superskip:
Slack thread:
https://dhis2.slack.com/archives/C45MWQ1QS/p1750422050038479


JoakimSM (approved):
LGTM 👍

PR #4245, opened 2025-07-24
fix: [DHIS2-15155] Filter Range Search Values before Prepopulation

henrikmv:
[DHIS2-15155](https://dhis2.atlassian.net/browse/DHIS2-15155)

This fix pull request introduces a filter function for filtering range-based search terms for prepopulation when creating new TEI's

### Fix: Filtering range-based search terms for prepopulation

* **New function for filtering range values**: Added `filteredRangeForPrepopulation`. This function returns an array without search terms where the "from" and "to" values of range-based fields are not equal.

### Integration: `SearchResults.container.js` and `SearchBox.container.js`

* Updated the `handleCreateNew` dispatch function to use `filteredRangeForPrepopulation` for filtering search terms before dispatching `setPrepopulateDataOnNewPage`.

[DHIS2-15155]: https://dhis2.atlassian.net/browse/DHIS2-15155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/SearchBox/SearchBox.container.js
+++ b/src/core_modules/capture-core/components/SearchBox/SearchBox.container.js
@@ -12,6 +12,7 @@ import { useCurrentTrackedEntityTypeId } from '../../hooks';
 import { ResultsPageSizeContext } from '../Pages/shared-contexts';
 import { usePreselectedProgram } from './hooks';
 import { setPrepopulateDataOnNewPage } from '../Pages/New/NewPage.actions';
+import { filteredRangeForPrepopulation } from './hooks/prepopulateRangeFilter';
 
 export const SearchBox = ({ programId }: { programId: string }) => {
     const dispatch = useDispatch();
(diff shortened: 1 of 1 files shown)


Wedikahisp:
**Test Summary**

1. Search worked successfully using all searchable attributes including:
- First name, Last name, Gender, Age, Address, City, State, Zip code, Email, Phone number, Mother maiden name, Occupation, Company, TB number, Vehicle, Blood type, Weight, Height

2. ❌ Residence location caused an error during search:
- Displays: “There is a problem with this search, please change the search terms or try again later.For more details open the Console tab of the Developer tools”
- ?? expected behavior since geo-coordinates may not be searchable.

3. Fixed behavior > confirmed:
- If different values are entered for "From" and "To" in a range attribute (e.g., Height: 150–180), the value is not copied to the new TEI form — expected behavior.

4. Correct behavior > confirmed:
-  If same value is entered for both “From” and “To” (e.g., Weight: 75–75), then that single value is carried forward to the TEI form.

5. ❌ Issue noted with ZIP code:
- Entered: 0777
-  In TEI form: becomes 777 >> leading zero is stripped

PR #4270, opened 2025-08-01
feat: [18904] Working List Filter for `null` and `not null` Values

henrikmv:
[DHIS2-18904](https://dhis2.atlassian.net/browse/DHIS2-18904)


simonadomnisoru (commented):
Hi @henrikmv. This is a great start, and it's good that you already separated the `TEXT` filter from the more complex ones 👍 . There are some other things that come to mind that we need to consider: 

- We should include a toggle feature to only display the null filter starting from version `2.42`.
- When a null filter is applied, the user should be able to save it via a `POST` request to the following three endpoints: `/trackedEntityInstanceFilters`, `/programStageWorkingLists`, and `/eventFilters`
- When a null filter is applied, the user should also be able to update it via a `PUT` request to the same three endpoints.
- When a stored WL containing a null filter is selected by the user, it should be loaded (`GET` request from the three endpoints mentioned above) and then applied to all three WL types.

JoakimSM (approved):

PR #4290, opened 2025-08-07
fix: [DHIS2-19877] filter rule engine values containing undefined

superskip:
[DHIS2-19877](https://dhis2.atlassian.net/browse/DHIS2-19877)

Attributes and data elements containing the value `undefined` become converted to `'undefined'` in the rule engine. We had the same issue before with `null` getting converted to `'null'`, and the solution was to filter out all form values containing `null`. This PR extends the solution to form values containing `undefined`.


simonadomnisoru (approved):
LGTM!

linadhis2:
LGTM from QA check.

PR #4370, opened 2025-08-28
fix: [DHIS2-15903] prod env fix for indexedDB caching

JoakimSM:
We need the ABSOLUTE base url for the instance when building the IndexedDB caches. We relied on the `baseUrl` property of the root object returned by `useConfig` (from app-runtime) for this, but it is not reliable as it will return a relative path for: some backend versions (v41 and below) / app-shell versions.

We will use the `contextPath` set by the backend instead.


simonadomnisoru (approved):
LGTM!

PR #4376, opened 2025-09-10
fix: use rule engine's user roles as fallback

superskip:
[DHIS2-20600](https://dhis2.atlassian.net/browse/DHIS2-20060?actionerId=61bcf639e399d700715e8abf&sourceType=assign)

The rule engine generally does not receive a user role array directly as input, and in this case it should use the one with which it was initialized.


simonadomnisoru (approved):
LGTM!

PR #4468, opened 2026-02-27
feat: [DHIS2-19940] Implement "No value" filtering for all filter types in working list

henrikmv:
[DHIS2-19940](https://dhis2.atlassian.net/browse/DHIS2-19940)
### Pattern for Adding `isEmpty` Filtering

**1. Types**
- Union the filter data type with `EmptyValueFilterData`
- Add `string` to the `Value` type (used to carry the isEmpty value)

**2. Component**
- Wrap the filter UI with `WithEmptyValueFilter`
- This renders the checkboxes and passes `undefined` to children when `isEmpty` is active

**3. Data Getter**
- Check `isEmptyValueFilter(value)` before normal conversion
- If true, return `getEmptyValueFilterData(value)`

**4. Manager**
- When restoring from a saved filter, check `isEmptyFilterData(filter)`
- If true, return `getEmptyValueFilterValue(filter)`

**5. Disable Flag**
- Set `disableEmptyValueFilter` to `true` for:
  - Main properties
  - `transformRecordsFilter` columns

[DHIS2-19940]: https://dhis2.atlassian.net/browse/DHIS2-19940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/FiltersForTypes/Date/DateFilter.component.tsx
```diff
@@ -197,16 +226,14 @@ class DateFilterPlain extends Component<Props, State> implements UpdatableFilter
                 />
                 {isAbsoluteRange && (
                     <AbsoluteRangeFilter
-                        value={{ from: value?.from, to: value?.to }}
+                        value={{ from: objValue?.from, to: objValue?.to }}
                         submitAttempted={this.state.submitAttempted}
                         onFieldBlur={this.handleFieldBlur}
-                        onFieldChange={this.handleFieldChange}
-                        onKeyDown={this.handleKeyDown}
                     />
```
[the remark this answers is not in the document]
henrikmv: The bug is a false positive because `AbsoluteRangeFilter` uses `D2Date` which only accepts `onBlur` — it never declared `onFieldChange` or `onKeyDown` in its `OwnProps`. Those props were silently ignored when passed before. `RelativeRangeFilter` needs them because it uses `D2TextField`, which has a custom `onKeyDown` contract where `Enter` must be manually forwarded to the parent.

src/core_modules/capture-core/components/FiltersForTypes/Assignee/assignee.types.ts
```diff
@@ -0,0 +1,28 @@
+import type { assigneeFilterModes } from './assignee.const';
+
+export type AssigneeFilterData = {
+    assignedUserMode: typeof assigneeFilterModes[keyof typeof assigneeFilterModes];
```
simonadomnisoru: Could use `assignedUserMode: AssigneeMode` from below

src/core_modules/capture-core/components/FiltersForTypes/DateTime/dateTimeFilterDataGetter.ts
```diff
@@ -38,7 +49,7 @@ export function getDateTimeFilterData(value: NonNullable<Value>): DateTimeFilter
     }
 
     if (value.to?.date) {
-        const le = buildIsoDateTime(value.to, '00:00');
+        const le = buildIsoDateTime(value.to, '23:59');
```
simonadomnisoru, line 51: Even though this is out of this ticket scope, I agree that having the Before time set to `23:59` rather than `00:00` is the most sensible approach 👍

src/core_modules/capture-core/components/FiltersForTypes/types/index.ts
```diff
@@ -1,29 +1,29 @@
...
-import type { OrgUnitFilterData } from '../OrgUnit/types';
-import type { TextFilterData } from '../Text/types';
-import type { TimeFilterData } from '../Time';
-import type { TrueOnlyFilterData } from '../TrueOnly/types';
-import type { UsernameFilterData } from '../Username/types';
+import type { AssigneeFilterData } from '../Assignee/assignee.types';
+import type { BooleanFilter } from '../Boolean/boolean.types';
+import type { DateFilter } from '../Date/date.types';
+import type { DateTimeFilter } from '../DateTime/dateTime.types';
+import type { EmptyOnlyFilterData } from '../EmptyOnly/emptyOnly.types';
```
simonadomnisoru: NIT: Could probably be renamed to `EmptyOnlyFilter` to keep the same naming convention

src/core_modules/capture-core/components/FiltersForTypes/Boolean/BooleanFilter.component.tsx
```diff
@@ -3,42 +3,23 @@ import { withStyles, type WithStyles } from 'capture-core-utils/styles';
...
-
-        if (!value && value !== false) {
-            return null;
-        }
-
-        if (!allowMultiple) {
-            return getSingleSelectBooleanFilterData(value);
-        }
-
-        return getMultiSelectBooleanFilterData(value);
```
simonadomnisoru, line 41: Is `allowMultiple` and `getMultiSelectBooleanFilterData` logic safe to remove? I tried to trace their original logic, but did not find a metadata setup that triggers them

src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.ts
```diff
@@ -96,8 +88,6 @@ const getMainApiFilterQueryArguments = (filters: { [id: string]: any } | null, m
                         queryArgsForCurrentMain = getStatusQueryArgs(filter);
                     } else if (key === 'assignee') {
                         queryArgsForCurrentMain = filter;
-                    } else if (key === 'orgUnitId') {
-                        queryArgsForCurrentMain = getOrgUnitFilterQueryArgs(filter);
```
simonadomnisoru, line 100: Did we have an `orgUnit` main filter in the Events WLs prior to [PR #4461](https://github.com/dhis2/capture-app/pull/4461), or has this code always been unreachable? Thanks!

src/core_modules/capture-core/components/FiltersForTypes/Boolean/BooleanFilter.component.tsx
```diff
@@ -61,20 +42,28 @@ class BooleanFilterPlain extends Component<Props> implements UpdatableFilterCont
...
-                    onBlur={onCommitValue}
-                    orientation={orientations.VERTICAL}
-                />
-            </div>
+                {filteredValue => (
+                    <div
+                        className={classes.selectBoxesContainer}
+                        onKeyDownCapture={this.handleKeyDown}
+                    >
+                        <D2TrueFalse
```
simonadomnisoru, line 47: The BOOLEAN filter seems to have some layout issues. Could you please check it? 

Image

src/core_modules/capture-core/components/FiltersForTypes/OptionSet/OptionSetFilter.component.tsx
```diff
@@ -41,20 +35,28 @@ class OptionSetFilterPlain extends Component<Props> implements UpdatableFilterCo
...
-                        nullable
-                    />
-                </div>
-            </div>
+                {filteredValue => (
+                    <div
+                        className={classes.selectBoxesContainer}
+                    >
+                        <div className={classes.selectBoxesInnerContainer}>
+                            <SelectBoxes
```
simonadomnisoru: The OptionSet filter seems to have some layout issues. The issue looks similar to the Boolean filter one. Could you please check it?

Image

src/core_modules/capture-core/components/FiltersForTypes/Boolean/BooleanFilter.component.tsx
henrikmv: The removal of `getSingleSelectBooleanFilterData` and `getMultiSelectBooleanFilterData` as separate exports is safe because the new unified `getBooleanFilterData` handles both cases. 

It normalises the value into an array with `const values = Array.isArray(value) ? value : [value]`, which is equivalent to what the old `getSingleSelectBooleanFilterData` did (wrapping in `[value]`).

src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.ts
henrikmv: Thank you for raising the question! The Org unit main filter is a confusing one for both Tracker and Event programs. I will try to summarise my findings below:

The OrgUnit main filter became visible after [dhis2/capture-app#4461]. Before that PR, only ordering on `orgUnit` was supported. However, even though the filter was visible, it did not work — the `orgUnit` filter value was not persisted when saving a working list view.

We had already disabled the OrgUnit main filter for Tracker programs because filtering (and ordering) on `programOwnerId` is not supported by the API. It was not immediately obvious that it also needed to be disabled for Event programs, since Event programs use `orgUnit` as the `apiName` rather than `programOwnerId`.

After investigating, I found that the `orgUnit` parameter in the Event program request is set by the context selector. The main filter value is not persisted because, unlike `status`, `occurredAt`, and `assignee`, the backend does not store `orgUnit` as part of the event query criteria. Routing the `orgUnit` filter through `dataFilters` instead does persist correctly, but this inconsistency with how the other main filters are handled is not a clean solution. Enabling the filter properly would require backend support for persisting `orgUnit` in the criteria.

Given this, in my opinion, there are good reasons for hiding the filter:
- It aligns Event program behaviour with Tracker program behaviour, where the filter is already hidden.
- Users can still filter events by org unit through the context selector.
- Persisting the orgUnit main filter requires backend support for storing orgUnit in the event query criteria.

Alternatively, we should ask the backend team to support us, but let's create another ticket for that. For now, I will keep the OrgUnit main filter cleanup in this PR.

Thank you!

src/core_modules/capture-core/components/FiltersForTypes/Boolean/boolean.types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type BooleanFilter = BooleanFilterData | EmptyValueFilterData;
+
+export type Value = Array<boolean> | boolean | string | null | undefined;
+
+export type BooleanFilterProps = {
+    value: Value;
+    onCommitValue: (value: Value) => void;
+    onUpdate?: (commitValue?: Value) => void;
+    allowMultiple: boolean;
```
simonadomnisoru, line 15: Based on this [discussion](https://github.com/dhis2/capture-app/pull/4468/changes#r3095346654), we can also delete `allowMultiple`, as the prop is no longer used by BooleanFilter.component.tsx. WDYT?

src/core_modules/capture-core/components/WorkingLists/EventWorkingLists/epics/getEventListData.ts
simonadomnisoru: Given the technical history and current API limitations, I agree with the decision to hide the filter for now. Thanks for the explanation!

src/core_modules/capture-core/components/FiltersForTypes/Boolean/boolean.types.ts
henrikmv: The **Follow up** filter ([useFiltersOnly.ts](https://github.com/dhis2/capture-app/blob/17fd3b6d8f8c09a1b53ab9db76916ca622f8aaef/src/core_modules/capture-core/components/WorkingLists/TrackerWorkingLists/Setup/hooks/useFiltersOnly.ts#L76)) is a `BOOLEAN` with `multiValueFilter: false`, which means `allowMultiple=false`. Removing it would break that filter.

Do you agree that we keep it? Thanks!

simonadomnisoru (approved):
Good job! 👏

geethaalwan (approved):
Tested Successfully on 2.44 version

PR #4493, opened 2026-03-20
fix: [DHIS2-20985] Username field selects first suggestion instead of clicked user

henrikmv:
[DHIS2-20985](https://dhis2.atlassian.net/browse/DHIS2-20985)

The UserField component has a search input with a suggestions dropdown. When focus leaves both the input and the suggestions, onExitSearch() is called to close the dropdown. This blur detection previously relied on checking getAttribute('name') on `<div>` elements, which is non-standard and fails when event.relatedTarget is a child element (e.g. inside <MenuItem>) that doesn't carry the name attribute.

This PR makes two fixes:
1. Switches to data-suggestion-name for identifying suggestion elements in blur handlers, checked via element.dataset.suggestionName,
2. Adds preventDefault() on mouseDown to prevent the classic blur-before-click race condition when clicking a suggestion.

Some existing bugs for long values are also been improved:

bilde

Skjermbilde 2026-03-20 kl  19 55 06

[DHIS2-20985]: https://dhis2.atlassian.net/browse/DHIS2-20985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/FormFields/UserField/Input.component.tsx
+++ b/src/core_modules/capture-core/components/FormFields/UserField/Input.component.tsx
@@ -17,20 +17,16 @@ type Props = {
     placeholder?: string | null;
 };
 
-const hasSuggestionAttribute = (element: any, suggestionName: string) =>
-    element?.getAttribute?.('data-suggestion-name') === suggestionName;
+const hasDataSuggestionName = (element: EventTarget | null, suggestionName: string): boolean =>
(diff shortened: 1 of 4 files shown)


simonadomnisoru (approved):

geethaalwan (approved):

PR #4514, opened 2026-04-17
fix: [DHIS2-21136] disable Enrollement actions buttons on ready only access

henrikmv:
[DHIS2-21136](https://dhis2.atlassian.net/browse/DHIS2-21136)

Users without write access to the program are not able to perform any of the enrollment actions.

Previously, the action buttons were always enabled, allowing read-only users to trigger actions like “Transfer”, which would fail at the API level.

The only button with validation was the “Add new” button, which checks TET write access. But clicking the  “Add new” button without program access will show a “no access” message on the New Page.
bilde

Because of this, all actions depend on program write access. Disabling the entire enrollment actions button ensures consistent behaviour and prevents users from triggering actions they do not have permission to complete.

[DHIS2-21136]: https://dhis2.atlassian.net/browse/DHIS2-21136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):

geethaalwan (approved):

PR #4516, opened 2026-04-17
fix: [DHIS2-20032] Update "Last updated" in enrollment widget on change

henrikmv:
[DHIS2-20032](https://dhis2.atlassian.net/browse/DHIS2-20032)

This PR also includes a fix for a pre-existing stale closure issue in the onError handler. `useDataMutation` may capture the callback only once, so if the enrollment state changes between mutation and error, the rollback (`setEnrollment(enrollment)`) could restore a stale state. This is solved by using a `useRef` to snapshot the enrollment right before each mutation — the same pattern used in [WidgetAssignee.container.tsx](https://github.com/dhis2/capture-app/blob/master/src/core_modules/capture-core/components/WidgetAssignee/WidgetAssignee.container.tsx). While not introduced by this PR, the optimistic `updatedAt` update makes it slightly more impactful, so it made sense to address it here.

[DHIS2-20032]: https://dhis2.atlassian.net/browse/DHIS2-20032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/WidgetEnrollment/hooks/useUpdateEnrollment.ts
+++ b/src/core_modules/capture-core/components/WidgetEnrollment/hooks/useUpdateEnrollment.ts
@@ -31,8 +31,8 @@ export const useUpdateEnrollment = ({
     const [updateEnrollmentMutation] = useDataMutation(enrollmentUpdate, {
         onError: (e) => {
             setEnrollment(prevEnrollment.current);
-            updateHandler && updateHandler(prevEnrollment.current[propertyName]);
-            onError && onError(processErrorReports(e));
+            updateHandler?.(prevEnrollment.current[propertyName]);
+            onError?.(processErrorReports(e));
         },
     });
 
@@ -43,6 +43,6 @@ export const useUpdateEnrollment = ({
         updatedEnrollment.updatedAt = fromClientDate(new Date()).getServerZonedISOString();
         setEnrollment(updatedEnrollment);
         updateEnrollmentMutation(updatedEnrollment);
-        updateHandler && updateHandler(value);
+        updateHandler?.(value);
     }, [enrollment, setEnrollment, propertyName, updateHandler, updateEnrollmentMutation, fromClientDate]);
 };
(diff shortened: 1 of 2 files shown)


src/core_modules/capture-core/components/WidgetEnrollment/hooks/useUpdateEnrollment.ts
```diff
@@ -26,19 +26,23 @@ export const useUpdateEnrollment = ({
...
+            updateHandler?.(prevEnrollment.current[propertyName]);
+            onError?.(processErrorReports(e));
         },
     });
 
     return useCallback((value: string) => {
+        prevEnrollment.current = enrollment;
         const updatedEnrollment = { ...enrollment };
         updatedEnrollment[propertyName] = value;
+        updatedEnrollment.updatedAt = fromClientDate(new Date()).getServerZonedISOString();
```
simonadomnisoru, line 43: Hi! Could we also update the conditional tooltip to reflect the correct timezone? Now when pointing at dev play server, it shows a 2-hour offset difference.

https://github.com/user-attachments/assets/ea853d70-a312-47e8-a3f3-d5992e79eb2d

Thank you!

simonadomnisoru (commented):
> This PR also includes a fix for a pre-existing stale closure issue in the onError handler. `useDataMutation` may capture the callback only once, so if the enrollment state changes between mutation and error, the rollback (`setEnrollment(enrollment)`) could restore a stale state. This is solved by using a `useRef` to snapshot the enrollment right before each mutation.

Interesting insights regarding the rollback that could restore a stale state. I suspect something similar is happening with the difficult-to-reproduce bug [DHIS2-21152](https://dhis2.atlassian.net/browse/DHIS2-21152), so I’ll use your discoveries to investigate further 🤔.

src/core_modules/capture-core/components/WidgetEnrollment/hooks/useUpdateEnrollment.ts
henrikmv: Thank you for catching this pre-existing bug. This PR is a good place to fix it.

simonadomnisoru (approved):

geethaalwan (approved):

PR #4519, opened 2026-04-20
chore: [DHIS2-19014] Ensure moment format is using Western Arabic 

henrikmv:
[DHIS2-19014](https://dhis2.atlassian.net/browse/DHIS2-19014)

This PR renames the date-formatting utility and broadens its signature, then migrates every call site across the codebase.

## Core change

```ts
// Before
getFormattedStringFromMomentUsingEuropeanGlyphs(
  momentDate: Moment,
  format?: string
)

// After
formatMomentEn(
  input: MomentInput,
  format?: string
)
```

The new function accepts any `MomentInput` (not just a `Moment` instance), constructs the moment internally, and forces the `en` locale.

### 1. Bug fixes
Several sites used bare `moment().format()` without `.locale('en')`, which would produce locale-dependent numerals (e.g. Arabic `٠١٢٣`). These now go through `formatMomentEn`.

### 2. Mechanical renames
Sites that already called the old helper or manually did `.locale('en').format()` are consolidated to the new one-liner.

[DHIS2-19014]: https://dhis2.atlassian.net/browse/DHIS2-19014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core-utils/date/date.utils.ts
+++ b/src/core_modules/capture-core-utils/date/date.utils.ts
@@ -14,8 +14,8 @@ import moment from 'moment';
  * @returns the formatted string with Western-Arabic numerals
  */
 export function formatMomentEn(
-    momentDate: moment.Moment,
+    input: moment.MomentInput,
     format = 'YYYY-MM-DD',
 ) {
-    return momentDate.clone().locale('en').format(format);
+    return moment(input).locale('en').format(format);
 }
--- a/src/core_modules/capture-core-utils/date/date.utils.ts
+++ b/src/core_modules/capture-core-utils/date/date.utils.ts
@@ -1,16 +1,21 @@
 import moment from 'moment';
 
 /**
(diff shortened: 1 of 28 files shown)


simonadomnisoru (commented):
The approach looks good overall. I only noticed [convertDateTimeForView](https://github.com/dhis2/capture-app/blob/0602038db9fc98033f6aa2dc777aa386d11ece53/src/core_modules/capture-core/converters/clientToView.tsx#L16) and [convertDateTimeForListDisplay](https://github.com/dhis2/capture-app/blob/0602038db9fc98033f6aa2dc777aa386d11ece53/src/core_modules/capture-core/converters/clientToList.tsx#L17), which are still using `.locale('en').format()` instead of the helper function. Could those be refactored as well? 
Thanks!

simonadomnisoru (approved):

PR #4521, opened 2026-04-22
fix: [DHIS2-21348] Console log errors if a TEA associated with a optionSet is empty

henrikmv:
[DHIS2-21348](https://dhis2.atlassian.net/browse/DHIS2-21348)

Suppresses spurious error logs when a working list option value is null or empty.

Previously, the code logged an error for every missing option lookup—including when clientValue was `null` or `''`, which is expected and not an error. The logic is restructured to only log when a non-empty value genuinely has no matching option:

[DHIS2-21348]: https://dhis2.atlassian.net/browse/DHIS2-21348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):

geethaalwan (approved):

PR #4522, opened 2026-04-22
fix: [DHIS2-17625] Use Capture scope for working lists without org unit

henrikmv:
[DHIS2-17625](https://dhis2.atlassian.net/browse/DHIS2-17625)

Switch the working-list requests from **ACCESSIBLE** to **CAPTURE** org unit mode when no org unit is selected, and update the empty-state on the Main page so users get the right next step for their program type.

## Behavior
- **Org unit selected** — unchanged: requests use **SELECTED** scope.
- **No org unit selected, working list fetch/download** — requests now use **CAPTURE** scope (was **ACCESSIBLE**), so users only see records they can capture into.
- **No org unit selected, event program** — empty state shows a **See working list without organisation unit** link.
- **No org unit selected, tracker program with displayFrontPageList** — empty state shows both **See working list without organisation unit** and **Search for a {{trackedEntityName}}** links.
- **No org unit selected, tracker program without displayFrontPageList** — empty state shows only the **Search for a {{trackedEntityName}}** link.

[DHIS2-17625]: https://dhis2.atlassian.net/browse/DHIS2-17625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henrikmv:
I’ve asked @karolinelien  to help refine the info box message. However, the logic is ready for review. Thanks!

simonadomnisoru (commented):
Code-wise looks good! Agree that the info box message needs some refinement from Product. I'll wait for that feedback before approving.
Could you also update the description for [DHIS2-17625](https://dhis2.atlassian.net/browse/DHIS2-17625)? Since the final PR implementation followed Joakim's suggestion rather than the initial description. Thank you!

karolinelien:
Maybe just very concise:
Select an organisation unit to see the working list.

See records you can edit in Child Programme

Search for a specific record.

@cooper-joe Thoughts?

cooper-joe:
I agree with @karolinelien, let's keep it concise here even if it means simplifying the exact logic.

I suggest:

```
Choose an organisation unit to see the working list.
See all records you can edit in this program
Search for a specific record
```

image

---

I also agree with something @henrikmv mentioned on slack: this component could use some design work. I think there's certainly room for improvement here, but I'm going to follow it up as a separate task so this bug fix isn't blocked.

henrikmv:
Current suggestion for _message box content_ is created with help from @cooper-joe and @karolinelien. Thanks!

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.44,2.43.1,2.42.5,2.41.9 versions

PR #4524, opened 2026-04-23
fix: [DHIS2-18592] Not saving an option that starts or ends with a space

henrikmv:
[DHIS2-18592](https://dhis2.atlassian.net/browse/DHIS2-18592)

Stop trimming option-set codes so values like " foo " round-trip and match correctly. Affects the create/edit form, the search page, and the profile widget's unique-attribute lookup.

## Behavior
- Option set, create/edit — don't trim; server stores the code verbatim.
- Option set, search — don't trim; stored value is verbatim, so the filter must include spaces.
- MULTI_TEXT, create/edit — don't trim; server splits and normalizes each segment on save.
- MULTI_TEXT, search — trim; stored segments are already normalized, so the filter must be the trimmed code.

[DHIS2-18592]: https://dhis2.atlassian.net/browse/DHIS2-18592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/converters/searchFormToServer.ts
+++ b/src/core_modules/capture-core/converters/searchFormToServer.ts
@@ -9,8 +9,7 @@ type FormValues = { [key: string]: any };
 const convertValuePipe = pipeD2(convertFormToClient, convertClientToServer);
 
 const convertString = (formValue: string, dataElement: DataElement, searchOperator: SearchOperator) => {
-    const skipTrim = dataElement.optionSet && dataElement.type !== dataElementTypes.MULTI_TEXT;
-    const sanitizedString = skipTrim ? formValue : formValue.trim();
+    const sanitizedString = dataElement.optionSet ? formValue : formValue.trim();
     const convertedString = dataElement.convertValue(sanitizedString, convertValuePipe);
     return `${dataElement.id}:${searchOperator.toLowerCase()}:${escapeString(convertedString)}`;
 };
--- a/src/core_modules/capture-core/converters/searchFormToServer.ts
+++ b/src/core_modules/capture-core/converters/searchFormToServer.ts
@@ -9,7 +9,8 @@ type FormValues = { [key: string]: any };
 const convertValuePipe = pipeD2(convertFormToClient, convertClientToServer);
 
 const convertString = (formValue: string, dataElement: DataElement, searchOperator: SearchOperator) => {
-    const sanitizedString = dataElement.optionSet ? formValue : formValue.trim();
+    const skipTrim = dataElement.optionSet && dataElement.type !== dataElementTypes.MULTI_TEXT;
+    const sanitizedString = skipTrim ? formValue : formValue.trim();
     const convertedString = dataElement.convertValue(sanitizedString, convertValuePipe);
     return `${dataElement.id}:${searchOperator.toLowerCase()}:${escapeString(convertedString)}`;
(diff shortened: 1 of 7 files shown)


src/core_modules/capture-core/converters/searchFormToServer.ts
```diff
@@ -9,7 +9,7 @@ type FormValues = { [key: string]: any };
 const convertValuePipe = pipeD2(convertFormToClient, convertClientToServer);
 
 const convertString = (formValue: string, dataElement: DataElement, searchOperator: SearchOperator) => {
-    const sanitizedString = formValue.trim();
+    const sanitizedString = dataElement.optionSet ? formValue : formValue.trim();
```
simonadomnisoru: The search flow is still failing for `MULTI_TEXT` value with options containing start/end spaces, even though we are now sending the non-trimmed values. Is this because the API stores MULTI_TEXT as a comma-separated string?

https://github.com/user-attachments/assets/6b2563b4-0b54-4b9e-bfef-78e9ba756af2

Also, any idea if the BE plans to refactor MULTI_TEXT soon? I feel we keep on adding workarounds in Capture to compensate for the current BE limitations, but the functionality and UX for this value type are still not where they need to be.

henrikmv: Thank you for spotting this issue, @simonadomnisoru. Searching with `MULTI_TEXT` worked previously, so this only became a problem after we stopped trimming values before running the query.

I agree that the number of frontend workarounds for `MULTI_TEXT` is a strong signal that we soon should prioritise a proper backend fix. That said, a refactor on the backend side is a significant effort and needs careful planning. In the meantime, we should document the workarounds so they can be easily removed once the refactor is complete.

Would you like me to create a ticket to track the rollback of these workarounds and tag the relevant code with the ticket number? We can also discuss this in our next FE meeting.

Thanks a lot!

simonadomnisoru: Yes, please create a ticket so we can start tracking the FE workarounds for `MULTI_TEXT` and we can also discuss it in more detail in our next FE meeting. Thanks!

simonadomnisoru (approved):

henrikmv:
@simonadomnisoru, the rollback ticket for our workarounds is created: [DHIS2-21409](https://dhis2.atlassian.net/browse/DHIS2-21409).

However, after speaking to @ameenhere, we decided to fix the `MULTI_TEXT` search bug in the backend (See ticket [DHIS2-21414)](https://dhis2.atlassian.net/browse/DHIS2-21414). Therefore, the workaround is reverted, and the QA will be informed about the limitation before testing.

Thank you both for helping me!

[DHIS2-21409]: https://dhis2.atlassian.net/browse/DHIS2-21409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

simonadomnisoru (approved):

geethaalwan (approved):
Tested successfully on 2.44,2.43.1,2.42.5,2.41.9 versions

PR #4529, opened 2026-04-28
feat: [DHIS2-21371] Read-only enrollment dashboard

henrikmv:
[DHIS2-21371](https://dhis2.atlassian.net/browse/DHIS2-21371)

When ALL write access is missing, `showWidgetBadge=false` suppresses per-widget badges, and the page-level `EnrollmentReadOnlyBadge` is shown instead. When only SOME access is missing, `showWidgetBadge=true` enables per-widget badges while no page-level badge appears.

| Action / Element | Missing TET access | Missing Program access | Missing Program stage access | All missing |
|---|---|---|---|---|
| Add event (Program stage) | possible | possible | button hidden per stage | button hidden per stage |
| Edit event (button on event) | possible | possible | button hidden per stage | button hidden per stage |
| Delete event (overflow menu) | possible | possible | overflow menu hidden per stage | overflow menu hidden per stage |
| Quick actions widget visible | shown | shown | widget hidden | widget hidden |
| Modify enrollment notes | possible | widget in read-only | possible | widget in read-only |
| Modify event notes | possible | possible | widget in read-only | widget in read-only |
| Modify Relationship | button hidden, read-only badge | possible | possible | button hidden, read-only badge |
| Edit person | "Show profile" button, read-only modal | possible | possible | "Show profile" button, read-only modal |
| Delete person | disabled per WidgetProfile | possible | possible | disabled per WidgetProfile |
| Enrollment actions button | shown | hidden | shown | hidden |
| Enrollment date / incident date / coordinator edit | possible | hidden | possible | hidden |
| Top-right page badge | no | no | on view event page | shown on enrollment dashboard |
| Linked event widget (view + unlink actions) | shown | shown | "View only" badge on linked stage; unlink + unlink-and-delete actions disabled | "View only" badge; unlink actions disabled |
| Link to related event (WidgetRelatedStages) | possible | possible | widget hidden if current OR target stage lacks write access | widget hidden |

Note: `Block data entry after completion` and `Expired period` used to disable the `Edit event` button and show a tooltip. They now follow the read-only pattern — the button is hidden and the badge explains why.

 `completeEventsExpiryDays` is not respected and will be added in https://github.com/dhis2/capture-app/pull/4583

[DHIS2-21371]: https://dhis2.atlassian.net/browse/DHIS2-21371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/PageLayout/DefaultPageLayout.constants.ts
```diff
@@ -24,6 +24,7 @@ export const DefaultPageLayout: PageLayoutConfig = {
         {
             type: WidgetTypes.COMPONENT,
             name: 'TrackedEntityRelationship',
+            settings: { readOnlyMode: true },
```
simonadomnisoru, line 27: Just double-checking if the TrackedEntityRelationships Widget default layout should be in `readOnlyMode` on the edit and add event pages. Thanks!

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/EnrollmentPageLayout.tsx
```diff
@@ -49,10 +51,52 @@ const getEnrollmentPageStyles: Readonly<any> = () => ({
...
+    breadcrumbRow: {
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'space-between',
+    },
 });
 
 const isValidHex = (color: string) => /^#[0-9A-F]{6}$/i.test(color);
 
+const EnrollmentReadOnlyBadge = () => {
```
simonadomnisoru: I suggest moving `EnrollmentReadOnlyBadge` to its own file and trying to keep EnrollmentPageLayout file smaller and cleaner

src/core_modules/capture-core/components/WidgetEnrollment/MapModal/Coordinates/Coordinates.component.tsx
```diff
@@ -98,6 +99,7 @@ const CoordinatesPlain = ({
         <Map
             center={center ?? initialCenter}
             zoom={13}
+            zoomControl={false}
```
simonadomnisoru, line 102: Why `zoomControl` is always set to `false`? Perhaps it should depend on readOnly value? 🤔

src/core_modules/capture-core/components/WidgetProfile/DataEntry/DataEntry.component.tsx
```diff
@@ -21,18 +33,39 @@ export const DataEntryComponent = ({
...
+    accessReadOnly,
+}: PlainProps & WithStyles<typeof styles>) => (
     <Modal large onClose={onCancel} dataTest="modal-edit-profile">
-        <ModalTitle>{i18n.t('Edit {{trackedEntityName}}',
-            { trackedEntityName, interpolation: { escapeValue: false } },
-        )}</ModalTitle>
+        <ModalTitle>
+            <div className={classes.title}>
+                <span>
+                    {readOnly
```
simonadomnisoru: There are quite a few `readOnly` checks in this component. Do you think it would be cleaner to split this into two components: `DataEntryComponent` and `DataEntryReadOnlyComponent`?

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentAccessContext/EnrollmentAccessContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const Context = createContext<EnrollmentAccessContextValue>(fallback);
+
+type ProviderProps = {
+    program?: TrackerProgram;
+    currentStageId?: string;
+    children: React.ReactNode;
+};
+
+export const EnrollmentAccessProvider = ({ program, currentStageId, children }: ProviderProps) => {
```
simonadomnisoru: I like the choice of using `Context` for this feature's architecture. Great decision 👍 !
I was wondering if we should also wrap `EnrollmentAddEventPageDefault.component.tsx` with `EnrollmentAccessProvider`, as you did in the EnrollmentPageDefault container and the EnrollmentEditEventPage component?

src/core_modules/capture-core/components/ReadOnlyBadge/ReadOnlyBadge.tsx
```diff
@@ -0,0 +1,70 @@
...
+}: Props) => {
+    if (programWriteAccess && trackedEntityTypeWriteAccess && programStageWriteAccess) return null;
+
+    const access: Access = {
+        program: programWriteAccess,
+        trackedEntityType: trackedEntityTypeWriteAccess,
+        programStage: programStageWriteAccess,
+    };
+    const message = getMissingAccessMessage(access, trackedEntityName, multipleStages);
+    const labelText = inlineLabel && message ? `${i18n.t('View only')} - ${message}` : i18n.t('View only');
```
simonadomnisoru: Perhaps `message` could be part of the translatable string? It may make a difference in other languages where the words order is different from English.

src/core_modules/capture-core/components/WidgetNote/WidgetNote.component.tsx
```diff
@@ -1,16 +1,56 @@
...
+}: Props & WithStyles<typeof styles>) => {
     const [open, setOpenStatus] = useState<boolean>(true);
+    const {
+        programWriteAccess,
+        currentStageWriteAccess,
+        trackedEntityTypeName,
+        showWidgetBadge,
+    } = useEnrollmentAccessContext();
+    const isEventScope = scope === 'event';
+    const readOnly = isEventScope ? !currentStageWriteAccess : !programWriteAccess;
```
simonadomnisoru: To keep `WidgetNote` generic, we should not add any event/enrollment scope dependencies inside it. Let's move those computations into the parent widgets, `WidgetEventNote` and `WidgetEnrollmentNote`, and pass the results to WidgetNote as generic props.

simonadomnisoru (commented):
Good job, design-wise it looks OK to me 👍 . I leave it up to Product for the more subtle UX decisions.

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/PageLayout/DefaultPageLayout.constants.ts
henrikmv: @karolinelien, @cooper-joe,

IMO, it makes sense to align the `TrackedEntityRelationships widget` with `Profile widget `and `Enrollement widget`. WDYT?

src/core_modules/capture-core/components/WidgetEnrollment/MapModal/Coordinates/Coordinates.component.tsx
henrikmv: `zoomControl={false}` disables Leaflet's default zoom control. It's not disabling Zoom functionality – line 111 renders a custom `<ZoomControl>` which handles RTL. Zoom is a viewing affordance, not an editing one, so it should stay available even when the map is read-only.

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/PageLayout/DefaultPageLayout.constants.ts
simonadomnisoru: In this case, I am not sure we want to align with the `Profile` or `Enrollment` Widgets. In my head, the trackedEntity Relationship Widget is aligned to the Event Program Relationship Widget, and the user is currently able to 'Add relationship' from a single event page -> https://play.im.dhis2.org/stable-2-43-0/apps/capture#/viewEvent?viewEventId=aa0fdac55d0
This is probably a Product call, so please confirm before changing the default layout. 
Thank you!

src/core_modules/capture-core/components/WidgetProfile/DataEntry/dataEntry.types.ts
```diff
@@ -2,28 +2,6 @@ import type { Geometry } from './helpers/types';
...
-    onGetValidationContext: () => Record<string, any>;
-    modalState: string;
-    errorsMessages: Array<{ id: string; message: string }>;
-    warningsMessages: Array<{ id: string; message: string }>;
-    center?: Array<number>;
-    orgUnitId: string;
-    pluginContext?: PluginContext;
-    readOnly?: boolean;
-    accessReadOnly?: boolean;
-};
```
simonadomnisoru, line 24: Any advantage to moving these Props to DataEntry.component.tsx? Keeping them in `dataEntry.types.ts` looks cleaner to me.

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.tsx
```diff
@@ -103,11 +107,17 @@ const WidgetProfilePlain = ({
...
-        !readOnlyMode,
-    [hasNoAttributes, readOnlyMode, trackedEntityTypeAccess]);
+    const isEditable = useMemo(
+        () => !hasNoAttributes && trackedEntityTypeWriteAccess && !readOnlyMode,
+        [hasNoAttributes, trackedEntityTypeWriteAccess, readOnlyMode],
+    );
+
+    const profileButtonLabel = useMemo(() => {
+        if (readOnlyMode || hasNoAttributes) return null;
+        if (!isEditable) return i18n.t('Show profile');
```
cooper-joe:
```suggestion
        if (!isEditable) return i18n.t('View profile');
```

src/core_modules/capture-core/components/ReadOnlyBadge/ReadOnlyBadge.tsx
```diff
@@ -0,0 +1,100 @@
...
+    inlineLabel?: boolean;
+};
+
+type Access = {
+    program: boolean;
+    trackedEntityType: boolean;
+    programStage: boolean;
+};
+
+const getEnrollmentMessage = (): string => i18n.t('You only have view access to enrollment');
```
cooper-joe:
```suggestion
const getEnrollmentMessage = (): string => i18n.t('You only have view access to this enrollment');
```

src/core_modules/capture-core/components/ReadOnlyBadge/ReadOnlyBadge.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const getProgramMessage = (): string => i18n.t('You only have view access to program');
```
cooper-joe:
```suggestion
const getProgramMessage = (): string => i18n.t('You only have view access to this program');
```

src/core_modules/capture-core/components/ReadOnlyBadge/ReadOnlyBadge.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const getTrackedEntityMessage = (trackedEntityName: string | undefined): string => (trackedEntityName
+    ? i18n.t('You only have view access to {{trackedEntityName}}', { trackedEntityName, escapeValue: false })
```
cooper-joe:
```suggestion
    ? i18n.t('You only have view access to this {{trackedEntityName}}', { trackedEntityName, escapeValue: false })
```

src/core_modules/capture-core/components/ReadOnlyBadge/ReadOnlyBadge.tsx
```diff
@@ -0,0 +1,100 @@
...
+    : i18n.t('You only have view access to this tracked entity type'));
+
+const getProgramStageMessage = (multipleStages: boolean): string => (multipleStages
+    ? i18n.t('You only have view access to these program stages')
+    : i18n.t('You only have view access to this program stage'));
+
+const getExpiredPeriodMessage = (): string => i18n.t('This event is outside the valid period');
```
cooper-joe:
```suggestion
const getExpiredPeriodMessage = (): string => i18n.t('This event is outside the valid editing period');
```

cooper-joe (approved):
A few suggested label changes, otherwise looks good.

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/PageLayout/DefaultPageLayout.constants.ts
karolinelien: So I think the question is specifically if the relationship widget should be read-only within the program stage workspace, and I think it should, just like with profile/enrollment. it might actually be confusing that relationship widget is editable on view/edit event because they might think it's to add a relationship to the event you're looking at, but it is a TE relationship. Which you can add from the enrollment dashboard.

@simonadomnisoru your link is to the event program, where I think the relationship widget is for the event and therefore it makes sense that it is still available.

Does that make sense?

simonadomnisoru: @karolinelien Yes, it makes sense! Thanks for the explanation

simonadomnisoru (approved):
Good job!

PR #4530, opened 2026-04-29
fix: [DHIS2-21286] Scroll validation error into center of viewport

karolinelien:
## What is the fix?

Replace the two-step scroll approach (`scrollIntoView()` + manual `-48px` offset) with `scrollIntoView({ block: 'center' })`.

The previous implementation had a hardcoded 48px header offset that did not fully account for the fixed header bar, causing the field label to scroll behind it on save/complete. Using `block: 'center'` positions the element in the middle of the viewport, guaranteeing visibility regardless of header height.

**Changed file:** `src/core_modules/capture-core/components/FormFields/New/HOC/withGotoInterface.tsx`

## Test plan

- [ ] Trigger save/complete on a program stage form with missing mandatory fields
- [ ] Verify the first validation error scrolls into view with the field label fully visible
- [ ] Test with different screen heights

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/core_modules/capture-core/components/DataEntry/dataEntryField/internal/DataEntryField.component.tsx
+++ b/src/core_modules/capture-core/components/DataEntry/dataEntryField/internal/DataEntryField.component.tsx
@@ -19,7 +19,7 @@ class DataEntryFieldPlain extends React.Component<Props> {
 
     goto() {
         if (this.gotoInstance) {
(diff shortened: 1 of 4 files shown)


karolinelien:
Fix does not work, not going to spend more time on this one for now.

PR #4531, opened 2026-04-29
fix: [DHIS2-20022] Always center map on org unit when opening map picker

karolinelien:
## What is the fix?

The `onOpenMap` callback in `withCenterPoint.tsx` previously only fetched the org unit geometry when the coordinate field had **no** existing value (`setFetchEnabled(!hasValue)`). When the field already had a value, the fetch was skipped and the map opened centered on the Leaflet default — London `[51.505, -0.09]`.

This fix always enables the org unit geometry fetch when the map opens, so the map is consistently centered on the org unit regardless of whether a coordinate value already exists.

**Changed file:** `src/core_modules/capture-core/components/FormFields/New/HOC/withCenterPoint.tsx`

## Test plan

- [ ] Open the profile widget for a TE that already has a coordinate set and verify it still opens where coordinate is set
- [ ] Verify that TEs enrolled in programs with no coordinate still open the map centered on the org unit (Geometry on enrollment)
- [ ] Verify that TEs created without program (So need Geometry on TET) still open the map centered on the org unit

[DHIS2-20022](https://dhis2.atlassian.net/browse/DHIS2-20022)

_Generated with [Claude Code](https://claude.com/claude-code)_

[DHIS2-20022]: https://dhis2.atlassian.net/browse/DHIS2-20022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/FormFields/New/HOC/withCenterPoint.tsx
+++ b/src/core_modules/capture-core/components/FormFields/New/HOC/withCenterPoint.tsx
@@ -93,8 +93,8 @@ const getCenterPoint = (InnerComponent: ComponentType<any>) => ({ orgUnitId, ...
         return null;
     }, [data, orgUnitFetchId]);
 
-    const onOpenMap = (hasValue) => {
-        setFetchEnabled(!hasValue);
+    const onOpenMap = () => {
+        setFetchEnabled(true);
     };
 
     const onCloseMap = () => {


superskip:
If `hasValue` is `true`, then logically I would expect there to already exist some coordinates that could be used. Why is Capture showing the London coordinates even though `hasValue` is `true`?

karolinelien:
> If `hasValue` is `true`, then logically I would expect there to already exist some coordinates that could be used. Why is Capture showing the London coordinates even though `hasValue` is `true`?

Very strange to me too. But apparently the whole component is deprecated according to @simonadomnisoru so I am mainly fixing the London issue and not digging any further 😁

src/core_modules/capture-core/components/FormFields/New/HOC/withCenterPoint.tsx
```diff
@@ -93,8 +93,8 @@ const getCenterPoint = (InnerComponent: ComponentType<any>) => ({ orgUnitId, ...
         return null;
     }, [data, orgUnitFetchId]);
 
-    const onOpenMap = (hasValue) => {
-        setFetchEnabled(!hasValue);
+    const onOpenMap = () => {
+        setFetchEnabled(true);
```
simonadomnisoru, line 97: Hi,
This approach introduces an unnecessary API call when coordinates are already present, so this is not a sensible solution.
But funny enough, while reviewing this PR, I noticed I can no longer reproduce the bug. Tried on both the latest 2.42 stable play and dev play as you can see in the video below.

https://github.com/user-attachments/assets/261969b2-0903-46f7-aa43-7f38ad8973da

Are you still able to reproduce it @karolinelien? If not, perhaps we could just close the Jira ticket. And if the bug is still reproducible, we'll need to rethink the approach for the fix. Thanks!

JoakimSM (commented):
The following is not criticising you in any way @karolinelien. It's great that you look into some of these bugs. But some honest feedback on the Claude solution here:

This is a "first day at work" kind of solution, not fixing the underlying issue and taking the easy way out, not caring about the impact it has on performance, maintainability and readability of the codebase.

This is what I worry about the most, at some point Claude will dig a hole so deep it will not get out of it and no human will ever want to touch an unreadable codebase filled with easy way out solutions and no architecture / structure.

I have not been been able to keep to up to date on many of the Claude/Devin PRs that we have done, but we still need a human to think here and make good architectural solutions.

I do think we can utilise Cloud and AI of course, but let's be super critical of the solutions suggested and make sure we keep the codebase maintainable, readable and performant. If not, Claude is on its own to fix its own mess.
 
Rant over.

@simonadomnisoru @henrikmv @superskip @ameenhere

src/core_modules/capture-core/components/FormFields/New/HOC/withCenterPoint.tsx
superskip: Actually I did fix an error like this as part of this -> [DHIS2-20475](https://github.com/dhis2/capture-app/pull/4414) :sweat_smile: @karolinelien

karolinelien: Amazing! Thanks @superskip 🎉

karolinelien:
💯 agree @JoakimSM - trying to focus on bugs where there are small things that solve what it actually needs to solve. This was a good example where it was just fixing something superficially. I will always defer and count on you all for good human eyes and knowledge, and yes, please be very critical of any PRs I may raise 😊  Hence also just let me know if something is not as easy as it looks or should not be a quick fix etc. Trust you way more than Claude 😂 

It seems like this issue was actually fixed either way, @superskip had already looked into it as part of a larger effort! 🎉 

Hence closing both this PR and the related JIRA issue. 👍🏻

PR #4561, opened 2026-05-14
chore: setup Dependabot cooldown

simonadomnisoru:
Given the recent increase in supply chain attacks, which are often detected and fixed within hours/days, I propose that we prevent Dependbot from opening PRs for any packages published less than 7 days ago.


henrikmv (approved):
Great security improvement. Thank you!

PR #4562, opened 2026-05-14
chore: [DHIS2-21506] Consolidate me API requests

henrikmv:
[DHIS2-21506](https://dhis2.atlassian.net/browse/DHIS2-21506)

Introduces a `CurrentUser` singleton, populated once at app startup from a single `/me` request, and replaces ~10 runtime `/me` fetches scattered across hooks and epics with synchronous reads.

`init.ts` already fetches `/me` at startup. This PR widens that fetch to include the fields previously queried ad-hoc (`firstName`, `surname`, `username`, `userRoles{id}`) and stores the result in a new `CurrentUser` class:

```ts
// Before: every consumer re-fetched /me
const { data } = useDataQuery({ me: { resource: 'me', params: { fields: '...' } } });
const locale = data?.me?.settings?.keyUiLocale;

// After: single init-time fetch, synchronous access
const locale = CurrentUser.get().uiLocale;
```

This change should not increase the initial payload size in practice.

It adds three small scalar fields — `firstName`, `surname`, and `username` — but reduces the amount of data requested for the four list fields: `userRoles`, `userGroups`, `organisationUnits`, and `teiSearchOrganisationUnits`.

For those collections, the request now fetches only `id`(and `path` where needed)

| Removed | Replaced by |
|---|---|
| `useUserLocale` hook (5 call sites) | `CurrentUser.get().uiLocale` |
| `useUserRoles` hook (WidgetProfile) | `CurrentUser.get().userRoles` |
| `useNoteDetails` hook (WidgetEventSchedule) | `CurrentUser.get()` + `generateUID()` inline |
| `useDataQuery({ me })` in `OrgUnitField` | `CurrentUser.get().organisationUnits` |
| `useApiMetadataQuery('me')` in `useSearchScopeWithFallback` | `CurrentUser.get().{teiSearchOrganisationUnits, organisationUnits}` |
| `querySingleResource('me')` in 5 note epics | `CurrentUser.get()` |

[DHIS2-21506]: https://dhis2.atlassian.net/browse/DHIS2-21506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.tsx
```diff
@@ -95,11 +93,7 @@ const WidgetProfilePlain = ({
         trackedEntityTypeAccess,
         geometry,
     } = useTrackedEntityInstances(teiId, programId, storedAttributeValues, storedGeometry);
-    const {
-        loading: userRolesLoading,
-        error: userRolesError,
-        userRoles,
-    } = useUserRoles();
+    const userRoles = CurrentUser.get().userRoles;
```
simonadomnisoru: Perhaps the `WidgetProfile` call should be left out of this refactor? That way, we stick to its initial self-contained architecture design and avoid introducing a dependency on the external `CurrentUser` singleton.

src/core_modules/capture-core/components/WidgetEnrollment/WidgetEnrollment.container.tsx
```diff
@@ -73,10 +73,10 @@ export const WidgetEnrollment = ({
     const { error: errorOrgUnit, displayName } = useOrgUnitNameWithAncestors(
         typeof ownerOrgUnit === 'string' ? ownerOrgUnit : undefined,
     );
-    const { error: errorLocale, locale } = useUserLocale();
+    const locale = CurrentUser.get().uiLocale;
```
simonadomnisoru: Same question regarding whether the `WidgetEnrollment` call should be left out of this refactor to stick as much as possible to the initial self-contained architecture?

src/core_modules/capture-core/components/WidgetEnrollment/TransferModal/OrgUnitField/useSearchScopeWithFallback.ts
```diff
@@ -1,28 +1,16 @@
...
-            enabled: !searchText,
-            select: (data) => {
-                const { teiSearchOrganisationUnits, organisationUnits } = data as any;
-                return teiSearchOrganisationUnits.length
-                    ? teiSearchOrganisationUnits
-                    : organisationUnits;
-            },
-        },
-    );
+    const orgUnitRootsFromUser = useMemo(() => {
```
simonadomnisoru: The `TransferModal` feature lives inside `WidgetEnrollment`, so the same concern applies here regarding our self-contained architecture principle. Thanks!

simonadomnisoru (commented):
Great refactor initiative @henrikmv 🙌 ! My only concern is whether the two self-contained widgets should be included or not. I'm curious to hear your thoughts on this. Thanks!

simonadomnisoru (approved):

PR #4572, opened 2026-05-18
chore: [DHIS2-21515] Improve performance for large option sets

henrikmv:
[DHIS2-21515](https://dhis2.atlassian.net/browse/DHIS2-21515)

This change introduces incremental client-side rendering using the onEndReached prop already exposed by @dhis2/ui's `SimpleSingleSelect`. Options remain fully in-memory (no server-side change), but only a window is mounted at a time:

- Activates only when filteredOptions.length > 200; smaller option sets render as before.
- Renders the first 100 options initially; onEndReached appends another 100 as the user scrolls.
- Window resets to the first 100 on every filter change so search results always start from the top.
- Selected value is still resolved against the full options array, so it displays correctly even when outside the rendered window.

[DHIS2-21515]: https://dhis2.atlassian.net/browse/DHIS2-21515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/FormFields/New/Fields/SingleSelectField/SingleSelectField.component.tsx
```diff
@@ -48,20 +49,24 @@ const NewSingleSelectFieldComponentPlain = ({
         [options, filterValue, filterable],
     );
 
+    const { visibleOptions, onEndReached, resetWindow } = useWindowedOptions(filteredOptions);
```
simonadomnisoru, line 52: Does it make sense to also call `resetWindow` inside `handleBlur` on line 74?

simonadomnisoru (approved):

linadhis2:
LGTM from QA perspective, was concentrating on option sets and no regressions spotted.

PR #4573, opened 2026-05-20
fix: [DHIS2-21475] revert patch for gregorian calendar for iso8601 to handle chrome bug

simonadomnisoru:
Reverting the patch from dhis2/capture-app#4560. The underlying Chromium issue with `Intl.DateTimeFormat` for the `iso8601` calendar has been fixed and is available in Chrome `v148.0.7778.179` (see [Chromium Issue 512514742](https://issues.chromium.org/issues/512514742)).


henrikmv (approved):
Thanks Simona for reverting this outdated bug fix.

PR #4579, opened 2026-05-22
fix: [DHIS2-21546] User can save empty note

henrikmv:
[DHIS2-21546](https://dhis2.atlassian.net/browse/DHIS2-21546)

- Prevent saving empty notes from the enrollment/event note widget by treating whitespace-only input as "not editing"
- Hide the Save/Cancel buttons whenever the field has no meaningful content
- Make "Save note" the primary action button

[DHIS2-21546]: https://dhis2.atlassian.net/browse/DHIS2-21546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetNote/NoteSection/NoteSection.tsx
```diff
@@ -141,23 +141,26 @@ const NoteSectionPlain = ({
...
-                <Button
-                    dataTest="cancel-note-btn"
-                    onClick={onCancel}
-                    secondary
-                    small
-                >
-                    {i18n.t('Cancel')}
-                </Button>
-            </div>}
+            {isEditing && (
```
simonadomnisoru: Design-wise, do you think it's cleaner to enable/disable the buttons instead of show/hide? I feel like show/hide the buttons creates too much movement on the page because it shifts all the other Widgets below. Any thoughts? 

https://github.com/user-attachments/assets/56616e56-8e9a-44d9-8637-ecc4cdefd70b

henrikmv: Thanks @simonadomnisoru, great review comment.

I have improved note-editing behaviour by adding a blur handler and disabling the `Save note` button when the note input field is empty. WDYT?

simonadomnisoru (approved):

PR #4582, opened 2026-05-24
chore: [DHIS2-21549] Align view single event right-column widgets

henrikmv:
[DHIS2-21549](https://dhis2.atlassian.net/browse/DHIS2-21549)

**The warning** is due to the branch name containing an en dash (`–`, `U+2013`) instead of a regular hyphen (`-`). Should not be blocking, so I leave the branch name unchanged.

[DHIS2-21549]: https://dhis2.atlassian.net/browse/DHIS2-21549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/Pages/ViewEvent/RightColumn/AssigneeSection/AssigneeSection.container.tsx
```diff
@@ -1,32 +1,48 @@
-import React from 'react';
-import type { ApiEnrollmentEvent } from 'capture-core-utils/types/api-types';
-import { WidgetAssignee } from '../../../../WidgetAssignee';
```
simonadomnisoru, line 3: Rollback the `WidgetAssignee` import. See the other comment.

src/core_modules/capture-core/components/Pages/ViewEvent/RightColumn/AssigneeSection/AssigneeSection.component.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import i18n from '@dhis2/d2-i18n';
+import { IconUser24 } from '@dhis2/ui';
+import type { ComponentType } from 'react';
+import { ViewEventListSection } from '../../Section/ViewEventListSection.component';
+import { DisplayMode } from '../../../../WidgetAssignee/DisplayMode.component';
+import { EditMode } from '../../../../WidgetAssignee/EditMode.component';
```
simonadomnisoru: We shouldn't import directly from inside the `Widget*` sub-folders. WidgetAssignee should be treated as a ready-to-use component that will manage its own internal state based on the props received from the parent.

PR #4583, opened 2026-05-24
fix: [DHIS2-21550] Not respecting completeEventsExpiryDays

henrikmv:
[DHIS2-21550](https://dhis2.atlassian.net/browse/DHIS2-21550)

### Summary
Completed events become read-only once the program's `completeEventsExpiryDays` window has passed. The scattered "can this event be edited?" checks are also consolidated into a single hook (`useEventEditPermissions`) so widgets, badges and the delete action all agree.

### Implementation notes

- **Users with `F_EDIT_EXPIRED` skip the expiry check.** Instead of adding an `if (hasAuthority)` everywhere the check is used, `useCompleteEventsExpiryForUser` just returns `undefined` for those users. The validator already treats an empty/`undefined` value as "no expiry limit," so these users automatically bypass it — same as when a program hasn't configured an expiry at all.

- **`F_UNCOMPLETE_EVENT` now also controls editing.** A completed event on a `blockEntryForm` stage can only be edited by users with this authority. We're reusing an "uncomplete" permission to mean "edit a completed locked event".

- **The expiry deadline is the last editable day + 1.** We compute `firstInvalidDate = completedAt + completeEventsExpiryDays` and allow editing only while `today < firstInvalidDate`. Example: completed on day 0 with `expiryDays = 3` → editable on days 0, 1, 2 and locked from day 3.

- **`DeleteActionButton` repeats the permission logic instead of using the hook.** It renders per event row, where the hook's program/stage inputs aren't easily available, so it recomputes `canEditCompletedEvent` / `isWithinCompleteExpiry` itself. Intentional duplication — both places must be updated if the rules change.

- **`completeEventsExpiryDays` is now stored** in `storePrograms` and added to the program API types.

- **Read-only/badge wiring** extended across `WidgetEventEdit`, `ViewEvent`, `EnrollmentEditEventPage`, `ReadOnlyBadge`, `WidgetHeader`.

[DHIS2-21550]: https://dhis2.atlassian.net/browse/DHIS2-21550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/EventRow/DeleteActionButton/DeleteActionButton.tsx
```diff
@@ -7,41 +7,68 @@ import {
...
     const occurredAtClientView = convertFn(occurredAt, dataElementTypes.DATE);
 
-    return (
-        <ConditionalTooltip
-            content={i18n.t('{{occurredAt}} belongs to an expired period. Event cannot be deleted', {
+    const completedAtClient = convertServerToClient(completedAt, dataElementTypes.DATE) as string;
+    const isWithinCompleteExpiry = isWithinCompleteEventsExpiry(completedAtClient, completeEventsExpiryDays);
+    const canEditCompletedEvent = canEditExpired || !(blockEntryForm && eventStatus === eventStatuses.COMPLETED);
+
+    const canDelete = isWithinValidPeriod && isWithinCompleteExpiry && canEditCompletedEvent;
```
simonadomnisoru: Could we use `useEventEditPermissions`? And perhaps move `canDelete` checks inside. Something like this:

```
    const {
        isEventWithinValidPeriod,
        canEditCompletedEvent,
        canDelete,
    } = useEventEditPermissions({
        programId,
        stage: programStage,
        eventStatus,
        occurredAtClient: convertServerToClient(occurredAt, dataElementTypes.DATE) as string,
        completedAtClient: convertServerToClient(completedAt, dataElementTypes.DATE) as string,
    });
....
```

henrikmv: Thanks, Simona! Yes, `useEventEditPermissions` should be used. However, `canDelete` (event) will be a duplicate of the `readOnly` prop. However, it can still be worth adding that prop for clarity, but at this point, I have not included it.

WDYT?

simonadomnisoru (approved):

PR #4590, opened 2026-05-28
chore: [DHIS2-21571] dev container (incl. safe-chain) and CI safe-chain

JoakimSM:
Adds a Dev Container config for VSCode. Using a Dev Container ensures that code runs in an isolated environment and helps protect developer machines against supply-chain attacks, malware, and similar threats. The Dev Container should work with Cypress. If you use the `cy:open` script command, you will need to open http://localhost:6080/vnc.html in your browser to see the Cypress Interface (The url is also printed in the terminal/shell when you run the command)

AikidoSec safe-chain is used to wrap the commands for yarn, npm, etc in the container; preventing the use of packages published within the last five days, except for DHIS2 packages (we can change the number of days, let's just agree on something, but 5-7 days is good in my book). This helps protect developers from downloading malware and reduces the risk of publishing a compromised app.

If you want other extensions installed in the docker container: 1) Make a case for it and we can add it to the list of extensions that should be auto-installed for the Capture App 2) Add the extension to your personal list of extensions that are always installed when using a Dev Container (user config) 3) Install them manually on each container rebuild (you should not need to rebuild often).

Updated the CI workflows to use safe-chain and other best practices for caching. Update some actions using deprecated node versions to the newest ones.


.nvmrc
```diff
@@ -0,0 +1 @@
+24
```
simonadomnisoru, line 1: Nice! Glad to finally drop Node 18 and upgrade to 24 🥳 !

package.json
```diff
@@ -74,8 +74,8 @@
         "jsdoc": "NODE_ENV=development jsdoc -c jsdoc-conf.json",
         "linter:check": "yarn workspaces run linter:check && yarn workspaces run build && eslint -c .eslintrc . --quiet",
         "tsc:check": "./node_modules/.bin/tsc --noEmit && yarn workspaces run tsc:check",
-        "cy:open": "export NODE_OPTIONS=--openssl-legacy-provider && concurrently --kill-others \"yarn start:forCypress\" \"wait-on 'http-get://127.0.0.1:3000' && cypress open\"",
-        "cy:run": "export NODE_OPTIONS=--openssl-legacy-provider && concurrently --kill-others \"yarn start:forCypress\" \"wait-on 'http-get://127.0.0.1:3000' && cypress run\"",
+        "cy:open": "[ \"$IN_DEVCONTAINER\" = \"true\" ] && ./scripts/package/cy-open-devcontainer.sh || ./scripts/package/cy-open-local.sh",
```
simonadomnisoru, line 77: I'm getting login errors when opening and running Cypress tests using the VNC link http://localhost:6080/vnc.html. I tried it both inside VS Code and in an external browser. Any idea what I am doing wrong?

https://github.com/user-attachments/assets/4d243450-45c2-4414-a3c2-3245a64a1c4b

Thank you!

simonadomnisoru (approved):
Even after refreshing the cache, when opening Cypress, I still get the [login error](https://github.com/dhis2/capture-app/pull/4590#discussion_r3333546069) inside the devContainer. 
But, since Cypress works fine in the CI and also locally outside the devContainer. I think we can leave it as-is and merge.

PR #4597, opened 2026-06-08
fix: plugin issues

superskip:
There is no Jira ticket for this, as it is tricky to reproduce the issues; they were all observed in a particular production instance. The problems went away after these changes were made :)

I think code review should be possible as is. Also not a lot of changes, and if you want bite-sized chunks you could check out the individual commits separately.


simonadomnisoru (approved):
It's difficult to properly review this PR without having reproduction steps and the full context of the issue. However, the code changes look good on their own, and I don't see any risky behavior, so I'll go ahead and approve.

PR #4598, opened 2026-06-09
ci: verified signatures for dhis2-bot commits

JoakimSM:
Makes dhis2-bot commits verified (Prerequisites: v1 of https://github.com/dhis2/action-commit-signing must be released, the DHIS2_BOT_SSH_SIGNING_KEY secret must be available and the dhis2_bot account must know its public key)

### DIFF
--- a/.github/workflows/verify-app.yml
+++ b/.github/workflows/verify-app.yml
@@ -228,6 +228,7 @@ jobs:
       - uses: dhis2/action-commit-signing@v1
         with:
           ssh-signing-key: ${{ secrets.DHIS2_BOT_SSH_SIGNING_KEY }}
+          ssh-signing-key-passphrase: ${{ secrets.DHIS2_BOT_SSH_SIGNING_PASSPHRASE }}
 
       - uses: dhis2/action-semantic-release@use-yarn-version-prepare
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-app.yml
```diff
@@ -225,6 +225,10 @@ jobs:
           path: '**/node_modules'
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
 
+      - uses: dhis2/action-commit-signing@v1
+        with:
+          ssh-signing-key: ${{ secrets.DHIS2_BOT_SSH_SIGNING_KEY }}
```
radnov, line 230: The default `dhis2-bot` signing key is passphrase protected, so you need to set the `ssh-signing-key-pasphrase` input, too.
```suggestion
          ssh-signing-key: ${{ secrets.DHIS2_BOT_SSH_SIGNING_KEY }}
          ssh-signing-key-passphrase: ${{ secrets.DHIS2_BOT_SSH_SIGNING_PASSPHRASE }}
```

radnov (changes requested):

.github/workflows/verify-app.yml
JoakimSM: Ok, thanks

PR #4600, opened 2026-06-09
fix: [DHIS2-21618] Fix form layout for scheduled events

henrikmv:
[DHIS2-21618](https://dhis2.atlassian.net/browse/DHIS2-21618)

### Summary
Fixes the layout of the schedule event form so it renders better in narrow/responsive widget widths. The schedule date field and its label now stack and wrap more gracefully, and the info box no longer shows a title.

### Changes
- **ScheduleDate** (`ScheduleDate.component.tsx`): Field wrapper now uses a column direction by default. Added a dedicated `autoScheduledWrapper` (row + wrap) used when the due date is hidden, and gave the field label a `flexBasis`/`minWidth` so it wraps cleanly instead of squeezing the input.
- **InfoBox** (`InfoBox.component.tsx`): Removed the redundant "Schedule date info" `NoticeBox` title.
- **WidgetEventSchedule** (`WidgetEventSchedule.component.tsx`): Dropped the fixed `minWidth: '300px'` on the wrapper so the widget can shrink to fit its container.

[DHIS2-21618]: https://dhis2.atlassian.net/browse/DHIS2-21618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):
Nice UX improvement 🥳 !

PR #4605, opened 2026-06-10
feat: [DHIS2-20842] Deactivate tracked entity

henrikmv:
[DHIS2-20842](https://dhis2.atlassian.net/browse/DHIS2-20842)

Adds deactivate/reactivate from the enrollment dashboard, and makes the app respect the `inactive` state: write access locked while inactive, and inactive entities dimmed + non-selectable in working lists.

## Implementation notes 

1. **Read-only is enforced centrally** in `EnrollmentAccessContext` — when inactive, all write access (program, TET, every stage) is forced `false` at the source, so widgets go read-only with no per-widget changes. Exception: `canToggleTrackedEntityStatus` uses the raw TET write access, so the toggle survives the read-only state it creates (otherwise you could never reactivate).

2. **`inactive` is stored in redux, fed by the existing guarded enrollment query** — not a separate query. It inherits that query's freshness (the DHIS2-21152 guard + `staleTime:0`/`cacheTime:0`), so it's as fresh as the enrollment with no extra request. Because the query is guarded (`storedEnrollmentId !== enrollmentId`) and won't refetch on the same enrollment, the toggle updates the flag **optimistically via dispatch** (`setTrackedEntityInactiveStatus(!current)`) instead of invalidating.

3. **`readOnly` is gated on** `enrollmentId === storedEnrollmentId`, so a previous entity's status is never applied during a transition.

4. **Toggle = partial UPDATE** (`{trackedEntity, trackedEntityType, orgUnit, inactive}`); the menu item is hidden unless the TEI has both `orgUnit` and a TET id.

5. **Working lists exclude inactive rows from selection** (select-all, row-click, checkbox), plumbed through both `convertToClientTeis` and `useDataSource`.

Cypress test to be implemented in https://github.com/dhis2/capture-app/pull/4607!

[DHIS2-20842]: https://dhis2.atlassian.net/browse/DHIS2-20842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.tsx
```diff
@@ -1,28 +1,161 @@
...
 import { FlyoutMenu, IconMore16, MenuItem } from '@dhis2/ui';
 import i18n from '@dhis2/d2-i18n';
 import type { PlainProps } from './OverflowMenu.types';
 import { DeleteMenuItem, DeleteModal } from './Delete';
+import { DeactivateMenuItem } from './Deactivate/DeactivateMenuItem';
+import { DeactivateModal } from './Deactivate/DeactivateModal';
 import { OverflowButton } from '../../Buttons';
 import { TrackedEntityChangelogWrapper } from './TrackedEntityChangelogWrapper';
 
+type MenuFlyoutProps = {
```
simonadomnisoru: Please move the `MenuFlyoutProps` type definition in the `OverflowMenu.types.ts` file

src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useDataSource.ts
```diff
@@ -79,6 +79,7 @@ export const useDataSource = (
                     return acc;
                 }, {});
 
+            (listRecord as any).inactive = Boolean(eventRecord.inactive);
```
simonadomnisoru: Is there any reason why the `inactive` isn't computed along with the rest of the values in the reducer above when applying the `convertClientToList`?

src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.tsx
```diff
@@ -1,28 +1,161 @@
-import React, { useState } from 'react';
+import React, { useState, type Dispatch, type SetStateAction } from 'react';
 import { FlyoutMenu, IconMore16, MenuItem } from '@dhis2/ui';
 import i18n from '@dhis2/d2-i18n';
 import type { PlainProps } from './OverflowMenu.types';
 import { DeleteMenuItem, DeleteModal } from './Delete';
+import { DeactivateMenuItem } from './Deactivate/DeactivateMenuItem';
+import { DeactivateModal } from './Deactivate/DeactivateModal';
```
simonadomnisoru: Replace with `import { DeactivateModal, DeactivateMenuItem } from './Deactivate';` and add an `index.ts` file in the Deactivate folder with the corresponding exports.

src/core_modules/capture-core/components/List/OnlineList/OnlineList.component.tsx
```diff
@@ -27,6 +28,12 @@ const getStyles: Readonly<any> = {
             alignItems: 'flex-end',
         },
     },
+    inactiveRow: {
```
simonadomnisoru, line 31: How should we handle the `Inactive` column in the WLs? Should it display a Yes/No value, or perhaps it brings no value, and we should hide it completely?

Image

Thanks!

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.tsx
```diff
@@ -202,6 +205,17 @@ const WidgetProfilePlain = ({
         trackedEntity: trackedEntity ? (trackedEntity.trackedEntity || teiId) : teiId,
     }), [trackedEntity, teiId]);
 
+    const trackedEntityForToggle = useMemo(() => {
+        const trackedEntityTypeId = program?.trackedEntityType?.id;
+        const teiOrgUnit = (trackedEntity as any)?.orgUnit;
+        if (!trackedEntity || !trackedEntityTypeId || !teiOrgUnit) return null;
+        return {
+            trackedEntity: trackedEntity.trackedEntity || teiId,
```
simonadomnisoru: Isn't trackedEntity already computed in `trackedEntityProp` above? We might be able to clean a bit and just use a single `useMemo` here.

src/core_modules/capture-core/components/WidgetProfile/OverflowMenu/OverflowMenu.component.tsx
```diff
@@ -1,28 +1,161 @@
...
+                canWriteData={canWriteData}
+                canCascadeDeleteTei={canCascadeDeleteTei}
+                setActionsIsOpen={setActionsIsOpen}
+                setDeleteModalIsOpen={setDeleteModalIsOpen}
+            />
+        )}
+    </FlyoutMenu>
+);
+
+type ModalsProps = {
```
simonadomnisoru: Please move the `ModalsProps` type definition in the `OverflowMenu.types.ts` file

src/core_modules/capture-core/components/WorkingLists/WorkingListsCommon/hooks/useDataSource.ts
henrikmv: The reducer only loops over visible columns and converts each column’s value into a display string via `convertClientToList`.

`inactive` is different from those values - it’s not a column, but a top-level flag on the record. So I don’t think it should go in the reducer, but happy to change my mind.

Thanks!

simonadomnisoru (approved):

PR #4619, opened 2026-06-18
fix: [DHIS2-17544] enforce org unit opening/closing dates in Capture

karolinelien:
## What & why

[DHIS2-17544](https://dhis2.atlassian.net/browse/DHIS2-17544) — Org unit **opening/closing dates were never enforced** in the Capture app, so users could enroll people and add/edit events with dates outside the selected org unit's open/close range. (Verified the dhis2-core tracker importer doesn't enforce this either — see _Follow-up_ below.)

## What this does

Enforces the range — **both bounds inclusive, day granularity, calendar-safe** (handles non-Gregorian calendars) — on the date fields that represent activity at an org unit, **each validated against its own org unit**:

| Field | Flows |
|---|---|
| Enrollment date | new + edit |
| Event occurred date | new + edit |
| Event scheduled/due date | scheduling |
| Enrollment transfer | hard-block when the enrollment date is outside the **destination** org unit's range |

Each constrained field disables out-of-range dates in the calendar picker **and** adds a blocking validator (existing "no future date" limits are combined with the closing date). Org units with no opening/closing date leave that side unconstrained.

### Key decisions
- **Hard block** on both new and existing records. Existing out-of-range records can't be saved until the date is corrected — a deliberate product call (documented risk below).
- **Incident date is intentionally NOT constrained** — it doesn't represent activity at the org unit (can legitimately be a future/onset date).
- **Transfer** checks the **enrollment date only** against the destination; blocks via a disabled button + error notice.

### Implementation
- New `getWithinOrgUnitDateRangeValidator` + `isIsoDateWithinOrgUnitRange` (shared boundary core) in form validators.
- New shared `getOrgUnitCalendarBounds` helper for calendar min/max.
- `calendarMin` plumbed through the shared date component to `@dhis2/ui` `CalendarInput`.
- `CoreOrgUnit` type + both org-unit fetch paths now request `openingDate,closedDate`; event/transfer flows enrich their selected org unit via `useCoreOrgUnit`.

## Testing
- Unit tests for the validator + ISO helper (within range, both boundaries inclusive, single-bound, empty/no-dates).
- `tsc --noEmit`: clean. ESLint: clean (local `linebreak-style` noise is the Windows CRLF checkout only).

## Known risks (deliberate)
1. Existing out-of-range records are uneditable until the date is corrected.
2. A blocked **transfer** has no in-app remediation — only an admin widening the destination org unit's dates enables it.

## Not in this PR
- **Backend companion** (dhis2-core `DateValidator` org-unit-range check) — the importer currently accepts out-of-range dates, so API/import/Android still bypass the rule. To be filed separately.
- P2: inline working/line-list date edits; disabling fully-closed org units in selection.

AI Assisted

[DHIS2-17544]: https://dhis2.atlassian.net/browse/DHIS2-17544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/DataEntries/Enrollment/fieldValidators/enrollmentDate.validatorContainersGetter.ts
```diff
@@ -24,6 +30,10 @@ export const getEnrollmentDateValidatorContainer = () => {
         { validator: isValidNonFutureDate,
             errorMessage: i18n.t('A date in the future is not allowed'),
         },
+        {
+            validator: getWithinOrgUnitDateRangeValidator(orgUnit),
+            errorMessage: '',
```
ameenhere, line 36: Something is not right here. I am guessing there should have been a i18n string for error message here as well if the date is outside range.

karolinelien: Claude: "Good question — it's already internationalized, just not as a static container string. The message is built dynamically inside the validator (getWithinOrgUnitDateRangeValidator → buildErrorMessage), where every variant goes through i18n.t(...) and interpolates the org unit's opening/closing dates and label. That's why the validator container's static errorMessage is intentionally left ''.

The framework picks the dynamic message over the static one: getValidationError in dataEntryField.utils.ts uses result.errorMessage || validatorContainer.errorMessage, so the static '' is only a fallback for when a validator returns no message — and this one always returns a non-empty, translated message on failure. It's the same pattern validateNotExpired already uses. I've added a clarifying comment at the container so this isn't mistaken for a missing string. Let me know if you'd still prefer a static string as well."

PR #4595, opened 2026-06-01
feat: [DHIS2-21266] Integrate react Markdown lib

fernandogmz:
- Integrate react markdown library to support standard md elements
- Apply corresponding legend sets for rule actions where apply
- Create Markdown e2e testing for markdown integration

PR #4621, opened 2026-06-18
feat: [DHIS2-21266] Integrate react Markdown lib in WidgetFeedback

simonadomnisoru:
[DHIS2-21266](https://dhis2.atlassian.net/browse/DHIS2-21266)

The work in the PR belongs to @fernandogmz and was implemented in https://github.com/dhis2/capture-app/pull/4595. 

I am using this branch as a workaround to run CI checks, since CI currently doesn't trigger on external fork PRs. We are aware that we need a better process to accommodate and encourage external contributions. Following an internal discussion, we've decided to stick with this workaround for now rather than investing time into optimizing the process.

[DHIS2-21266]: https://dhis2.atlassian.net/browse/DHIS2-21266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/rules/RuleEngine/helpers/rulesEffectsProcessor.ts
```diff
@@ -175,6 +175,8 @@ export function getRulesEffectsProcessor(
             displayText: {
                 id: effect.id,
                 message: `${message} ${sanitiseFalsy(effect.data)}`,
+                priority: effect.priority,
```
henrikmv: `priority` is copied onto each feedback item but never read. Ordering is already done at the rule level in [RulesEngine.ts:91](https://github.com/dhis2/capture-app/blob/e990bb845c1c866acccde908b0cf66a373c08894/packages/rules-engine/src/RulesEngine.ts#L91).

src/core_modules/capture-core/rules/RuleEngine/helpers/rulesEffectsProcessor.ts
```diff
@@ -190,6 +192,8 @@ export function getRulesEffectsProcessor(
                 id: effect.id,
                 key: effect.displayContent,
                 value: typeof data === 'number' ? numberToString(data) : String(data),
+                priority: effect.priority,
```
henrikmv: Same as [the above comment](https://github.com/dhis2/capture-app/pull/4621/changes#r3475641042).

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/MarkdownWrapper/MarkdownWrapper.tsx
```diff
@@ -0,0 +1,166 @@
...
+                { color && <span className={classes.legend} style={{ backgroundColor: color }} /> }
+                <div className={classes.markdown}>
+                    <Markdown remarkPlugins={[remarkGfm]}>{content}</Markdown>
+                </div>
+            </div>
+        }
+    </div>
+);
+
+export const MarkdownWrapper = withStyles(styles)(MarkdownWrapperComponent) as ComponentType<MarkdownWrapperProps>;
```
henrikmv, line 169: `MarkdownWrapper` renders 1–2 `<Markdown>` instances per row, which may re-parse on each parent render even when the text hasn’t changed

Wrapping it in `memo` might be a cheap way to skip re-parsing every feedback row when parent re-renders. WDYT?

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.tsx
henrikmv: `String items` render raw while `object items` go through Markdown. Is that intended?

henrikmv (commented):
Thanks for the contribution @fernandogmz, and thank you @simonadomnisoru for creating this PR as a workaround for the blocking security rule on our CI checks. 

I have left a few comments. Thank you!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.tsx
fernandogmz: Hi @henrikmv 
For this, in which cases will this block will be executed:  `} else if (typeof rule === 'string') { return renderString(rule, index); }` ? Currently, both `DisplayText` and `DisplayKeyValuePair` are handled correctly.

src/core_modules/capture-core/rules/RuleEngine/helpers/rulesEffectsProcessor.ts
fernandogmz: This was added for debugging, it is no longer necessary. Removing it.

src/core_modules/capture-core/rules/RuleEngine/helpers/rulesEffectsProcessor.ts
fernandogmz: This was added for debugging, it is no longer necessary. Removing it.

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/MarkdownWrapper/MarkdownWrapper.tsx
fernandogmz: Agree. Wrapping the component in `memo` since the markdown content won't change.
Also, I have moved the remark plugin definition out of the component to prevent unnecesary re-renders since it is an array:
`const REMARK_PLUGINS = [remarkGfm];`

fernandogmz:
@henrikmv @simonadomnisoru I have pushed a new commit on the fork branch, please update this one as needed.

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.tsx
henrikmv: Good question, @fernandogmz!

Seems like both feedback producers only emit objects, either `displayText` or `displayKeyValuePair`. The store path reads those same objects from the reducer, so every item resolves to either `renderKeyValue` or `renderTextObject`.

That means the string case is never reachable, so we can remove `renderString` and drop `string` from the union. WDYT?

Thanks!

simonadomnisoru: Great catch! 🙌 

I agree that it is safe to clean the `string` from the union. There are currently no metadata combinations that will enter that `if` case. I also confirmed in the original ticket's design and only `displayText` or `displayKeyValuePair` are mentioned ([DHIS2-11212](https://dhis2.atlassian.net/browse/DHIS2-11212) and [DHIS2-10981](https://dhis2.atlassian.net/browse/DHIS2-10981)).

Since this is unrelated to the Markdown feature, I went ahead and cleaned it up myself in both the `WidgetFeedback` and `WidgetIndicator`.

Thank you both!

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.types.ts
```diff
@@ -2,16 +2,20 @@ export type FilteredFeedbackText = {
...
+    priority?: number;
+    legendSetId?: string;
 }
 
 export type FilteredFeedbackKeyValue = {
     id: string;
     key: string;
     value: string;
     color?: string;
+    priority?: number;
```
henrikmv: `priority` can be removed.

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.types.ts
```diff
@@ -2,16 +2,20 @@ export type FilteredFeedbackText = {
     id: string;
     message: string;
     color?: string;
+    priority?: number;
```
henrikmv: `priority` can be removed.

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedback.types.ts
```diff
@@ -2,16 +2,20 @@ export type FilteredFeedbackText = {
...
 }
 
 export type FilteredFeedbackKeyValue = {
     id: string;
     key: string;
     value: string;
     color?: string;
+    priority?: number;
+    legendSetId?: string;
 }
```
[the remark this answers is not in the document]
simonadomnisoru: The `string` cleanup was already validated and discussed in https://github.com/dhis2/capture-app/pull/4621#discussion_r3490426957

src/core_modules/capture-core/components/WidgetFeedback/WidgetFeedbackContent/WidgetFeedbackContent.tsx
```diff
@@ -52,43 +59,35 @@ const WidgetFeedbackContentComponent = ({ feedback, feedbackEmptyText, classes }
...
-    const renderString = (item: string, index: number) => (
-        <li
-            key={index}
-            className={classes.listItem}
-        >
-            {item}
-        </li>
-    );
+    const renderKeyValue = (item: FilteredFeedbackKeyValue) => {
+        const color = resolveColor(item.legendSetId, item.value);
```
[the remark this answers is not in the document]
simonadomnisoru: This is consistent with the old code where `item.color` was also never rendered in the feedback widget. I confirm it in the UX designs from [DHIS2-11212](https://dhis2.atlassian.net/browse/DHIS2-11212)

henrikmv (approved):
Thank you @fernandogmz and @simonadomnisoru!

linadhis2:
LGTM from QA perspective.

PR #4622, opened 2026-06-22
fix: [DHIS2-21669] omit inaccessible data elements in enrollment dashboard

karolinelien:
## Summary

Fixes [DHIS2-21669](https://dhis2.atlassian.net/browse/DHIS2-21669): a **non-superuser** without metadata read access to a data element used by a program stage crashed the **Enrollment dashboard** (`TypeError: Cannot read properties of undefined (reading 'id')`).

Confirmed the fix fixes the crash and loads the enrollment dashboard.

## Root cause

The enrollment dashboard loads program metadata and data-element metadata via two separate queries and joins them assuming every stage-referenced data element is present. When a user lacks access to a data element, the **program** query still returns the reference, but the **data element** collection query filters it out by sharing — so it is absent from the dictionary. `buildProgramMetadata` then dereferenced the missing data element and threw.

Verified against a real instance: as a non-superuser, the program query returns the data element id while `GET /api/dataElements?filter=id:in:[…]` omits it (and a direct `GET` is `404`).

## Change

- `useProgramMetadata`: extract the transform into an exported `buildProgramMetadata`, and filter out program-stage data elements absent from the loaded dictionary (logging each). The inaccessible data element is omitted — matching the ticket's expected result ("dashboard loads without showing DE").
- `useProgramStages`: guard the consumer (`displayInReports && dataElement`).
- `WidgetProfile` was checked and intentionally left untouched — its live program query returns the full data element inline, so it never crashed.

## Testing

- New `useProgramMetadata.test.js`: omits an inaccessible data element without throwing, logs it, and still maps accessible data elements + option sets. Passes.
- ESLint clean on changed files.

AI Assisted

[DHIS2-21669]: https://dhis2.atlassian.net/browse/DHIS2-21669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):

linadhis2:
LGTM from the QA check

PR #4629, opened 2026-06-26
fix: [DHIS2-21708] Org unit selector is expanded when form open if no org unit in top bar

henrikmv:
[DHIS2-21708](https://dhis2.atlassian.net/browse/DHIS2-21708)

Refactors the `SingleOrgUnitSelectField` from an always-visible org unit tree into a collapsed dropdown with a popover. When no value is selected, the user sees a combobox trigger with a chevron icon. Clicking it opens a `Popover` containing the org unit tree, with a debounced search input embedded in the trigger.

Key changes:

- `SingleOrgUnitSelectField`: replaces inline tree with a trigger + `Popover` pattern, adds keyboard navigation, debounced search, and open/close state management.
- `OrgUnitField`: adds `hideSearchField` and `searchText` props so the parent can control search externally.

[DHIS2-21708]: https://dhis2.atlassian.net/browse/DHIS2-21708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/SingleOrgUnitSelectField.component.tsx
```diff
@@ -70,20 +187,80 @@ class SingleOrgUnitSelectFieldPlain extends React.Component<Props, SingleOrgUnit
...
+                </div>
+                {open && !disabled && this.renderPopover()}
+            </React.Fragment>
         );
     }
 
     render() {
         const { value } = this.props;
-        return value ? this.renderSelectedOrgUnit(value) : this.renderOrgUnitField();
+        return value ? this.renderSelectedOrgUnit(value) : this.renderCollapsedOrgUnitField();
```
simonadomnisoru: Was this bug confirmed by Product or UX? I am asking because it doesn't look like a bug to me. The org unit component is expanded throughout the entire app (e.g in forms with OrgUnit TEAs or DEs, when adding a new relationship, or when linking events, etc.). The only place that comes to mind where it is currently collapsed is the Context Top bar. So, in general, we seem to tend to have it expanded by default if there is space. I've attached a few examples where it appears expanded, but there are many others.


Image
Image
Image


Thank you!

karolinelien: Thanks for checking @simonadomnisoru! We have decided we will collapse it to keep it consistent with the other value types/fields in the form which are always collapsed.

src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/SingleOrgUnitSelectField.component.tsx
```diff
@@ -23,24 +79,92 @@ type OrgUnitValue = {
...
+};
+
+const AutoSelectSingleOrgUnit = ({ onAutoSelect }: AutoSelectSingleOrgUnitProps) => {
+    const { data: autoSelectOrgUnits } = useOrgUnitAutoSelect();
+    const hasPreselected = React.useRef(false);
+
+    React.useEffect(() => {
+        if (!hasPreselected.current && (autoSelectOrgUnits as any)?.length === 1) {
+            hasPreselected.current = true;
+            const orgUnit = (autoSelectOrgUnits as any)[0];
```
simonadomnisoru: Could we try to avoid using `as any`? We should have access to the exact type returned by the `useOrgUnitAutoSelect` hook

src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/SingleOrgUnitSelectField.component.tsx
```diff
@@ -53,9 +177,46 @@ class SingleOrgUnitSelectFieldPlain extends React.Component<Props, SingleOrgUnit
 
     onDeselectOrgUnit = () => {
         this.props.value && this.setState({ previousOrgUnitId: this.props.value.id });
+        this.hasDeselected = true;
+        this.shouldOpenOnClear = this.props.autoSelectSingleOrgUnit !== false;
```
simonadomnisoru: `hasDeselected` and `shouldOpenOnClear` are plain instance variables, so mutating them won't trigger a re-render, even though they're used directly in `render()` function. Is this intentional, or should we move them to `this.state` instead?

src/core_modules/capture-core/components/FormFields/New/Fields/OrgUnitField/SingleOrgUnitSelectField.component.tsx
```diff
@@ -70,20 +231,97 @@ class SingleOrgUnitSelectFieldPlain extends React.Component<Props, SingleOrgUnit
...
+    }
+
+    renderCollapsedOrgUnitField = () => {
+        const { classes, disabled } = this.props;
+        const { open, inputValue } = this.state;
+        const triggerClassName = [
+            classes.trigger,
+            open && classes.triggerOpen,
+            disabled && classes.triggerDisabled,
+        ].filter(Boolean).join(' ');
```
simonadomnisoru: For custom `className`, could you please use `cx from @emotion/css`? It helps keep things consistent since a lot of our other components already follow this pattern

simonadomnisoru (approved):

PR #4630, opened 2026-06-26
fix: [DHIS2-21693] Stop auto switching org unit in context selector

henrikmv:
[DHIS2-21693](https://dhis2.atlassian.net/browse/DHIS2-21693)

**IMPORTAN BEHAVIOUR CHANGE:**
> From now on, the org unit selector should reflect the user’s current context, rather than the org unit the TEI belongs to. Since the org unit can now be set, changed, or removed directly inside the form, validating the org unit before opening the form is no longer required.

This PR makes the context (top bar) org unit only ever change from explicit user action — it removes the auto-switch behavior on the enrollment page, removes the now-unnecessary org unit validation against the current program on the EnrollmentAddEvent page, and cleans up the remaining places that auto-selected the org unit.

**Enrollment page — remove auto-switch**
- Removes `autoSwitchOrgUnitEpic` and its helper `captureScopeQuery` (and its registration)
- Removes `programOwnerId` from the `saveEnrollments` action payload (the removed epic was its only consumer)

**EnrollmentAddEvent — remove org unit validation**
- Removes the `ORG_UNIT_INVALID` status and its associated UI/validation logic
- Drops `orgUnitId` from the `useValidatedIDsFromCache` call

**Keep the user's context org unit stable elsewhere**
- EnrollmentEditEvent: preserve `orgUnitId` when navigating back to the enrollment dashboard (save/back/cancel) — previously dropped, which blanked the top bar once the auto-switch no longer repopulated it
- Tracker working lists: clicking a record no longer falls back to the record's own org unit when none is selected in the top bar
- Search: a unique-ID search result no longer sets the context org unit to the found TEI's org unit

**Note:** the single-org-unit auto-select on app init is intentionally **kept** — when a user has exactly one capture-scope org unit there is no real choice to make, so it is still pre-selected.

[DHIS2-21693]: https://dhis2.atlassian.net/browse/DHIS2-21693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JoakimSM (approved):
LGTM 👍

PR #4636, opened 2026-07-01
fix: [DHIS2-21741] Failed event deletion is not rolled back in the UI

henrikmv:
[DHIS2-21741](https://dhis2.atlassian.net/browse/DHIS2-21741)

Simplifies the delete-event rollback mechanism in `DeleteActionModal` by replacing the React Query cache lookup with a direct prop.

The delete removes the event from Redux, but the rollback previously read the event data from a separate React Query cache. That cache can be stale, and newly created in-session events are not always present there, which meant rollback could be skipped on delete failure.

Previously, on mutation start the component reconstructed a query key from `teiId`, `programId`, and `enrollmentId`, then dug into the cached enrollment data to find the event for rollback. Now, the already-rendered `eventDetails` object from `EventRow` is passed in as a prop and used directly:

```diff
- queryClient.getQueryData(key)
-   ?.enrollments
-   ?.flatMap(...)
-   ?.find(e => e.event === eventId)
+ const eventToRollbackOnFail = eventDetails;
```

This ensures the event is always available for rollback and restored on failure, including newly created in-session events.

This also removes the `useQueryClient` hook, the `ReactQueryAppNamespace` import, and three individual ID props in favor of a single typed `eventDetails: ApiEnrollmentEvent` prop.

[DHIS2-21741]: https://dhis2.atlassian.net/browse/DHIS2-21741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/EventRow/DeleteActionModal/DeleteActionModal.tsx
+++ b/src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/StageDetail/EventRow/DeleteActionModal/DeleteActionModal.tsx
@@ -3,17 +3,14 @@ import log from 'loglevel';
 import i18n from '@dhis2/d2-i18n';
 import { Button, ButtonStrip, Modal, ModalActions, ModalContent, ModalTitle } from '@dhis2/ui';
 import { useAlert, useDataEngine } from '@dhis2/app-runtime';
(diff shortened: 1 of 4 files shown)


simonadomnisoru (approved):
Nice!

linadhis2:
LGTM from QA perspective.

PR #4647, opened 2026-07-20
fix: [DHIS2-21685] show min characters filter validation only on update attempt

henrikmv:
[DHIS2-21685](https://dhis2.atlassian.net/browse/DHIS2-21685)

Fixes the min-characters filter validation so that the error message is only shown after the user attempts to apply/update the filter, rather than on every blur event.

The key change in `withMinCharsToSearchValidation`:

```diff
- if (isBlur) { setCommittedValue(value); /* shows error on any blur */ }
+ const errorsVisibleRef = useRef(false);
+ // showValidationErrors (invalid update attempt) sets errorsVisibleRef.current = true
+ if (errorsVisibleRef.current) { setCommittedValue(value); /* live revalidation once shown */ }
```

This prevents premature validation feedback while the user is still typing/navigating between fields. Once an invalid update attempt has shown the error, validation turns live for the rest of the popup session, so the error clears as soon as the value becomes valid instead of shifting the layout mid-click on the Update button.

[DHIS2-21685]: https://dhis2.atlassian.net/browse/DHIS2-21685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):
Niice!

simonadomnisoru (approved):

PR #4648, opened 2026-07-20
docs: [DHIS2-15916] clarify ScopeSelector top bar context

karolinelien:
## Summary

Documentation-only change addressing [DHIS2-15916](https://dhis2.atlassian.net/browse/DHIS2-15916) ("Review ScopeSelector documentation").

The ticket originated from field confusion about the enrollment dashboard top bar showing a **tracked entity UID instead of a readable name**. Investigation showed this is not a bug but an intentional fallback in `deriveTeiName` (`useTeiDisplayName.ts`): the label uses the first two tracked entity type attributes marked *Display in reports*, then the first two TET attributes, then the UID. That behaviour was undocumented, so the docs couldn't answer the users' question.

This PR adds that explanation to `docs/user/using-the-capture-app.md`.

## Changes

- **"What each context box shows"** — describes the org unit, program, tracked entity (type as title + instance name), and enrollment boxes.
- **"How the tracked entity display name is built"** — documents the exact derivation order and adds an implementer/admin note explaining how to avoid a UID appearing (mark attributes as *Display in reports*). This is the concrete resolution for DHIS2-15916.
- **"Clearing all selections"** — documents the reset-all action and its discard-changes confirmation.

## Notes / considered but not included

- Program-vs-tracked-entity-type scoping and the attribute option combo selector are already documented elsewhere in the same guide, so they were not duplicated here.
- Commit contains **only** the docs file. The repo's pre-commit hook regenerated `i18n/en.pot` (unrelated extractor-version churn — a docs edit adds no translatable strings), so that file was restored; the pre-push lint/tsc hook was bypassed since a markdown-only change cannot affect it (root `tsc --noEmit` passes clean).

AI Assisted

[DHIS2-15916]: https://dhis2.atlassian.net/browse/DHIS2-15916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/user/using-the-capture-app.md
+++ b/docs/user/using-the-capture-app.md
@@ -931,22 +931,22 @@ You can change your context by clicking the "x" button.
 
 The context selector shows one box for each level of your context, from left to right:
 
-- **Program** — the selected program. You only see programs that are shared with you and your user group, and if you have selected an organisation unit, that are assigned to that organisation unit.
-- **Organisation unit** — the organisation unit you are working in.
-- **Tracked entity** — this box is titled with the *tracked entity type* (for example "Person" or "Building"), and its value shows the display name of the specific tracked entity instance (see [How the tracked entity display name is built](#capture_tei_display_name) below).
-- **Enrollment** — the selected enrollment, labelled with its enrollment date.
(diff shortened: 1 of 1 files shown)


simonadomnisoru (approved):
Nice!

PR #4666, opened 2026-07-27
fix: [DHIS2-21871] preserve working-list view sharing on update

gqcorneby:
[DHIS2-21871](https://dhis2.atlassian.net/browse/DHIS2-21871)

### Summary
Saving a working list view (event filter, TEI filter, or program stage working list) wipes its sharing: user and user group accesses are cleared and public access resets to default.

### Root cause
The update epics send a full `PUT` (`type: 'replace'`) with only the legacy sharing fields (`publicAccess`, `externalAccess`, `userAccesses`, `userGroupAccesses`) and no `sharing` object. From api/41 those legacy fields are ignored, so a full replace resets sharing to defaults. Regression of DHIS2-8420 (whose fix, DHIS2-10493 / DHIS2-10651, added those legacy fields).

### Changes
- Switch the three update epics (`eventFilters`, `trackedEntityInstanceFilters`, `programStageWorkingLists`) from `type: 'replace'` to `type: 'json-patch'`, patching only `name` and the query criteria. Sharing and ownership stay out of the payload, so the server leaves them untouched. (This is DHIS2's documented partial update path; a plain `application/json` PATCH returns 415 here.)
- Add a small pure helper `toJsonPatchReplaceOps`, unit tested.
- Widen the `mutate` type in `capture-core-utils/types/global.ts` to include `'json-patch'`.
- Dropping `user` (owner) also fixes the non-owner 409 on save (DHIS2-13020).
- Update the e2e intercepts from `PUT` to `PATCH`.

### Testing
Verified on play 2.41, 2.43, 2.44: sharing preserved after saving a view. JSON Patch is documented back to 2.40.

[DHIS2-21871]: https://dhis2.atlassian.net/browse/DHIS2-21871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/WorkingLists/EventWorkingLists/EventWorkingListsUser/EventWorkingListsUser.js
+++ b/cypress/e2e/WorkingLists/EventWorkingLists/EventWorkingListsUser/EventWorkingListsUser.js
@@ -480,7 +480,7 @@ When(/^you update the view with the name (.+)$/, (_name) => {
     cy.get('[data-test="list-view-menu-button"]')
         .click();
 
-    cy.intercept('PUT', '**/eventFilters/**').as('updateEventFilter');
+    cy.intercept('PATCH', '**/eventFilters/**').as('updateEventFilter');
     cy.contains('Update view')
         .click();
     cy.wait('@updateEventFilter', { timeout: 30000 });
--- a/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.js
+++ b/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.js
@@ -485,7 +485,7 @@ When('you update the list with the name My custom list', () => {
     cy.get('[data-test="list-view-menu-button"]')
         .click();
 
-    cy.intercept('PUT', '**/trackedEntityInstanceFilters/**').as('editTrackedEntityInstanceFilters');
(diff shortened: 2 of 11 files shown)


henrikmv:
Hi, @gqcorneby!

Thanks for the work on this! We appreciate the effort and will give the PR a closer look soon. 

Thanks for your patience!

henrikmv:
Hi agin @gqcorneby!

Thanks for the contribution and for addressing both [DHIS2-13020](https://dhis2.atlassian.net/browse/DHIS2-13020) and [DHIS2-21871](https://dhis2.atlassian.net/browse/DHIS2-21871).

After discussing it internally, we’ve decided to go with [#4702](https://github.com/dhis2/capture-app/pull/4702), which fixes the same issues without introducing `PATCH`.

We’d prefer to consider adopting `PATCH` as part of a broader refactor so it can be introduced consistently across the app.

Thanks again for the contribution!

[DHIS2-13020]: https://dhis2.atlassian.net/browse/DHIS2-13020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-21871]: https://dhis2.atlassian.net/browse/DHIS2-21871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4668, opened 2026-07-28
fix: [DHIS2-21855] limit concurrent api requests

JoakimSM:
What caused the error was that too many api requests were fired at once from the org. unit component in the UI library.

The current org. unit component implementation in the UI library leaves a lot to be desired.

1. It makes two api requests for each root
2. Changing roots after the initial render crashes the component
3. Even if we made point 2. work, and we added "view more" roots functionality (only showing the first 100 at the start for example), org. data for the previous roots wouldn't be cached, so clicking "view more" would result in doing all the initial api requests once more in addition to api requests for the new roots. 

Given the above, and that a new org. unit component is currently being developed, I opted for just limiting the number of concurrent api requests a client can do as this should fix the error (if the client tries to fire more api requests, these will be put on hold until others are finished). The number of concurrent api requests is now capped at 100 (quite arbitrary number, feel free to suggest something else).

I will make a PR for the app-runtime later as I think having this limit makes sense in general. They might have some input on what the limit should be, if it should be configurable etc. so doing this package patch first.


simonadomnisoru (approved):
Nice temporary workaround until the new Org unit component gets released!

linadhis2:
LGTM from a QA check

PR #4533, opened 2026-04-29
docs: [DHIS2-20703] Fix typos, outdated terminology and content in Capture docs

karolinelien:
## What is fixed?

Comprehensive cleanup of the Capture documentation based on DHIS2-20703, covering `docs/user/using-the-capture-app.md` and `docs/developer/form-field-plugins/manual-setup.md`.

---

### Terminology & spelling sweep

- **`tracked entity instance` / `TEI` → `tracked entity`** across the user guide and the developer form-field-plugins doc. The **authority name** `Delete tracked entity instance and associated enrollments and events` is kept verbatim, since it is the actual authority string.
- **Relationship button labels updated to match the current app UI.** The app renders the tracked entity **type** name dynamically (e.g. *Person*, *Building*), so the hard-coded "Tracked Entity Instance" labels were replaced with a readable placeholder and a one-time clarifying note:
  - `Link to an existing [tracked entity type]`
  - `Create new [tracked entity type]`
  - `Save new [tracked entity type] and link` (the verb also changed in the app — it is now "Save new … and link", not "Create … and Link")
  - `Save as new [tracked entity type]`
  - `Delete [tracked entity type]`
  - This also replaced the old `${...}` template syntax with the more readable `[tracked entity type]` form.
- **Enrollment status corrected `Canceled` → `Cancelled`** to match the app UI.
- **British spelling:** `organization` / `Organization unit` → `organisation` / `Organisation unit` (including the mandatory column-header reference).
- **Removed em-dashes** (3), replaced with a full stop / semicolon / comma.

> ⚠️ The **Tracked Entity Audit Log** setting lives in the Metadata Management app (dhis2-core), not in this repo, so its exact current label could not be verified against source — worth double-checking on a live instance.

---

### Typos & spelling (15 fixes)

| Line | Was | Fixed |
|------|-----|-------|
| 273 | `enrollement` | `enrollment` |
| 304 | `pre-fielded` | `pre-filled` |
| 304 | `teId` | `teiID` |
| 340 | `attibutes` | `attributes` |
| 431 | `draging` | `dragging` |
| 451 | `fitered` | `filtered` |
| 451 | `rang` | `range` |
| 675 | `the the` | `the` |
| 744 | `entitiy` | `entity` |
| 744 | `retrurned` | `returned` |
| 796 | `the the` | `the` |
| 796 | `reorgainze` | `reorganize` |
| 947 | `###Quick actions` | `### Quick actions` |
| 1172 | `informations related with` | `information related to` |
[...]

PR #4677, opened 2026-07-30
docs: [DHIS2-21885] Recapture outdated enrollment dashboard screenshots

karolinelien:
## What is fixed?

Recaptures ~35 outdated screenshots in `docs/user/using-the-capture-app.md`'s enrollment-dashboard and list/sort sections, so they match the current Capture app UI. Split out from #4533 (which stays text-only) per earlier discussion.

All new screenshots follow a consistent style: full enrollment dashboard with a red box around the relevant widget/element (short crops only for dialog/message screenshots that aren't full dashboard height).

### Doc text changes (not just images)

- **"Selecting a program with enrollments" section**: removed the "multiple enrollments, choose one" chooser message and its screenshot. DHIS2 only allows one *active* enrollment per program at a time, so selecting a program you're already actively enrolled in now takes you straight to that enrollment's dashboard — there's no longer a scenario where this chooser appears. Replaced the paragraph with text describing the actual current behavior.
- **"Anna Jenkins" → "Anna Jones"**: the doc referenced a person ("Anna Jenkins") that doesn't exist on the shared demo instance used for these screenshots. Replaced with "Anna Jones" (already used throughout the rest of the section) so text and images stay consistent.
- **Removed the "Reset list" and "Go to full {stage} list" callouts** (and their screenshots) from the Stages and Events section. Checked against git history — both describe app behavior that no longer exists:
  - *Reset list* was fully implemented, then the entire button was removed from the codebase in a 2024 toolbar redesign (`012b60d3c`, DHIS2-17651).
  - *Go to full list* still exists in code but has been deliberately hidden via CSS since 2021 ("hide until fully implemented", `90dbff3c6`, DHIS2-11733) and was never finished.

  Neither is recoverable by changing test data — they simply aren't in the app anymore.

AI Assisted

https://dhis2.atlassian.net/browse/DHIS2-21885


simonadomnisoru (approved):
👏

PR #4682, opened 2026-07-31
docs: [DHIS2-21885] recapture remaining likely-outdated screenshots (batch 2)

karolinelien:
## What is fixed?

Second batch of fixes for https://dhis2.atlassian.net/browse/DHIS2-21885, following on from https://github.com/dhis2/capture-app/pull/4680 (first batch). Covers groups 1, 2, 3, 4, 6, 9, 10 from the doc-wide screenshot audit.

- **Search result cards** (Program + Tracked entity type scope): card layout was reworked in a March 2026 pass — confirmed live that the "Last updated" timestamp is gone from *program-scope* result cards specifically (tracked-entity-type-scope cards still show it). Recaptured all four action states: general list, View dashboard, View active enrollment, Re-enroll. The **Re-enroll** screenshot uses WHO RMNCH Tracker rather than Child Programme: Child Programme has `onlyEnrollOnce` set on the demo, so its Re-enroll button never leads anywhere genuine — WHO RMNCH actually supports re-enrollment, so it's a more honest example of the feature.
- **View-selector chips**: recaptured against real saved views (Malaria case registration event list, Malaria focus investigation TEI list without org unit, TB program's default enrollment-status views).
- **relationship_options.png**: confirmed live that the newer enrollment-dashboard relationship widget shows "Link to an existing {type}" (with type name) but plain "Create new" (no type name) — corrected the doc text to match exactly, since the old text claimed both had the type name.
- **Program rules execution**: the old example's configured rule (DOB < 1 year, on Child Programme) no longer exists there. Found the actual still-configured rule on WHO RMNCH Tracker ("The womans age is outside the normal range") and recaptured against it; rewrote the doc text to describe the rule generically instead of naming a scenario that isn't configured anymore.
- **Registration forms** (with/without enrollment): responsive-layout rework since the 2024 captures, labels/flow unchanged.
- **Search / register tracked entity**: recaptured within the actual "Adding a relationship" flow context, matching what the doc describes.
- **download_event_list.png**: refreshed 2022-era chrome and recreated the numbered (1)/(2) callouts pointing at the triple-dot menu and "Download data..." menu item, matching the original style.
- **enroll-existing-tei.png**: recaptured. The search-card **Re-enroll** button is still broken (tracked separately, see below), but the pre-filled "New Enrollment" form itself is reachable via a different, working path: **Enrollment actions → Add new** on a completed enrollment (only shown when the program doesn't have `onlyEnrollOnce`). Recaptured that flow against WHO RMNCH Tracker.

### Known issue (separate from this PR)

The search-card **Re-enroll** button (shown on `search-by-attributes-on-scope-program-overview-4.png`) still doesn't do what its label says — clicking it opens the tracked entity's existing completed enrollment instead of a new, pre-filled one. Tracked in JIRA: https://dhis2.atlassian.net/browse/DHIS2-21942. The `enroll-existing-tei.png` screenshot above uses the working **Enrollment actions → Add new** path instead, so the doc's example is honest even though the button-labeled-Re-enroll bug remains open.

AI Assisted


docs/user/resources/images/program-rules-on-creation-01.png
simonadomnisoru, line 1: In the new screenshot, the background is red, and the warning is no longer showing, likely because no birth date value is selected. Could we grab a new screenshot to ensure it aligns with the program rule warning example detailed in the [docs on line 288]( https://github.com/dhis2/capture-app/pull/4682/changes#diff-1a5919ed1ac425bce5fe1126ef801d781f2d0288088bcba3bdc61a53128096abR288)

karolinelien: Thanks for reviewing thoroughly, will do!

karolinelien: Updated!

simonadomnisoru (approved):
LGTM!

PR #4683, opened 2026-07-31
test: [DHIS2-13960] re-enable tracker working list assignee filter e2e test

karolinelien:
## What
Re-enables the `Show only teis with active enrollments and unassigned events using the filter` scenario in `TrackerWorkingListsUser.feature` (was `@skip` for https://dhis2.atlassian.net/browse/DHIS2-13960) and makes it deterministic across demo databases.

## Why now
DHIS2-13960 was closed as **Invalid** — the reported `/trackedEntities` assignee-filter inconsistency was confirmed not to be a real backend bug, so the scenario no longer needs to be disabled. A plain un-skip fails CI, though: the assertion checks a hardcoded result count, and demo DBs return a varying number of matching TEIs (CI saw `Found '4', expected '3'` across 2.40–2.43).

## How
Adapt the assertion to the actual `/trackedEntities` response instead of mocking it: 3 matching TEIs (4 rows including the header) — `UGB082875`, `ZDA984904`, `FSL05494` — rather than the previously hardcoded 2/3, per [review feedback](https://github.com/dhis2/capture-app/pull/4683#discussion_r3713275730).

AI Assisted

### DIFF
--- a/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.feature
+++ b/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.feature
@@ -32,8 +32,7 @@ Scenario: Show only teis with completed enrollments using the filter
   And for a tracker program the page navigation should show that you are on the first page
 
 Scenario: Show only teis with active enrollments and unassinged events using the filter
-  Given the tracked entities endpoint is stubbed with two active unassigned teis
-  And you open the main page with Ngelehun and Malaria focus investigation context
+  Given you open the main page with Ngelehun and Malaria focus investigation context
   When you set the enrollment status filter to active
   And you apply the current filter
   And you set the assginee filter to None
--- a/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.feature
+++ b/cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.feature
@@ -30,10 +30,10 @@ Scenario: Show only teis with completed enrollments using the filter
   And the list should display teis with a completed enrollment
   And rows per page should be set to 15
   And for a tracker program the page navigation should show that you are on the first page
-# DHIS2-13960: /trackedEntities filter by assignee results are not consistent
-@skip
+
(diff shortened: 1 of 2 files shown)


simonadomnisoru (commented):
Looks like the `Show only teis with active enrollments and unassinged events using the filter` is still failing in [all the versions](https://github.com/dhis2/capture-app/actions/runs/30823617039/job/91719615833?pr=4683).

karolinelien:
> Looks like the `Show only teis with active enrollments and unassinged events using the filter` is still failing in [all the versions](https://github.com/dhis2/capture-app/actions/runs/30823617039/job/91719615833?pr=4683).

Yup, it works locally, but not CI... Looking into whether it is straight forward to make some fixes.

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerWorkingListsUser/TrackerWorkingListsUser.feature
```diff
@@ -30,10 +30,10 @@ Scenario: Show only teis with completed enrollments using the filter
   And the list should display teis with a completed enrollment
   And rows per page should be set to 15
   And for a tracker program the page navigation should show that you are on the first page
-# DHIS2-13960: /trackedEntities filter by assignee results are not consistent
-@skip
+
 Scenario: Show only teis with active enrollments and unassinged events using the filter
-  Given you open the main page with Ngelehun and Malaria focus investigation context
+  Given the tracked entities endpoint is stubbed with two active unassigned teis
```
simonadomnisoru: I'd recommend avoiding a stub here. Since [DHIS2-13960](https://dhis2.atlassian.net/browse/DHIS2-13960?focusedCommentId=207145) was marked invalid by the BE, it seems more sensible to adapt the test to match the actual API response rather than stubbing. Specifically, we could update `the list should display teis with an active enrollment and unassigned events` to check for 3 items instead of 2:

```
Then('the list should display teis with an active enrollment and unassinged events', () => {
    const ids = [
        'UGB082875',
        'ZDA984904',
        'FSL05494',
    ];

    cy.get('[data-test="tracker-working-lists"]')
        .find('tr')
        .should('have.length', 4)
```

Image

karolinelien: Sounds good, updated!

simonadomnisoru (approved):

PR #4657, opened 2026-07-22
feat: [DHIS2-21875] Unify overflow menu between View Event page and Stages and Events widget

henrikmv:
[DHIS2-21875](https://dhis2.atlassian.net/browse/DHIS2-21875)

# Verification matrix by event status

**Baseline:** user has `canEditProgramStage`. Without it, no mutation actions appear at all (overflow menu section hidden).

| Event status | Edit event | Complete/Uncomplete | Skip/Unskip | Delete |
|---|---|---|---|---|
| **ACTIVE** | Editable if not expired ¹ | **Complete** — enabled if not expired ¹ | Hidden | Enabled if not expired ¹ |
| **COMPLETED** (`blockEntryForm=false`) | Editable if not expired ¹ | **Uncomplete** — enabled when user has `F_UNCOMPLETE_EVENT` **AND** (not expired **OR** has `F_EDIT_EXPIRED`) | Hidden | Enabled if not expired ¹ |
| **COMPLETED** (`blockEntryForm=true`) | Read-only — badge "This event has been completed"; must Uncomplete first | **Uncomplete** — enabled when user has `F_UNCOMPLETE_EVENT` **AND** (not expired **OR** has `F_EDIT_EXPIRED`) | Hidden | Enabled if not expired ¹ |
| **SCHEDULE** | Editable — user picks a valid `occurredAt` ² | Hidden | **Skip** — enabled if not expired ¹ | Enabled if not expired ¹ |
| **OVERDUE** | Editable — user picks a valid `occurredAt` ² | Hidden | **Skip** — enabled if not expired ³ | Enabled if not expired ³ |
| **SKIPPED** | Read-only — event is skipped | Hidden | **Unskip** — enabled if not expired ¹ | Enabled if not expired ¹ |

### ¹ Expiry rules

**"Not expired"** means the event's relevant date is inside the program's editing window:

- `occurredAt` for **ACTIVE / COMPLETED**
- `scheduledAt` for **SCHEDULE / OVERDUE / SKIPPED**

`F_EDIT_EXPIRED` bypasses expiry checks. However, it does not bypass other permission requirements. For example, **Uncomplete** still requires `F_UNCOMPLETE_EVENT`.

### ² SCHEDULE / OVERDUE edit exception

For **SCHEDULE** and **OVERDUE**, form editing checks expiry against the date currently entered in the form rather than the server-side date.

This means a user without `F_EDIT_EXPIRED` can still save the event by selecting a valid date inside the editing window.

[...]

PR #4690, opened 2026-08-05
feat: [DHIS2-21941] Self contained changelog widget

henrikmv:
[DHIS2-21941](https://dhis2.atlassian.net/browse/DHIS2-21941)

> Reason behind this initiative: The fourth implementation of the Changelog, introduced in https://github.com/dhis2/capture-app/pull/4657, highlighted the need to make it easier to reuse.

This PR makes changelog widgets self-contained by removing the `entityData` / `eventData` / `trackedEntityData` props that were passed down from parent components. Instead, the changelog now fetches current entity values directly from the tracker API via a new `useCurrentEntityValues` hook.

## Key changes

- **New `useCurrentEntityValues` hook** — fetches current values from `tracker/events/{id}` or `tracker/trackedEntities/{id}`, replacing the prop-based approach.

- **Improved `isLatestValue` detection** — previously compared `change.currentValue` against a prop-provided snapshot. Now validates against both the newest changelog entry per field **and** fresh API data.

- **Centralized `removeChangelogQueries`** — extracts duplicated cache invalidation into a shared utility, and adds invalidation for two new mutation sites: `CompletionMenuItem` and `SkipAction`.

- Opening the changelog now fires one extra tiny `tracker/{events|trackedEntities}/{id}?fields=…[…,value]` fetch, but it runs in parallel with the (slower) changelog list request, is cached with `staleTime: Infinity` so reopens are free, and is skipped entirely for changelogs without `FILE_RESOURCE/IMAGE` fields. In return, the widget becomes fully self-contained — no more entityData prop drilled from three different-shaped parent sources, and no more fragile `?.value` hack that silently broke on stale parent caches.

[DHIS2-21941]: https://dhis2.atlassian.net/browse/DHIS2-21941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/WidgetsChangelog/common/utils/removeChangelogQueries.ts
```diff
@@ -0,0 +1,15 @@
...
+import { ReactQueryAppNamespace } from '../../../../utils/reactQueryHelpers';
+import { CHANGELOG_ENTITY_TYPES } from '../Changelog/Changelog.constants';
+
+const removeChangelogQueries = (
+    queryClient: QueryClient,
+    entityType: typeof CHANGELOG_ENTITY_TYPES[keyof typeof CHANGELOG_ENTITY_TYPES],
+    entityId: string,
+) => queryClient.removeQueries([ReactQueryAppNamespace, 'changelog', entityType, entityId]);
+
+export const removeEventChangelogQueries = (queryClient: QueryClient, eventId: string) =>
```
simonadomnisoru: Overall, this refactor initiative looks ok to me. But I was thinking about invalidating the cache when an event's status changes, and I think the following use cases are not covered at the moment: 
1) The `Scenario: User can complete the enrollment and the active events` feature

Image

2) Completing events from the working lists

Image



Should we address these in preparation for [DHIS2-21919](https://dhis2.atlassian.net/browse/DHIS2-21919), or leave them for later? Thanks!

henrikmv: Nice catch! I’ll do all the preparation for [DHIS2-21919](https://dhis2.atlassian.net/browse/DHIS2-21919) now, which means including invalidating all the necessary caches.

Thanks!

simonadomnisoru (approved):
Good job! 👏

PR #4704, opened 2026-08-18
fix: [DHIS2-18620][DHIS2-21937] Hidden TEA is displayed in the TEI profile

henrikmv:
This PR fixes two bugs:

[DHIS2-18620](https://dhis2.atlassian.net/browse/DHIS2-18620) — Run program rules at the enrollment scope (new `useEnrollmentScopeRuleEffects`) and have WidgetProfile subscribe to `rulesEffectsHiddenFields` to filter its display list.

[DHIS2-21937](https://dhis2.atlassian.net/browse/DHIS2-21937) — Derive `ownerOrgUnitId` from the enrollment and resolve the org unit via `useCoreOrgUnit(ownerOrgUnitId)` instead of the `URL`, so the modal loads when no org unit is selected.

---
_Note: DHIS2-18620 breaks WidgetProfile's self-contained pattern (now depends on `rulesEffectsHiddenFields` populated externally). Tracked for a proper revamp in [DHIS2-21954](https://dhis2.atlassian.net/browse/DHIS2-21954)._

[DHIS2-18620]: https://dhis2.atlassian.net/browse/DHIS2-18620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-21937]: https://dhis2.atlassian.net/browse/DHIS2-21937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-21954]: https://dhis2.atlassian.net/browse/DHIS2-21954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/Pages/Enrollment/EnrollmentPageDefault/EnrollmentPageDefault.container.tsx
```diff
@@ -64,13 +63,15 @@ export const EnrollmentPageDefault = () => {
...
 
-    const program = useTrackerProgram(programId);
     const {
         error: enrollmentsError,
         enrollment,
         attributeValues,
         readOnly: trackedEntityInactive,
     } = useCommonEnrollmentDomainData(teiId, enrollmentId, programId);
+    const program = useTrackerProgram(programId);
+    const programOwnerId = useSelector(({ enrollmentPage }: any) => enrollmentPage.programOwners?.[programId]);
```
simonadomnisoru: Since `programOwnerId` is already exposed by `useCommonEnrollmentDomainData` on line 71, we can use that directly instead of reading it from Redux

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.tsx
```diff
@@ -111,6 +115,15 @@ export const EnrollmentAddEventPageDefault = ({
 
     const dataEntryHasChanges = useSelector((state: ReduxState) => getDataEntryHasChanges(state, widgetReducerName));
     const { program } = useProgramInfo(programId);
+    const trackerProgram = useTrackerProgram(programId);
+    const { orgUnit: programOwnerOrgUnit } = useCoreOrgUnit(programOwnerId);
+
+    const ruleEffects = useRuleEffects({
```
simonadomnisoru: TEst: Are the rules effect run twice??

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.tsx
```diff
@@ -156,7 +162,13 @@ const EnrollmentEditEventPageWithContextPlain = ({
     const hideWidgets = useHideWidgetByRuleLocations(
         program.programRules.concat(programStage?.programRules as ProgramRule[]),
     );
-
+    const { orgUnit: programOwnerOrgUnit } = useCoreOrgUnit(programOwnerId);
+    const ruleEffects = useRuleEffects({
```
simonadomnisoru, line 166: Same issue here for EnrollmentEditEvent. Since the RuleEngine was already run once when the form was initiated, we shouldn't need to re-run it, and we can just read the computed effects from Redux.

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/EnrollmentPageLayout/LayoutComponentConfig/LayoutComponentConfig.ts
```diff
@@ -143,17 +143,19 @@ export const ProfileWidget: WidgetConfig = {
...
+        programOwnerId,
         onUpdateTeiAttributeValues,
         onDeleteTrackedEntitySuccess,
         onStatusToggleSuccess,
+        ruleEffects,
     }: any): WidgetProfileProps => ({
         teiId,
         programId: program.id,
-        orgUnitId,
+        orgUnitId: programOwnerId,
```
simonadomnisoru: Should we also rename the WidgetProfile prop to `programOwnerId`?

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.tsx
```diff
@@ -111,6 +115,15 @@ export const EnrollmentAddEventPageDefault = ({
...
+        orgUnit: programOwnerOrgUnit,
+        program: trackerProgram,
+        apiEnrollment: enrollment ?? undefined,
+        apiAttributeValues: attributeValues ?? undefined,
+    });
```
simonadomnisoru, line 126: In EnrollmentAddEvent, the RuleEngine already runs once when the form is initiated and `getApplicableRuleEffectsForTrackerProgram` is called. We can just read the computed effects directly from Redux (they are in the `rulesEffects*` keys) here instead of rerunning the Engine and recomputing the effects.

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.tsx
henrikmv: I think we can not read from Redux here.

When the edit-event initializes, it runs the rule engine and dispatch effects to Redux. But it passes `stage.stageForm` as the postProcess foundation, and `postProcessRulesEffects` filters out any effect whose target isn't in that form. Since `stage.stageForm` only has data elements (no TEAs), all the TEA `HIDE_FIELD` effects get stripped before they hit `rulesEffectsHiddenFields`. So reading from that slice would give us nothing to filter with, and the bug would come back.

There's also a rule run inside `WidgetProfile.DataEntry` that does preserve TEA effects, but that only mounts when the user opens the edit-profile modal, so it doesn't help with the display list we're fixing.

Am I missing a slice somewhere? Otherwise I think useRuleEffects needs to stay.

WDYT?

src/core_modules/capture-core/components/Pages/EnrollmentAddEvent/EnrollmentAddEventPageDefault/EnrollmentAddEventPageDefault.container.tsx
henrikmv: Same reason as on the edit-event side. Happy to explore a shared enrollment-scope execution as a follow-up.

src/core_modules/capture-core/components/WidgetStagesAndEvents/Stages/Stage/Stage.component.tsx
```diff
@@ -22,16 +25,15 @@ const styles = {
...
 export const StagePlain = ({
-    stage, events, classes, onCreateNew, ruleEffects, ...passOnProps
+    stage, events, classes, onCreateNew, ...passOnProps
 }: Props & WithStyles<typeof styles>) => {
     const [open, setOpenStatus] = useState(true);
     const { id, name, icon, description, dataElements, hideDueDate, repeatable, enableUserAssignment } = stage;
-    const preventAddingNewEvents = rulesEffectHideProgramStage(ruleEffects, id);
+    const { enrollmentId } = useLocationQuery();
+    const hiddenProgramStageIds = useSelector(({ rulesEffectsHiddenProgramStage }: any) =>
+        (enrollmentId ? rulesEffectsHiddenProgramStage?.[getEnrollmentScopeFormId(enrollmentId)] : undefined));
```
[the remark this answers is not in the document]
henrikmv: Nice catch!

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useRuleEffects.ts
```diff
@@ -1,78 +1,12 @@
...
-    return {
-        enrolledAt: convertDate(enrolledAt),
-        occurredAt: occurredAt ? convertDate(occurredAt) : undefined,
-        enrollmentId,
-        enrollmentStatus: status,
-        programName: getTrackerProgramThrowIfNotFound(program).name,
-    };
-}, [enrollment]);
-
 export const useRuleEffects = ({ orgUnit, program, apiEnrollment, apiAttributeValues }: UseRuleEffectsInput) => {
```
simonadomnisoru: Since `useRuleEffects` and `UseRuleEffectsInput` have been replaced by useEnrollmentScopeRuleEffects, they are no longer in use. Can we delete them?

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useEnrollmentScopeRuleEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+            const effects = getApplicableRuleEffectsForTrackerProgram({
+                program,
+                orgUnit,
+                otherEvents,
+                attributeValues,
+                enrollmentData,
+            });
+            dispatch(updateRulesEffects(effects, formId));
+        }
+    }, [dispatch, formId, attributeValues, enrollmentData, orgUnit, otherEvents, program]);
```
simonadomnisoru: I noticed that on page refreshes or direct URL navigation, the effect gets triggered a few extra times (`enrollmentEventEdit` fires 3x and `enrollmentEventNew` fires 2x), which runs the RuleEngine more often than needed.
The hook on the `enrollment` page is working really smoothly and only fires once for both refresh and direct URL navigation. Could we align these other pages to prevent unnecessary executions? 
Thanks!

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useEnrollmentScopeRuleEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+    orgUnit,
+    program,
+    apiEnrollment,
+    apiAttributeValues,
+    force = false,
+}: Input) => {
+    const dispatch = useDispatch();
+    const formId = enrollmentId && getEnrollmentScopeFormId(enrollmentId);
+    const hasWarmStoreOnMount = useSelector(({ rulesEffectsHiddenFields }: any) =>
+        (formId ? rulesEffectsHiddenFields?.[formId] !== undefined : false));
```
simonadomnisoru: Does it make sense to check for other effects as well, so in case a Program has no hiddenFields effects, the page could still benefit from the `skipOnce` logic?

henrikmv: I think checking `rulesEffectsHiddenFields` alone should be enough.

The reducer updates that slice on every `UPDATE_RULES_EFFECTS`, either with the computed hidden fields or with `null` if the program doesn’t have any. So even if a program only has warnings, we’ll still end up with `{ [formId]: null }`, and since `null !== undefined`, the skip will still kick in.

The other slices are updated by the same action, so they should all be initialised at the same time. In that sense, checking any one of them should work.

WDYT?

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useEnrollmentScopeRuleEffects.ts
henrikmv: Thanks for flagging this. I traced the extra fires back to `useCommonEnrollmentDomainData`.

Several components on the edit/new pages mount that hook independently, and each instance was dispatching `setCommonEnrollmentSiteData` once the data loaded. That then caused `useEnrollmentScopeRuleEffects` to run multiple times.

Adding the `storedEnrollmentId !== enrollmentId` check makes sure we only dispatch once. The first instance updates Redux, and the others see that the data is already there and skip the dispatch.

Results:

* `EnrollmentAddEventPageDefault`: 2 → 1
* `EnrollmentEditEventPage`: 3 → 1
* `EnrollmentPageDefault`: 1 → 1 (unchanged)

simonadomnisoru (approved):
Nice! 🥳

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useCommonEnrollmentDomainData/useCommonEnrollmentDomainData.ts
```diff
@@ -4,6 +4,13 @@ import { setCommonEnrollmentSiteData } from '../enrollment.actions';
 import type { Output } from './useCommonEnrollmentDomainData.types';
 import { useApiDataQuery } from '../../../../../utils/reactQueryHelpers';
 
+// Program rules can reference program-scoped TEAs (stored on the enrollment) as well as
+// TET-scoped TEAs. Merge both so the value map fed to the rules engine is complete.
+const mergeAttributeValues = (enrollmentAttrs: Array<any> = [], teAttrs: Array<any> = []) => [
```
simonadomnisoru: NIT: Rename `teAttrs` to `trackedEntityAttributes` as it is the common name used in other files

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useCommonEnrollmentDomainData/useCommonEnrollmentDomainData.ts
```diff
@@ -21,7 +29,7 @@ export const useCommonEnrollmentDomainData = (teiId: string, enrollmentId: strin
             id: teiId,
             params: {
                 program: programId,
-                fields: ['enrollments[*,!attributes],attributes,inactive'],
+                fields: ['enrollments[*],attributes,inactive,programOwners[program,orgUnit]'],
```
simonadomnisoru: I'm worried that removing `!attributes` might reintroduce the bug reported in [DHIS2-17161](https://dhis2.atlassian.net/browse/DHIS2-17161), which was fixed in [#3601](https://github.com/dhis2/capture-app/pull/3601). But the `useCommonEnrollmentDomainData` hook has changed quite a bit since that fix so it might be fine now, could you please double-check? Thank you!

henrikmv: Thanks for the heads up. With that information in mind, I have split the fixes into two PRs:
- This PR — applies program rule effects to the WidgetProfile display list and fixes the infinite loading when no org. unit is selected.
- [Follow-up PR](https://dhis2.atlassian.net/browse/DHIS2-22080) — makes program rules in the edit-profile modal resolve against the option's display name by default, matching every other scope.

simonadomnisoru (approved):

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useEnrollmentScopeRuleEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+                return;
+            }
+            const effects = getApplicableRuleEffectsForTrackerProgram({
+                program,
+                orgUnit,
+                otherEvents,
+                attributeValues,
+                enrollmentData,
+            });
+            dispatch(updateRulesEffects(effects, formId));
```
JoakimSM: The redux state that this action populates was meant for the forms. I suggest we don't mix our data here with the form data, but instead add the needed ruleEffects to the enrollmentDomain Redux state.

src/core_modules/capture-core/components/WidgetProfile/WidgetProfile.component.tsx
```diff
@@ -86,6 +87,9 @@ const WidgetProfilePlain = ({
         storedGeometry: trackedEntityInstance?.geometry,
         hasError: trackedEntityInstance?.hasError,
     }));
+    const enrollmentId = useSelector(({ enrollmentDomain }: any) => enrollmentDomain?.enrollmentId);
+    const hiddenAttributeIds = useSelector(({ rulesEffectsHiddenFields }: any) =>
+        (enrollmentId ? rulesEffectsHiddenFields?.[getEnrollmentScopeFormId(enrollmentId)] : undefined));
```
JoakimSM: I might be missing something, but:
Since the rulesEffects run from an useEffect, the browser can paint before the rules are run, meaning that that hidden attributes can be visible on the screen. Even though you shouldn't rely on the rules to hide your sensitive data, I think we should avoid showing the fields before the rules have been executed. If we move the rulesEffects to the enrollmentDomain Redux state (see other comment), we can check whether the rulesEffects object exists + enrollmentId equals Redux state enrollmentId (if not, don't render the profile just yet)

src/core_modules/capture-core/components/Pages/common/EnrollmentOverviewDomain/useRuleEffects/useEnrollmentScopeRuleEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+    force?: boolean;
+};
+
+export const useEnrollmentScopeRuleEffects = ({
+    enrollmentId,
+    orgUnit,
+    program,
+    apiEnrollment,
+    apiAttributeValues,
+    force = false,
```
JoakimSM: I think the retriggering logic changes quite a bit when we move the rulesEffects in the Redux state (see other comment). Initial thinking would be that after the enrollmentDomain data is set (can probably check if Redux store enrollmentId equals input enrollmentId) we should run the rules if the rulesEffects data is missing in the Redux state.

The rules should probably also retrigger when we update dependencies, like attributeValues.

I think we should be able to do this separately from the useCommonEnrollmentDomainData (i.e., having a separate hook like we have now)

JoakimSM (commented):
Let's talk about this one @henrikmv. I would really like us to change this quite a bit.

src/core_modules/capture-core/components/Pages/EnrollmentEditEvent/EnrollmentEditEventPage.container.tsx
```diff
@@ -156,7 +162,14 @@ const EnrollmentEditEventPageWithContextPlain = ({
...
     );
-
+    const { orgUnit: ownerOrgUnit } = useCoreOrgUnit(ownerOrgUnitId);
+    useEnrollmentScopeRuleEffects({
+        enrollmentId,
+        orgUnit: ownerOrgUnit,
+        program,
+        apiEnrollment: enrollmentSite,
+        apiAttributeValues: attributeValues ?? undefined,
+    });
```
JoakimSM, line 172: Looking at the `useEnrollmentScopeRuleEffects` hook, wouldn't this cause the rules execution to always run when you navigate from the enrollment dashboard to the edit event page for example? Or am I missing something? In `useEnrollmentScopeRuleEffects` I think we need to look at whether the rulesEffects in the Redux state is populated and, also, we might need to split into two useEffects to get the behavior we want.

henrikmv: Tanks! Fixed in `useEnrollmentScopeRuleEffects`: at mount, we check if Redux already has `ruleEffects` for this `enrollmentId`, and if so a `useRef` flag makes the first `useEffect` fire skip. Any later change to the effect's deps still triggers a normal re-run. Okay?

JoakimSM: Great, I think this works 👍

JoakimSM (approved):
🎉 Good work!

linadhis2:
LGTM from QA perspective

PR #4707, opened 2026-08-20
fix: [DHIS2-21727] program rules on enrollment+event registration page

superskip:
Introduces a new argument `isEnrollmentFormWithEvent` in the rule engine, which should be set to `true` when the user is in a combined enrollment+event registration form. The enrollment registration page for Child Programme is an example of such a form. The rule engine will now use `evaluateAll` when evaluating the rules for this kind of form.

The rule engine runs once during the form initialization step, but in the case of a combined enrollment+event registration form the **event** data is missing. I've addressed this problem in the first commit. Ideally we should build the event object in the same way as we do for subsequent rule engine runs, and I looked into whether this would be possible, but it turns out subsequent runs gets data from the redux store that is not present at the time we do the initial run. So what I did instead was to just create a dummy event that looks exactly like the object would have, had it been generated by means of a common procedure.


src/core_modules/capture-core/rules/RuleEngine/helpers/evaluateRules.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (!isFirstStageEventForm) {
+        return ruleEngine.evaluateEvent(
+            currentEvent,
+            enrollment,
+            events,
+            executionContext,
+        );
+    }
+    const duplicateActionIds = new Set<string>;
```
simonadomnisoru: NIT: I know it works, but the missing parentheses on the constructor just feel strange to me. Could we use `new Set<string>();` instead?

src/core_modules/capture-core/components/DataEntries/Enrollment/actions/open.actionBatchs.ts
```diff
@@ -70,6 +70,13 @@ export const openDataEntryForNewEnrollmentBatchAsync = async ({
     const formId = getDataEntryKey(dataEntryId, itemId);
     const addFormDataActions = addFormData(`${dataEntryId}-${itemId}`, formValues);
     const firstStageDataEntryPropsToInclude = firstStage && getDataEntryPropsToInclude(firstStage);
+
+    const fabricateFirstStageEvent = (stage: ProgramStage) => ({
+        programStageId: stage.id,
+        programStageName: stage.name,
+    });
```
simonadomnisoru, line 77: This is not ideal, but since the alternative would require a big refactor, I agree with using this `fabricateFirstStageEvent` workaround to serve as a dummy event during the form initialization run.

simonadomnisoru (approved):
Looks good! 👏

src/core_modules/capture-core/rules/RuleEngine/helpers/evaluateRules.ts
```diff
@@ -0,0 +1,52 @@
...
+    events,
+    executionContext,
+    isFirstStageEventForm,
+}: {
+    ruleEngine: RuleEngineJs
+    enrollment: RuleEnrollmentJs | null,
+    currentEvent: RuleEventJs | null | undefined,
+    events: Array<RuleEventJs>,
+    executionContext: RuleEngineContextJs,
+    isFirstStageEventForm: boolean | undefined,
```
JoakimSM, line 21: NIT: Can we improve the name? isEnrollmentFormWithFirstStageEvent/isEnrollmentFormWithEvent/isEnrollmentWithFirstStageEventForm/isEnrollmentWithEventForm or something like that.

linadhis2:
LGTM from QA checks

PR #4709, opened 2026-08-22
fix: [DHIS2-22007] comparison between the new and the cached rule messages

superskip:
This PR fixes the implementation of a function called `getRuleMessages`.
The purpose of this function is to decide if two objects `ruleMessage` and `prevRuleMessage` are identical.

There was a case that was handled incorrectly by the previous implementation: the case where the keys of `ruleMessage` is a **subset** of the keys of `prevRuleMessage`.

The new implementation is based on the assumption that the keys of the two objects must always be a subset of the **values** of the `messageStateKeys`-object (`'error'`, `'warning'`, `'errorOnComplete'`, and `'warningOnComplete'`).


henrikmv (approved):
Nice fix, @superskip!

simonadomnisoru (approved):

linadhis2:
LGTM from QA check

PR #4706, opened 2026-08-20
test: [DHIS2-22025] Cypress tests for working list view sharing on update

karolinelien:
[DHIS2-22025](https://dhis2.atlassian.net/browse/DHIS2-22025)

Cypress test for [DHIS2-21871](https://dhis2.atlassian.net/browse/DHIS2-21871) and [DHIS2-13020](https://dhis2.atlassian.net/browse/DHIS2-13020)

Replaces a broken Cypress test for working list view sharing with a new, comprehensive test suite under `WorkingListsSharing/`.

**Why the old test was broken:** it asserted sharing via the Share dialog UI, which renders stale client state. After a `PUT` that silently dropped `userAccesses`, the dialog still showed the old access — so the test passed on broken code.

**What changes:**

- The old `"Ensure sharing settings are preserved"` scenario and its step definitions are deleted from `EventWorkingListsUser`.
- A new `WorkingListsSharing` suite adds 6 scenarios covering event, tracker, and program-stage working lists, in both owner and non-owner roles.
- Assertions now read the **server** (`GET /api/<resource>/<id>?fields=sharing`) instead of the dialog, checking both `userAccesses` and `publicAccess`.
- Non-owner scenarios are seeded via API and use `@user:trackerAutoTestRestricted`. An `After` hook + pre-scenario sweep ensures cleanup even on failure.

[DHIS2-21871]: https://dhis2.atlassian.net/browse/DHIS2-21871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-22025]: https://dhis2.atlassian.net/browse/DHIS2-22025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-13020]: https://dhis2.atlassian.net/browse/DHIS2-13020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4714, opened 2026-08-31
chore: [DHIS2-18447] make TEI bulk action tests data independent

karolinelien:
## Why

Two of the five scenarios re-enabled in https://github.com/dhis2/capture-app/pull/4681 are failing on master and on every open PR, e.g. https://github.com/dhis2/capture-app/pull/4706:

- `the user should be able to bulk delete enrollments` — `expected [ Array(3) ] to deep include { enrollment: 'PvJFfKjNWbq' }`
- `the user should be able to bulk delete only active enrollments` — `cy.click() failed because this element is disabled`

Nothing in the app changed. The scenarios select **"the first 3 rows"** of the Malaria case diagnosis working list and then assert **fixed enrollment uids** in the request payload. The list sorts on `createdAt` descending (`getOrderQueryArgs.ts`), so any record created in that program at Ngelehun pushes the expected records out of the selection.

That is what happened: 44 tracked entities leaked into `qDkgAbB5Jlk` at Ngelehun during CI runs on 2026-08-31 (37 baseline demo records, so more junk than real data). `PvJFfKjNWbq` is still there and still COMPLETED — it is just no longer in the first three rows, and since every leaked enrollment is ACTIVE, the delete modal's "completed" checkbox has a count of 0 and renders disabled.

The leak comes from `NewPage.feature`. Its malaria registration scenario creates a **real** tracked entity, then deletes it again through the UI as the last steps of the scenario. When any step fails the record is left behind — and with `retries.runMode: 3` that is four records per run, one per attempt. The leaked records arrived in 11 bursts of exactly 4, ~32s apart, all named `Ana-<epoch>` / `Maria-<epoch>`, which is the generator in `NewPage.js`.

## What

**Stop the leak** (`NewPage`)
- Move the cleanup out of the scenario body into `After({ tags: '@with-malaria-entity-cleanup' })`, deleting through the API. An `After` hook runs even when the scenario fails, so a broken step can no longer leak records.
- Mirrors the existing hook in `BreakingTheGlass.js` (attribute filter → `importStrategy=DELETE`); the entity is found by the unique first name the fill step generates, so parallel shards cannot delete each other's records.
- Removes the now orphaned `you delete the recently added malaria entity` step.

**Make the assertions data independent** (`TrackerBulkActions`)
- Select rows by tracked entity id rather than by position. The table already exposes it — `OnlineList.component.tsx` renders each row with `dataTest={row[rowIdKey]}` and `rowIdKey="id"`, which for tracker working lists is the tracked entity uid. New step: `you select the rows for tracked entities <ids>`.
- Applied to all five scenarios that assert on specific records, so they act on the records they assert on:
  - Malaria case diagnosis: `QGyxOe1zewj` (COMPLETED `PvJFfKjNWbq`), `YZ64FEtYScT` (ACTIVE `Rkx1QOZeBra`), `IPJNnWA0jp6` (ACTIVE `hDVHG1OavhE`)
  - Malaria focus investigation: `s4NfKOuayqG` (COMPLETED), `QBYMQm9GU4T`, `dNpxRu1mWG5`, `neR4cmMY22o` (ACTIVE)
- The Malaria case diagnosis scenarios raise the page size to 100 first, because the target records are no longer on page 1 while the leaked records are still there. `the working list is displayed with 25 rows per page` is parameterised rather than copied.
- `you select row number 1` is replaced by the uid based step (the scenario needs an *active* enrollment, which row 1 only was by coincidence).

## Notes

- The failing CI shard here is the verification: these scenarios should now pass **against the polluted instance**, before any data cleanup happens.
- Registration Date is hidden as a column for this program (`visible: !hasDisplayInReportsAttributes`), so sorting ascending is not available as an alternative to raising the page size.
- Page size 100 covers the 81 records currently in that program. If pollution ever exceeds that, the fix is seeded records rather than a bigger page — worth doing if this recurs, but out of scope here.
- The 44 leaked records still need deleting on the instance; doing that separately.
- Not fixed here: the terminology branches that break the NewPage steps in the first place — reported to the branch owner.

AI Assisted.

### DIFF
--- a/cypress/e2e/NewPage/NewPage.js
+++ b/cypress/e2e/NewPage/NewPage.js
@@ -5,7 +5,6 @@ import { getCurrentYear } from '../../support/date';
 const MALARIA_FIRST_NAME_ATTRIBUTE = 'TfdH5KvFmMy';
 const MALARIA_PROGRAM = 'qDkgAbB5Jlk';
 
-// Set when the registration form is filled in, so the cleanup hook can find the entity again.
 let malariaEntityFirstName;
 
 const clearMalariaEntity = () => {
@@ -30,8 +29,6 @@ const clearMalariaEntity = () => {
         );
 };
 
-// The entity is created for real, so it has to be removed even when the scenario fails halfway
-// through. Deleting it from within the scenario leaks a record on every failed attempt.
 After({ tags: '@with-malaria-entity-cleanup' }, () => {
     clearMalariaEntity();
     malariaEntityFirstName = undefined;
--- a/cypress/e2e/NewPage/NewPage.js
+++ b/cypress/e2e/NewPage/NewPage.js
@@ -572,7 +572,7 @@ Given('you are in the Malaria case diagnosis, treatment and investigation progra
 });
 
 And('you fill the Malaria case diagnosis registration form with values', () => {
-    malariaEntityFirstName = `Ana-${Math.round((new Date()).getTime() / 1000)}`;
+    malariaEntityFirstName = `Ana-${Math.round(Date.now() / 1000)}`;
 
     cy.get('input[type="text"]')
         .eq(3)
@@ -580,7 +580,7 @@ And('you fill the Malaria case diagnosis registration form with values', () => {
         .blur();
     cy.get('input[type="text"]')
         .eq(4)
-        .type(`Maria-${Math.round((new Date()).getTime() / 1000)}`)
+        .type(`Maria-${Math.round(Date.now() / 1000)}`)
         .blur();
     cy.get('input[type="text"]')
         .eq(5)
--- a/cypress/e2e/NewPage/NewPage.js
+++ b/cypress/e2e/NewPage/NewPage.js
@@ -16,17 +16,21 @@ const clearMalariaEntity = () => {
         .buildApiUrl(
             'tracker',
             `trackedEntities?program=${MALARIA_PROGRAM}&orgUnitMode=ACCESSIBLE` +
-                `&filter=${MALARIA_FIRST_NAME_ATTRIBUTE}:like:${malariaEntityFirstName}` +
+                `&filter=${MALARIA_FIRST_NAME_ATTRIBUTE}:eq:${malariaEntityFirstName}` +
                 '&fields=trackedEntity&page=1&pageSize=5',
         )
         .then(url => cy.request(url))
-        .then(({ body }) =>
-            (body.trackedEntities || []).forEach(({ trackedEntity }) =>
-                cy
-                    .buildApiUrl('tracker?async=false&importStrategy=DELETE')
-                    .then(deleteUrl => cy.request('POST', deleteUrl, { trackedEntities: [{ trackedEntity }] })),
-            ),
-        );
+        .then(({ body }) => {
+            const trackedEntities = (body.trackedEntities || []).map(({ trackedEntity }) => ({ trackedEntity }));
+
+            if (!trackedEntities.length) {
+                return undefined;
+            }
+
+            return cy
+                .buildApiUrl('tracker?async=false&importStrategy=DELETE')
+                .then(deleteUrl => cy.request('POST', deleteUrl, { trackedEntities }));
+        });
 };
 
 After({ tags: '@with-malaria-entity-cleanup' }, () => {
@@ -569,7 +573,7 @@ Given('you are in the Malaria case diagnosis, treatment and investigation progra
 });
 
 And('you fill the Malaria case diagnosis registration form with values', () => {
-    malariaEntityFirstName = `Ana-${Math.round(Date.now() / 1000)}`;
+    malariaEntityFirstName = `Ana-${Date.now()}`;
 
     cy.get('input[type="text"]')
         .eq(3)
@@ -577,7 +581,7 @@ And('you fill the Malaria case diagnosis registration form with values', () => {
         .blur();
     cy.get('input[type="text"]')
         .eq(4)
-        .type(`Maria-${Math.round(Date.now() / 1000)}`)
+        .type(`Maria-${Date.now()}`)
         .blur();
     cy.get('input[type="text"]')
         .eq(5)
--- a/cypress/e2e/NewPage/NewPage.js
+++ b/cypress/e2e/NewPage/NewPage.js
@@ -21,7 +21,8 @@ const clearMalariaEntity = () => {
         )
         .then(url => cy.request(url))
         .then(({ body }) => {
-            const trackedEntities = (body.trackedEntities || []).map(({ trackedEntity }) => ({ trackedEntity }));
+            const apiTrackedEntities = body.trackedEntities || body.instances || [];
+            const trackedEntities = apiTrackedEntities.map(({ trackedEntity }) => ({ trackedEntity }));
 
             if (!trackedEntities.length) {
                 return undefined;
--- a/cypress/e2e/NewPage/NewPage.js
+++ b/cypress/e2e/NewPage/NewPage.js
@@ -1,6 +1,41 @@
-import { defineStep as And, Given, Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { After, defineStep as And, Given, Then, When } from '@badeball/cypress-cucumber-preprocessor';
 import moment from 'moment';
 import { getCurrentYear } from '../../support/date';
+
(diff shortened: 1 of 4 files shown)


cypress/e2e/NewPage/NewPage.js
```diff
@@ -1,7 +1,39 @@
...
+                '&fields=trackedEntity&page=1&pageSize=5',
+        )
+        .then(url => cy.request(url))
+        .then(({ body }) =>
+            (body.trackedEntities || []).forEach(({ trackedEntity }) =>
+                cy
+                    .buildApiUrl('tracker?async=false&importStrategy=DELETE')
+                    .then(deleteUrl => cy.request('POST', deleteUrl, { trackedEntities: [{ trackedEntity }] })),
+            ),
+        );
```
henrikmv: **One `DELETE` per entity instead of a single batched request**

The tracker `DELETE` endpoint accepts a trackedEntities array — no need to loop. Collapsing the forEach into one request avoids N round-trips and removes the detached-chain pattern (each iteration starts a fresh `cy.buildApiUrl(...).then(...)` inside a synchronous callback).

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.feature
```diff
@@ -44,7 +44,7 @@ Feature: User facing tests for bulk actions on Tracked Entity working lists
 
   Scenario: The user should see an error message when trying to bulk complete enrollments with errors
     Given you open the main page with Ngelehun and Malaria focus investigation context
-    And you select the first 3 rows
+    And you select the rows for tracked entities s4NfKOuayqG, QBYMQm9GU4T, dNpxRu1mWG5
```
henrikmv, line 48: **Scenario missing the 100 rows per page prep**

Both focus investigation scenarios select tracked entities by ID without first expanding the page size. It works today because those IDs happen to sit on the first page at the default 25 rows, but this is the kind of implicit data dependency we want to eliminate.

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.feature
```diff
@@ -55,7 +55,7 @@ Feature: User facing tests for bulk actions on Tracked Entity working lists
 
   Scenario: the user should be able to bulk complete enrollments and events
     Given you open the main page with Ngelehun and Malaria focus investigation context
-    And you select the first 4 rows
+    And you select the rows for tracked entities s4NfKOuayqG, QBYMQm9GU4T, dNpxRu1mWG5, Er7oTDZsq6N
```
henrikmv, line 60: **Scenario missing the 100 rows per page prep**

Both focus investigation scenarios select tracked entities by ID without first expanding the page size. It works today because those IDs happen to sit on the first page at the default 25 rows, but this is the kind of implicit data dependency we want to eliminate.

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.js
```diff
@@ -127,20 +127,21 @@ Then('the bulk complete enrollments modal should close', () => {
...
+    trackedEntityIds.split(', ').forEach(trackedEntityId =>
+        cy.get('[data-test="online-list-table"]')
+            .find(`[data-test="dhis2-uicore-tablebody"] tr[data-test="${trackedEntityId}"]`)
+            .find('[data-test="select-row-checkbox"]')
+            .click(),
+    );
 });
 
-Then('the working list is displayed with 25 rows per page', () => {
+Then(/^the working list is displayed with (\d+) rows per page$/, (rowsPerPage) => {
```
henrikmv: **Step name and keyword read as an assertion but the step performs an action**

The step clicks the rows-per-page selector and picks a value — it changes state. Both the Then keyword and the phrasing ("the working list is displayed with N rows per page") read like an assertion. 

Cucumber-preprocessor treats Given/When/Then as aliases so there's no runtime effect, but this is the moment to lock in a name that matches behaviour, since this step is about to spread to more scenarios:


```
When(/^you set the working list to display (\d+) rows per page$/, (rowsPerPage) => {
    ...
});
```

Update the feature files to `And you set the working list to display 100 rows per page`.

cypress/e2e/NewPage/NewPage.js
```diff
@@ -1,7 +1,39 @@
...
+const clearMalariaEntity = () => {
+    if (!malariaEntityFirstName) {
+        return undefined;
+    }
+
+    return cy
+        .buildApiUrl(
+            'tracker',
+            `trackedEntities?program=${MALARIA_PROGRAM}&orgUnitMode=ACCESSIBLE` +
+                `&filter=${MALARIA_FIRST_NAME_ATTRIBUTE}:like:${malariaEntityFirstName}` +
```
henrikmv: **Cheap parallel-safety win: use ms-precision timestamp**

Not blocking (parallel collisions already exist elsewhere in the suite), but the fix is actually shorter code: replace `Math.round(Date.now() / 1000)` with `Date.now()` on both name lines. Then, also swap `:like:` for `:eq:` to also make the filter exact-match.

henrikmv (commented):
Good start @karolinelien!

cypress/e2e/NewPage/NewPage.js
karolinelien: Thanks, fixed ✅

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.feature
karolinelien: Thanks, fixed ✅

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.js
karolinelien: Thanks, fixed ✅

cypress/e2e/NewPage/NewPage.js
karolinelien: Thanks, fixed ✅

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.feature
karolinelien: Thanks, fixed ✅

cypress/e2e/WorkingLists/TrackerWorkingLists/TrackerBulkActions/TrackerBulkActions.js
```diff
@@ -127,20 +127,21 @@ Then('the bulk complete enrollments modal should close', () => {
...
-        .eq(rowNumber)
-        .find('[data-test="select-row-checkbox"]')
-        .click();
+When(/^you select the rows for tracked entities (.*)$/, (trackedEntityIds) => {
+    trackedEntityIds.split(', ').forEach(trackedEntityId =>
+        cy.get('[data-test="online-list-table"]')
+            .find(`[data-test="dhis2-uicore-tablebody"] tr[data-test="${trackedEntityId}"]`)
+            .find('[data-test="select-row-checkbox"]')
+            .click(),
+    );
```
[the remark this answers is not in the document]
karolinelien: Done 👍🏻

cypress/e2e/NewPage/NewPage.js
```diff
@@ -1,7 +1,43 @@
...
+    return cy
+        .buildApiUrl(
+            'tracker',
+            `trackedEntities?program=${MALARIA_PROGRAM}&orgUnitMode=ACCESSIBLE` +
+                `&filter=${MALARIA_FIRST_NAME_ATTRIBUTE}:eq:${malariaEntityFirstName}` +
+                '&fields=trackedEntity&page=1&pageSize=5',
+        )
+        .then(url => cy.request(url))
+        .then(({ body }) => {
+            const trackedEntities = (body.trackedEntities || []).map(({ trackedEntity }) => ({ trackedEntity }));
```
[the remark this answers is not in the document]
karolinelien: Done :)

henrikmv (approved):
Nice! LGTM

PR #4715, opened 2026-09-01
test: [DHIS2-22043] Cypress tests in dev container

JoakimSM:
Fixes issues with Cypress tests not running from a dev container.

- To communicate with DHIS2 instances running on the host machine's localhost, a proxy/relay tool (socat) is used. The Cypress configuration remains unchanged, e.g. `CYPRESS_dhis2BaseUrl=http://localhost:8080` (see the [wiki](https://github.com/dhis2/capture-app/wiki/Cypress#run-cypress-tests-locally)). This removes the need to configure HTTPS for instances running on localhost.
- For instances running on other hosts, the proxy is not used.


simonadomnisoru (approved):

henrikmv (approved):
You never gave up, and I can confirm it works for me now!

PR #4718, opened 2026-09-03
chore: [DHIS2-22074] de-flake the RelatedStages, assignee and sharing dialog scenarios

karolinelien:
https://dhis2.atlassian.net/browse/DHIS2-22074

Six scenarios that depend on state no step establishes, plus one race.

- **RelatedStages** (×2) assert the schedule and enter-details actions are disabled, which only holds when the Baby Postnatal stage already has an event — and nothing creates one, so they rely on a leftover. New step seeds it through the API.
- **Assignee** (×2) assign a user and remove them as their last step. The `Assign` button renders only when nobody is assigned, so one interrupted run leaves the shared instance stuck for every branch. New step clears the assignment.
- **Sharing dialog** (×2) type into the search box while it is still disabled — `cy.type()` does not wait for that to clear. Gated with `should('be.enabled')` and scoped to the dialog.

Also: the `Enroll trackedEntity` scenarios clean up through an API `After` hook instead of the UI, so it runs when a scenario fails; and the two deletion steps always reload.

Verified against play stable-2-43-1, each run starting from the broken state — including the control: with the assignee guard removed, the CI failure reproduces exactly.

AI Assisted.


cypress/support/commands.js
```diff
@@ -4,6 +4,14 @@ Cypress.Commands.add('buildApiUrl', (...urlParts) =>
         .join('/'),
 );
 
+Cypress.Commands.add('importTracker', (importStrategy, payload) =>
```
henrikmv: Not sure about introducing `cy.importTracker` here. We already have ~20 inline tracker-import call sites across the specs, and none use a helper today.

Two concerns:
* Using the helper only for these 6 new sites leaves us half-migrated and creates two patterns.
* `expect(body.status).to.eq('OK')` adds a new invariant that other tracker imports don’t assert, which feels separate from this de-flake PR.

I recommend inlining these 6 to match the existing style, and instead migrate all ~25 call sites in a broader readfactor later.

cypress/e2e/WidgetsForEnrollmentPages/WidgetAssignee/index.js
```diff
@@ -1,4 +1,20 @@
...
+const EVENT_FIELDS = 'event,program,programStage,enrollment,orgUnit,occurredAt,scheduledAt,status,dataValues';
+
+Given(/^you make sure the event (.+) has no assigned user$/, (eventId) => {
+    cy.buildApiUrl('tracker', `events/${eventId}?fields=${EVENT_FIELDS},assignedUser`)
+        .then(url => cy.request(url))
+        .then(({ body }) => {
+            const { assignedUser, ...event } = body;
+
+            if (!assignedUser) {
+                return undefined;
```
henrikmv, line 8: Explicit `assignedUser: null` over field omission — makes the "clear" intent legible at the callsite.

cypress/e2e/WidgetsForEnrollmentPages/WidgetAssignee/index.js
```diff
@@ -1,4 +1,20 @@
-import { When, Then } from '@badeball/cypress-cucumber-preprocessor';
+import { Given, When, Then } from '@badeball/cypress-cucumber-preprocessor';
+
+const EVENT_FIELDS = 'event,program,programStage,enrollment,orgUnit,occurredAt,scheduledAt,status,dataValues';
```
henrikmv: Drop the `EVENT_FIELDS` allowlist and fetch the whole event body — matches the existing GET-then-UPDATE pattern (e.g. EnrollmentEditEventPageForm.js) and removes the drift risk of missing a field if the event shape gains a required one later.

cypress/e2e/RelatedStages/RelatedStages.js
```diff
@@ -1,26 +1,90 @@
...
+    return cy
+        .buildApiUrl(
+            'tracker',
+            `trackedEntities?program=${CHILD_PROGRAMME}&orgUnitMode=ACCESSIBLE` +
+                `&filter=${FIRST_NAME_ATTRIBUTE}:eq:${enrolledEntityFirstName}` +
+                '&fields=trackedEntity&page=1&pageSize=5',
+        )
+        .then(url => cy.request(url))
+        .then(({ body }) => {
+            const apiTrackedEntities = body.trackedEntities || body.instances || [];
```
henrikmv: Fallback for `body.instances` is for pre-v40 tracker responses.

cypress/support/commands.js
karolinelien: Done!

cypress/e2e/WidgetsForEnrollmentPages/WidgetAssignee/index.js
karolinelien: Fixed!

cypress/e2e/WidgetsForEnrollmentPages/WidgetAssignee/index.js
karolinelien: Done!

cypress/e2e/RelatedStages/RelatedStages.js
karolinelien: Fixed!

henrikmv (approved):
Thanks!

PR #4720, opened 2026-09-04
docs: document uncomplete authority, working list filter minimum characters and calendars

karolinelien:
## What this does

Fills three gaps found while auditing the Capture app's user-facing surface against the user docs.

**1. Reopening a completed event requires an authority.** Added one line to the existing note about why an event cannot be edited. The tooltip in the app tells the user *what* is wrong but not which authority to ask for. `canUncompleteEvent` does not feed the read-only calculation (`useEventEditPermissions.ts`), so the form stays editable and only the **Complete event** checkbox locks, and the wording reflects that. The authority display name **Uncomplete events** was read from `/api/authorities.json` on a 2.43.1 instance, not guessed.

**2. Search performance settings on tracked entity attributes.** Adds `minCharactersToSearch` next to the existing note about `preferredSearchOperator`, and rewrites both in the wording the UI actually uses.

- These four properties are **not** editable in the legacy Maintenance app, which the old paragraph claimed. That app only renders a field if it is listed for the model in `src/config/field-config/field-order.js`, and its `trackedEntityAttribute` list contains none of them. They are only editable in the **Metadata Management app**, under Tracked entity attributes → **Search performance** — a section that app hides below server version 2.43 (`hasAPISupportForFeature`, `{ minor: 43 }`), so the version requirement is stated too.
- Both paragraphs now use the labels and options the Metadata Management app shows — **Preferred search operator**, **Blocked search operators**, **Minimum characters required to search**, and *EQ | Equals* / *SW | Starts with* / *EW | Ends with* / *LIKE | Contains* — instead of raw API property names.
- Corrects the value types the operator settings apply to. `PERCENTAGE` was moved into `forcedSearchOperators` by 01cff4f28 (DHIS2-12881) and now ignores `preferredSearchOperator`, so the honoured types are text, long text, email and phone number. Unique attributes and attributes with an option set always resolve to `EQ`, which was not stated.
- On `minCharactersToSearch`, the runtime message is self-explanatory, so the paragraph focuses on the dead end: a filter can only hold as many characters as its value type allows (10 for a date, 5 for a time, 3 for a percentage — `withMinCharsToSearchValidation.tsx`), so a higher minimum can never be satisfied and the filter shows *Minimum characters to search is too high for this filter*.

**3. New "Dates and calendars" section.** Non-Gregorian calendar support had no user documentation at all. Covers where the calendar comes from, which surfaces it affects, and that values are stored as ISO 8601 so exports and the API show Gregorian equivalents. Includes a screenshot of the report date picker on a Nepali-calendar server, captured on a 2.43.1 instance.

AI Assisted.


docs/user/using-the-capture-app.md
```diff
@@ -680,6 +682,8 @@ The Capture app supports the working lists in tracker programs, and opening a tr
 
     ![](resources/images/search-by-attributes-on-scope-program-preferredSearchOperator.png)
 
+    Tracked entity attributes can also define **minCharactersToSearch** in Maintenance. The same minimum applies to the attribute's filter in working lists, which shows *Please enter at least N characters to filter* until enough characters are entered. Set this value with care for non-text attributes. A filter can only ever hold as many characters as its value type allows - 10 for a date (2026-09-04), 5 for a time (14:30) and 3 for a percentage (100) - so if the minimum is higher than that, the filter cannot be used at all and shows *Minimum characters to search is too high for this filter*.
```
henrikmv: `Maintenance` &rarr; `Metadata Management app`

henrikmv (approved):
Thank you, @karolinelien!

simonadomnisoru (approved):

docs/user/using-the-capture-app.md
karolinelien: Changing the search performance section a little bit to use the UI text etc. Have a look please @simonadomnisoru and @henrikmv

simonadomnisoru (approved):

PR #4724, opened 2026-09-05
chore(types): tighten organisation unit name retrieval

daryn-brown:
## Summary

- add strict types for organisation unit API responses, query results, cache entries, and subvalues
- make `useOrganisationUnit` generic so callers receive the fields they requested
- replace untyped query errors with `FetchError`
- align `getOrgUnitNames` types with its existing `{ id, name }` runtime response
- add focused coverage for organisation unit name and ancestor caching

## Why this is needed

The organisation unit name retrieval module is used across enrollment, event,
profile, tooltip, filtering, and relationship workflows. Its extensive use of
`any` allowed API response changes and invalid cache assumptions to bypass
TypeScript.

It also concealed a mismatch between the declared return type of
`getOrgUnitNames` and the value it actually returns. Strict response and cache
types now make these contracts explicit without changing the runtime data flow
or public behavior.

## How to test

1. Run:
   `yarn test src/core_modules/capture-core/metadataRetrieval/orgUnitName/__tests__/orgUnitName.test.js --runInBand --watchAll=false`
2. Run:
   `yarn tsc:check`
3. Verify organisation unit names and ancestor paths still render in enrollment
   and event views.

## Checklist

- [x] Tests run locally
- [x] Type checks pass
- [x] No breaking changes were introduced


henrikmv:
Hi @daryn-brown , thanks for the PR.

We’re going to close this one for now, but not because the changes are wrong. The `displayName`/`name` mismatch in `getOrgUnitNames` is a real inconsistency, and the generic on `useOrganisationUnit` is a reasonable improvement.

The main issue is that changes to this module have a wide impact and need to go through our Jira prioritisation and review process.

If this came from a concrete issue in capture-app, please open a ticket at https://jira.dhis2.org/ and we’ll take it from there.

Thank you!

PR #4727, opened 2026-09-07
feat: [DHIS2-21635] Display custom terminology labels across UI

henrikmv:
[DHIS2-21635](https://dhis2.atlassian.net/browse/DHIS2-21635)

Implements Custom Terminology in the Capture app by using these functions:

```ts
// Non-hook, from programCollection (epics, thunks, selectors)
getTermLabel(requests, { programId, stageId? })

// Non-hook, from a program object already in scope (self-contained widgets)
getTermLabelFromProgram(requests, { program })

// React hook (reads programId from Redux currentSelections if not passed)
useTermLabel(requests, { programId?, stageId? })

// HOC — injects resolved labels as props (class components, cross-cutting subtrees)
withCustomLabels(customLabels)(Component)
```

Requests are batched: `[LabelKeys.enrollmentSingular, LabelKeys.eventPlural] returns { enrollmentLabel, eventsLabel }`.

[DHIS2-21635]: https://dhis2.atlassian.net/browse/DHIS2-21635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4728, opened 2026-09-07
chore: [DHIS2-21969] Refine custom terminology label support

henrikmv:
[DHIS2-21969](https://dhis2.atlassian.net/browse/DHIS2-21969)

This PR replaces the multi-function custom terminology API (`getProgramLabel`, `getStageLabel`, `useProgramLabel`, `useStageLabel`, `useTrackedEntityTypeLabel`)

New functions:

```ts
// Non-hook, from programCollection (epics, thunks, selectors)
getTermLabel(requests, { programId, stageId? })

// Non-hook, from a program object already in scope (self-contained widgets)
getTermLabelFromProgram(requests, { program })

// React hook (reads programId from Redux currentSelections if not passed)
useTermLabel(requests, { programId?, stageId? })

// HOC — injects resolved labels as props (class components, cross-cutting subtrees)
withCustomLabels(customLabels)(Component)
```

Requests are batched: `[LabelKeys.enrollmentSingular, LabelKeys.eventPlural] returns { enrollmentLabel, eventsLabel }.
`

Additional changes:
- **i18n interpolator patch** — auto-capitalizes leading custom-term variables and disables HTML escaping for them
- **Feature-gated plural fields** — plural terminology fields are only requested from backend ≥ v43

[DHIS2-21969]: https://dhis2.atlassian.net/browse/DHIS2-21969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/metaData/helpers/customLabels.ts
```diff
@@ -0,0 +1,98 @@
...
+/** Use inside React components; `programId` falls back to `currentSelections.programId`. */
+export const useTermLabel = (
+    requests: ReadonlyArray<TermRequest>,
+    { programId, stageId }: OptionalProgramScope = {},
+): CustomLabels => {
+    const activeProgramId = useSelector(({ currentSelections }: any) =>
+        programId ?? currentSelections.programId);
+    return buildLabels(requests, (key, plural) =>
+        resolveFromCollection(activeProgramId, stageId, key, plural));
+};
```
simonadomnisoru: Let's wrapp thsiw ith a useMemo to avoid creating a new object reference every render

src/core_modules/capture-core/metaDataStoreLoaders/programs/quickStoreOperations/storePrograms.ts
```diff
@@ -163,16 +164,38 @@ const fieldsParam = [
...
     'categoryCombo[id,displayName,isDefault,categories[id,displayName]]',
-    `programStages[${programStageFields}]`,
     'programSections[id, displayDescription, displayFormName, sortOrder, trackedEntityAttributes]',
     `programTrackedEntityAttributes[${programTrackedEntityAttributeFields}]`,
-].join(',');
+];
+
+const pluralProgramFields = [
+    'displayEnrollmentsLabel',
+    'displayProgramStagesLabel',
```
simonadomnisoru: Is `displayProgramStagesLabel` more suitable to be in ` const pluralProgramStageFields`?

src/core_modules/capture-core/metaData/helpers/customLabels.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const buildLabels = (
+    requests: ReadonlyArray<TermRequest>,
+    resolve: (key: CustomLabelKey, plural: boolean) => string,
+): CustomLabels => {
+    const entries = requests.map((req) => {
+        const isString = typeof req === 'string';
+        const key = isString ? req : req.key;
+        const plural = !isString && (req.plural ?? false);
+        const outputKey = plural ? `${key}sLabel` : `${key}Label`;
```
simonadomnisoru: This conversion will produce errors for labels with irregular plurarls, like it will convert category to categorysLabel, but the correct form shpuld be categoriesLabel

src/core_modules/capture-core/metaData/helpers/customLabels.ts
simonadomnisoru, line 106: Feels like we should have sort of some memoization here. The old useProgramLabel / useStageLabel (now deleted) hooks used to wrap the return with a `useMemo`

src/i18n/setupFormatters.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Patches d2-i18n's interpolator for custom terminology.
+ *
+ *   1. If the template interpolates any variable listed in CUSTOM_TERM_VARS,
+ *      HTML escaping is disabled for that call.
```
simonadomnisoru: I am not sure patching i18n is the most sensible approche from a secirity perspective.
How about adding `{ interpolation: { escapeValue: false } }` as a per-call option at each i18n.t(...) call site that uses custom labels, rather than patching the global interpolator? I tried it with the examples in thsi PR at is smeaps to work ok

src/core_modules/capture-core/metaData/helpers/customLabels.ts
```diff
@@ -0,0 +1,98 @@
...
+const ALL_FIELD_NAMES = Object.values(LABELS as Record<string, LabelConfig>).flatMap(
+    ({ apiFieldSingular, apiFieldPlural }) => (apiFieldPlural ? [apiFieldSingular, apiFieldPlural] : [apiFieldSingular]),
+);
+
+const resolveDefault = (key: CustomLabelKey, plural: boolean): string => {
+    const label = getLabel(key);
+    return plural ? label.defaultPlural?.() ?? label.defaultSingular() : label.defaultSingular();
+};
+
+const resolveLabel = (
```
simonadomnisoru, line 26: I think `resolveLabel` (and maybe a few other functions here) might really benefit from caching. The idea is that once a custom label is resolved, we could store it so that on future calls with the same `programId`, `stageId`, `key`, and `plural`, we can just grab it from the cache instead of re-computing. We do something similar in `orgUnitName.ts` for computing the orgUnit name. Do you think this is useful and can help with the overall performance?

simonadomnisoru (commented):
Good job here 👏 ! I like the overall approach. I went a few times back and forth between patching i18n versus just capitalizing and escaping in place, but after reviewing the other PR (https://github.com/dhis2/capture-app/pull/4727), I think your patch approach is cleaner 👍 .
The only thing I think is worth having a second look at is improving the performance of the `customLabels.ts` functions (see the other, more detailed comments).

src/core_modules/capture-core/metaData/helpers/customLabels.ts
henrikmv: Looked into it –  I think `useMemo` isn't buying us anything here. The old `useProgramLabel`/`useStageLabel` returned a bare string, so memoisation was harmless bookkeeping. The new `useTermLabel` returns an object, but every call site immediately destructures it (`const { xLabel } = useTermLabel(...)`), so object identity is thrown away and only the string values flow downstream — strings compare by value, so a fresh object per render costs nothing.

Wrapping in `useMemo` would also need a stable dep for the `requests` array (callers pass inline literals), which means computing a signature on every render will be more work than just running `buildLabels`.

The real win is the resolve-level cache you mentioned in the other comment — see the new commit.

Please let me know if I have misunderstood something here.

simonadomnisoru (approved):
LGTM!

PR #4729, opened 2026-09-07
chore: [DHIS2-22063] Block pushes with hardcoded custom terminology in pot file

henrikmv:
[DHIS2-22063](https://dhis2.atlassian.net/browse/DHIS2-22063)

Adds a new lint-like check (`yarn i18n:verify`) that scans `i18n/en.pot` for custom-terminology words (e.g. "enrollment", "event", "note") used **outside `{{...}}` template placeholders**. These terms must be wrapped in placeholders so they can be overridden per program at runtime.

The check is enforced in two places:
- **Pre-push hook** — blocks local pushes
- **CI workflow** — new `i18n` job gates the publish step

The pipeline is:
```text
d2-app-scripts i18n extract
  → node scripts/verifyCustomTerminology.mjs
    → parse en.pot msgids
    → strip {{…}} placeholders
    → regex-match CUSTOM_TERMS
    → exit 1 if violations found
```

[DHIS2-22063]: https://dhis2.atlassian.net/browse/DHIS2-22063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/verifyCustomTerminology.mjs
```diff
@@ -0,0 +1,131 @@
...
+    { words: ['program stages'], suggestion: '{{programStagesLabel}}' },
+    { words: ['note'], suggestion: '{{noteLabel}}' },
+    { words: ['notes'], suggestion: '{{notesLabel}}' },
+    { words: ['relationship'], suggestion: '{{relationshipLabel}}' },
+    { words: ['relationships'], suggestion: '{{relationshipsLabel}}' },
+    { words: ['attribute'], suggestion: '{{attributeLabel}}' },
+    { words: ['attributes'], suggestion: '{{attributesLabel}}' },
+    { words: ['organisation unit', 'org unit', 'organization unit', 'registering unit'], suggestion: '{{orgUnitLabel}}' },
+    { words: ['follow-up', 'followup', 'follow up'], suggestion: '{{followUpLabel}}' },
+].sort((a, b) => Math.max(...b.words.map(w => w.length)) - Math.max(...a.words.map(w => w.length)));
```
simonadomnisoru: Could we drop `sort()`? From what I understand, its goal is to check longer custom terminology first, but I don't think removing it will affect the `findViolations` logic.

simonadomnisoru (commented):
Awesome! I just added one small comment

scripts/verifyCustomTerminology.mjs
```diff
@@ -0,0 +1,131 @@
...
+    // "event program" = DHIS2 programType, not user's event terminology
+    '{{programName}} is an event program and does not have {{enrollmentsLabel}}.',
+    // NoSelectionsInfoBox — only renders when no program is selected, so no program-specific label to use
+    'Choose a program and organisation unit to see existing data and create new records.',
+    // ProgramList — program picker shows generic label rather than any single program's custom label
+    'Some programs are being filtered by the chosen organisation unit',
+]);
+
+function extractStrings(potContents) {
+    const header = /^(msgid|msgstr(?:\[\d+\])?) "(.*)"$/;
```
simonadomnisoru, line 58: Should we check for `msgid_plural` too?  Like this `const header = /^(msgid_plural|msgid|msgstr(?:\[\d+\])?) "(.*)"$/;`

henrikmv: In the pot file, `msgid_plural` always equals `msgid`. `msgstr[1]` reflects `msgid_plural`, so it should already be covered.

simonadomnisoru (approved):

PR #4742, opened 2026-09-20
chore: [DHIS2-22132] Refactor programId handling in custom terminology

henrikmv:
[DHIS2-22132](https://dhis2.atlassian.net/browse/DHIS2-22132)

**Bugs addressed by QA**
  1. **Refresh / direct link on `/enrollment` shows default terms** in the top-bar org unit picker, "+ New <stage> event" button, and Notes widget — but navigating to the same record via the program overview shows the correct labels.
  2. **Cross-program label leak:** after viewing a program with custom terminology, records in a different program pick up the first program's labels. Same behaviour after deselecting a program — the org unit selector remains stuck on the previous program's custom term.

**Cause**

Both bugs share a root cause: `useTermLabel` reads `state.currentSelections.programId`, but on `/enrollment` that value was never being kept in sync — the route is intentionally excluded from `withAppUrlSync`, and the `resetProgramIdBase()` action designed to clear it on deselect has never actually been dispatched.

- **Bug 1:** on cold load of `/enrollment`, `currentSelections.programId` is empty → default labels.
- **Bug 2:** after switching or deselecting a program, the previous program's `programId` lingers in `currentSelections` → its custom terms leak.

Not a regression from [DHIS2-21635](https://dhis2.atlassian.net/browse/DHIS2-21635) — the state hygiene has been broken for years. The feature just introduced the first consumer (`useTermLabel`) that surfaced it.

[DHIS2-22132]: https://dhis2.atlassian.net/browse/DHIS2-22132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-21635]: https://dhis2.atlassian.net/browse/DHIS2-21635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core_modules/capture-core/components/ScopeSelector/hooks/useResetProgramId.ts
```diff
@@ -1,13 +1,17 @@
+import { useDispatch } from 'react-redux';
 import { useLocation } from 'react-router-dom';
 import { useNavigate, buildUrlQueryString, getLocationQuery } from '../../../utils/routing';
+import { resetProgramIdBase } from '../QuickSelector/actions/QuickSelector.actions';
 import type { PageContext } from './types';
 
 export const useResetProgramId = () => {
+    const dispatch = useDispatch();
```
simonadomnisoru: I'm a bit hesitant about this approach. The `ScopeSelector` hooks were meant strictly for URL navigation and their sole responsibility is to modify the URL params, leaving the Pages themselves responsible for dispatching Redux actions to clean/update their Redux data. Could we try adjusting the fix to keep those two concerns separate? Thanks!

henrikmv: Thanks, I've reworked it. `useResetProgramId` is back to URL-only.

Instead, when the enrollment page's own epic commits or clears its programId (via `COMMIT_TRACKER_PROGRAM_ID`, `COMMIT_NON_TRACKER_PROGRAM_ID`, or `PROGRAM_ID_ERROR`), the currentSelections reducer now mirrors that value. So the enrollment page still owns its own transitions — currentSelections just listens to them, the same way it already listens to `viewEvent.EVENT_FROM_URL_RETRIEVED`.

Heads-up: the underlying staleness has been there for years — on `/enrollment`, `currentSelections.programId` was never being cleared on deselect (that route is intentionally excluded from `withAppUrlSync` since the [DHIS2-16318](https://dhis2.atlassian.net/browse/DHIS2-16318) refactor, and the `resetProgramIdBase()` action designed to clear it has never actually been dispatched). It only became visible now because DHIS2-21635 introduced the first consumer (useTermLabel) that reads that value on `/enrollment`.

Also: while investigating, I noticed the same seam applies to `orgUnitId` on `/enrollment` — nothing keeps `currentSelections.orgUnitId` in sync there either. It's not on the label chain so it's out of scope for this fix, but I will create a ticket for it

henrikmv: Follow-up ticket for `orgunitId`: https://dhis2.atlassian.net/browse/DHIS2-22139

simonadomnisoru: Thanks for looking deeper into this! As you've noticed, `currentSelections` and `withAppUrlSync` have an odd behavior and are quite bug-prone and the long term plan is to get rid of them (see [DHIS2-20888](https://dhis2.atlassian.net/browse/DHIS2-20888)).

What you previously did in `useTermLabel`, to resolve activeProgramId first from the URL, then from enrollmentPage, and finally from currentSelections, that order made sense to me. My only feedback is to avoid dispatching and coupling useResetProgramId to Redux, as the Page itself should dispatch to Redux and handle the cleanup of their data.

henrikmv: @simonadomnisoru  Isn't it better to have `currentSelections` as the single source of truth? What advantages do we get from the fallback chain?

Thanks!

henrikmv: @simonadomnisoru Thank you for making me take a deep dive into this and consider different approaches.

`useTermLabel` now resolves programId from the `URL`, falling back to a `React context` (`ProgramIdContext`) on the three routes whose `URL` doesn't carry one (`/viewEvent`, `/enrollmentEventEdit`, `/enrollment`). Providers on those pages feed the context from the page's own program source. `currentSelections` is untouched.

`URL` is now the primary source; `context` is a per-page fallback. This approach was established with good help from @JoakimSM. Thanks!

simonadomnisoru: Thanks @henrikmv for digging deeper into this! Using `context` here is a great idea. Nice work!

simonadomnisoru (approved):
Nice!

PR #4744, opened 2026-09-20
fix: [DHIS2-21902] Not count SKIPPED events against non-repeatable stage limit

henrikmv:
[DHIS2-21902](https://dhis2.atlassian.net/browse/DHIS2-21902)

Capture was disabling "Create new event" on non-repeatable stages when the only existing event was `SKIPPED`. The backend does not count SKIPPED events against the one-event-per-stage limit, so the frontend now matches that behavior. The rule is extracted into a shared helper so the same check across the enrollment dashboard, quick actions, stage picker, and generate-next-visit stays consistent.

[DHIS2-21902]: https://dhis2.atlassian.net/browse/DHIS2-21902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonadomnisoru (approved):
LGTM!

